PR #31, opened 2018-10-05
Port custom rules

casserni:
brought over the custom rules that I have written so far as well as added some style rules that didn't need custom function. Thinking we can actually combine a few since they are validations and really shouldn't get turned off anyway?


rossmcdonald:
This looks good. One minor change I'd like to make is just to make all tags lower-case by default. There seems to be a mix of capital and no-capital tags right now.

And regarding: 

> combine a few since they are validations and really shouldn't get turned off anyway?

I agree. Right now let's leave them separate, but I'll create an issue for reviewing and consolidating these.

PR #33, opened 2018-10-08
added a set rules function for more clarity

casserni:


src/index.ts
```diff
@@ -40,14 +40,13 @@ export class Spectral {
   private _rules: IRuleStore = {};
 
   // the initial rule config, set on initialization
-  private readonly _rulesets: types.IRuleset[];
+  // @ts-ignore
+  private _rulesets: types.IRuleset[] = [];
```
rossmcdonald: Why are we keeping this around if it's not used anymore?

casserni: Changed the set rules function to use `this._rulesets` instead of input args. Not "really" being used right now but I think its safe to keep in incase we want a getRulesets function or something

PR #41, opened 2018-10-11
add api-host and api-schemes rule, update some language

casserni:
Added an api-host and api-schemes. We require these for use in the ui in lost of places so we should have a rule to enforce that

### DIFF
--- a/src/rulesets/oas/index.ts
+++ b/src/rulesets/oas/index.ts
@@ -79,6 +79,34 @@ export const commonOasRuleset = (): IRuleset => {
           type: RuleType.STYLE,
           tags: ['api'],
         },
+        'api-host': {
+          enabled: true,
+          function: RuleFunction.TRUTHY,
+          input: {
+            properties: ['host'],
+          },
+          path: '$',
+          summary: 'OpenAPI host `description` must be present and non-empty string.',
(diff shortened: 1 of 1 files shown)


src/rulesets/oas/index.ts
```diff
@@ -79,6 +79,34 @@ export const commonOasRuleset = (): IRuleset => {
...
+          },
+          path: '$',
+          summary: 'OpenAPI host `description` must be present and non-empty string.',
+          type: RuleType.STYLE,
+          tags: ['api'],
+        },
+        'api-schemes': {
+          enabled: true,
+          function: RuleFunction.TRUTHY,
+          input: {
```
rossmcdonald: This is a `schema` input. Is that correct?

src/rulesets/oas/index.ts
```diff
@@ -79,6 +79,34 @@ export const commonOasRuleset = (): IRuleset => {
...
+            schema: {
+              items: {
+                type: 'string',
+              },
+              minItems: 1,
+              type: 'array',
+            },
+          },
+          path: '$.schemes',
+          summary: 'OpenAPI host `description` must be present and non-empty array.',
```
rossmcdonald: Same description as above.

PR #46, opened 2018-10-18
Fix oas2 versus oas3 security schemes

marbemac:
They’re in different places.

### DIFF
--- a/src/rulesets/oas/index.ts
+++ b/src/rulesets/oas/index.ts
@@ -32,15 +32,6 @@ export const commonOasRuleset = (): IRuleset => {
           type: RuleType.STYLE,
           tags: ['operation'],
         },
-        'operation-security-defined': {
-          enabled: true,
-          function: 'oasOpSecurityDefined',
-          path: '$',
-          summary:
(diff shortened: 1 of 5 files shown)


src/rulesets/oas/index.ts
```diff
@@ -429,6 +420,36 @@ export const commonOasRuleset = (): IRuleset => {
           tags: ['api'],
         },
       },
+
+      oas2: {
```
casserni: should this be moved to `oas2/index`?

src/rulesets/oas/index.ts
```diff
@@ -429,6 +420,36 @@ export const commonOasRuleset = (): IRuleset => {
...
+          },
+          path: '$',
+          summary:
+            'Operation `security` values must match a scheme defined in the `securityDefinitions` object.',
+          type: RuleType.VALIDATION,
+          tags: ['operation'],
+        },
+      },
+
+      oas3: {
```
casserni: same as above, move to `oas3/index`

src/rulesets/oas/index.ts
casserni: Ahh maybe not because the function is defined under the oas folder

PR #50, opened 2018-11-19
New readme and template files

tbarn:
A few sections are still a WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ We also love to help and support contributors! If you are interested in contribu
 
 - [JSONPath Online Evaluator](http://jsonpath.com/), a helpful tool to determine what `path` you want
 - [stoplightio/json](https://github.com/stoplightio/json), a library of useful functions for when working with JSON
-- [stoplightio/yaml]((https://github.com/stoplightio/yaml)), a library of useful functions for when working with YAML, including parsing YAML into JSON with a source map that includes JSONPath pointers for every property in the result
+- [stoplightio/yaml]((https://github.com/stoplightio/yaml), a library of useful functions for when working with YAML, including parsing YAML into JSON with a source map that includes JSONPath pointers for every property in the result
 
 ## Thanks :)
 


README.md
```diff
@@ -85,6 +125,40 @@ console.log(spectral.run({ spec: 'oas2', target: myOAS }));
 //           light: false } } } ]
 ```
 
+## Example implmentations:
+
+- [Spectral Bot](), a GitHub pull request bot using the [Probot]() framework, built by Taylor Barnett
```
rossmcdonald: Maybe link to your user profile?

README.md
```diff
@@ -85,6 +125,40 @@ console.log(spectral.run({ spec: 'oas2', target: myOAS }));
 //           light: false } } } ]
 ```
 
+## Example implmentations:
```
rossmcdonald: Typo

rossmcdonald: Also probably no need for a colon since it's a heading.

README.md
```diff
@@ -85,6 +125,40 @@ console.log(spectral.run({ spec: 'oas2', target: myOAS }));
...
+
+- [Spectral Bot](), a GitHub pull request bot using the [Probot]() framework, built by Taylor Barnett
+
+## FAQs
+
+**I want to lint my OpenAPI Specification documents but don't want to implement Spectral right now.**
+
+No problem! A hosted version of Spectral comes **free** with the Stoplight platform. Sign up for a free account [here]().
```
rossmcdonald: https://next.stoplight.io/join

README.md
```diff
@@ -85,6 +125,40 @@ console.log(spectral.run({ spec: 'oas2', target: myOAS }));
...
+
+**What is the difference between Spectral and [Speccy](https://github.com/wework/speccy)?**
+
+With Spectral, lint rules can be applied to _any_ JSON object, not just OAS 3 documents. The rule structure is different between the two. Spectral uses [JSONPath](http://goessner.net/articles/JsonPath/) `path` parameters instead of the `object` parameters (which are OAS-specific). Rules are also more clearly defined (thanks to TypeScript typings) and now require specifying a `type` parameter. Some rule types have been enhanced to be a little more flexible.
+
+An example of this includes the ability to specify the object to be linted in the `path` parameter itself, instead of relying on rule-specific options to be applied.
```
rossmcdonald: Can probably just remove this last bit. I'm not sure it adds anything, and is more confusing than anything else.

README.md
```diff
@@ -85,6 +125,40 @@ console.log(spectral.run({ spec: 'oas2', target: myOAS }));
...
 
 - [JSONPath Tester](https://jsonpath.curiousconcept.com/)
+- [Library of useful functions for when working with JSON](https://github.com/stoplightio/json)
+- [Library of useful functions for when working with YAML](https://github.com/stoplightio/yaml)
+
+## Support
+
+If you have a bug or feature request, please open an issue [here](https://github.com/stoplightio/spectral/issues).
+
+If you need help using Spectral or have a support question, please use the [Stoplight Community forum](https://community.stoplight.io). We're created an open source category for these questions. It's also a great place to share your implementations.
```
rossmcdonald: > We're created 

We've created

README.md
tbarn: Going to get a different link for this from Scott.

tbarn:
Okay, the only last thing that needs to be done is complete the features/examples section. I think we need to make it 110% clear what it does. 

Do you have suggestions for those @marbemac? 

Examples could also be things we'd wish that people would build with Spectral.

marbemac:
Ditto Ross - good improvements.

I don't find the examples very helpful - particularly the faked "console.log" output. What do you think about replacing the inline code with links to code sandbox examples? I as a developer would be able to explore the output, and see an end to end working example (each one should stay as simple as possible).

Also, an example of creating a completely new rule to lint some arbitrary object could be interesting? Note, should be clearly distinct from what you can do with JSON Schema - ie should not just check that "x is a string", but rather something more complicated like if "x" is present then "y" should be a boolean, etc. Or some other interesting / fun example.

While we're on the AJV thought, an FAQ describing how this is different than AJV (and json schema validation in general) could be helpful?

README.md
```diff
@@ -1,41 +1,138 @@
...
+[![Maintainability](https://api.codeclimate.com/v1/badges/1aa53502913a428f40ac/maintainability)](https://codeclimate.com/github/stoplightio/spectral/maintainability)
+[![Test Coverage](https://api.codeclimate.com/v1/badges/1aa53502913a428f40ac/test_coverage)](https://codeclimate.com/github/stoplightio/spectral/test_coverage)
 
-This is an enhanced version of the [speccy](https://github.com/wework/speccy)
-project. Differences between this project and speccy include:
+## Features
 
-- Lint rules can be applied to _any_ JSON object, not just OAS3 specifications.
+- Allows you to create custom rules to lint _any JSON object_
+- Built-in set of functions to help build custom rules, like for patterns, parameter checks, alphabetical ordering, a specified number of characters, provided keys are present in an object, etc.
```
marbemac: Suggest add feature indicating can create custom functions for advanced use cases.

README.md
```diff
@@ -1,41 +1,138 @@
...
-- The rule structure has been modified slightly to use
-  [JSONPath](http://goessner.net/articles/JsonPath/) `path` parameters instead
-  of the `object` parameters (which were OAS-specific).
+```shell
+npm install @stoplight/spectral
+```
 
-- Rules are more clearly defined (thanks to TypeScript typings) and now require
-  specifying a `type` parameter.
+Supports Node v8.3+
```
marbemac: and modern browsers

README.md
```diff
@@ -1,41 +1,138 @@
...
 
-- Some rule types have been enhanced to be a little more flexible. An example of
-  this includes the ability to specify the object to be linted in the `path`
-  parameter itself, instead of relying on rule-specific options to be applied.
+## Usage
 
-- Ported to TypeScript.
+There are three key parts to working with Spectral: **Functions**, **rules**, and **rulesets**. **Rulesets** are made up of **rules**, which are built with **functions**.
```
marbemac: Not sure rules are "built" with functions per se? A rule is just a small json object - it does reference a function?

```
**Rulesets** are made up of **rules**, which reference **functions**.
```

README.md
```diff
@@ -1,41 +1,138 @@
...
+const rulesets = {
+  rulesets: [
+    {
+      rules: {
+        mySpec: {
+          snake_case: {
+            type: RuleType.STYLE,
+            enabled: true,
+            summary: 'Checks for snake case pattern',
+            function: RuleFunction.PATTERN,
```
marbemac: Suggest add `console.log(results)` at end of example, and then past into a js file and `node example.js`. You'll see a couple of issues :). First being that `RuleFunction` is not imported.

Should do this exercise with all of the examples - end user should be able to copy paste them and have them work when run file with node.

README.md
```diff
@@ -1,41 +1,138 @@
...
+            path: '$..*',
```
marbemac: Perhaps annotate this example with descriptive comments? In this line:

```js
// http://jsonpath.com selector indicating we want to apply this rule to every property in the object being validated
path: '$..*',
```

README.md
```diff
@@ -56,35 +153,49 @@ const spectral = new Spectral({ rulesets: [defaultRuleset()] });
...
+
+### Example Implementations
+
+- [Spectral Bot](https://github.com/tbarn/spectral-bot), a GitHub pull request bot that lints your repo's OAS document that uses the [Probot](https://probot.github.io) framework, built by [Taylor Barnett](https://github.com/tbarn)
+
+## FAQs
+
+**How is this different than [Ajv](https://github.com/epoberezkin/ajv)?**
+
+Ajv is a JSON Schema validator, not a linter. While Spectral runs your JSON objects through Ajv and returns the result to ensure you are using valid JSON, it goes a level beyond that and allows you to build a custom ruleset that lints your JSON objects.
```
marbemac: > While Spectral runs your JSON objects through Ajv and returns the result to ensure you are using valid JSON

Spectral does expose a `schema` function that you can use in your rules to validate all or part of the target object with JSON Schema (AJV is used under the hood). However, Spectral also provides a number of other functions and utilities that you can use to build up a linting ruleset that validates things JSON Schema is not suited for.

README.md
```diff
@@ -56,35 +153,49 @@ const spectral = new Spectral({ rulesets: [defaultRuleset()] });
...
 
-- [JSONPath Tester](https://jsonpath.curiousconcept.com/)
+- [JSONPath Online Evaluator](http://jsonpath.com/), a helpful tool to determine what `path` you want
+- [stoplightio/json](https://github.com/stoplightio/json), a library of useful functions for when working with JSON
+- [stoplightio/yaml]((https://github.com/stoplightio/yaml)), a library of useful functions for when working with YAML, including parsing YAML into JSON with a source map that includes JSONPath pointers for every property in the result
+
+## Credits
+
+- [Phil Sturgeon](https://github.com/philsturgeon) for collaboration and creating Speccy
+- [Ross McDonald](https://github.com/rossmcdonald) for creating the initial version of Spectral
```
marbemac: Suggest re-naming "Credits" to "Thanks"? Also, I think we can drop off Ross / Stoplight employees.

README.md
tbarn: I'll just add that.

PR #54, opened 2018-12-03
SL-787: same rules under different formats

chris-miaskowski:
Fix the original bug and did some refactoring.

The biggest change is that I removed `_paths`. It said that it tries to assure uniqueness, but the way it was implemented didn't make rules more unique than they are without paths.

Paths was a map of arrays which meant that each array could have multiple rules anyway.

The refactoring removes the redundancy and restructures the code a bit. Hopefully, simplifies it.

### DIFF
--- a/src/__tests__/spectral.test.ts
+++ b/src/__tests__/spectral.test.ts
@@ -12,7 +12,7 @@ describe('spectral', () => {
     });
 
     const results = s.run({ target: todosPartialDeref, spec: 'oas2' });
-    expect(results.length).toBeGreaterThan(0);
+    expect(results.length).toBe(0);
   });
 
   // Assures: https://stoplightio.atlassian.net/browse/SL-786
@@ -92,7 +92,7 @@ Array [
   Object {
     "apply": [Function],
     "format": "oas2",
-    "name": "oas2-ruleName1",
+    "name": "ruleName1",
     "rule": Object {
       "enabled": true,
       "function": "truthy",
@@ -111,7 +111,7 @@ Array [
   Object {
     "apply": [Function],
     "format": "oas3",
-    "name": "oas3-ruleName1",
+    "name": "ruleName1",
     "rule": Object {
       "enabled": false,
       "function": "notContain",
@@ -224,6 +224,6 @@ Array [
     const s = new Spectral({ rulesets });
     const results = s.getRules('oas2');
 
-    expect(results.length).toBe(2);
+    expect(results.length).toBe(1);
   });
 });
--- a/src/__tests__/spectral.test.ts
+++ b/src/__tests__/spectral.test.ts
@@ -1,4 +1,4 @@
-const merge = require('lodash.merge');
+const merge = require('lodash/merge');
 import { Spectral } from '../index';
 import { defaultRuleset } from '../rulesets';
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -79,9 +77,9 @@ export class Spectral {
...
     const rules = [];
 
-    for (const name in this._rules) {
-      if (!this._rules.hasOwnProperty(name)) continue;
-      const { rule, format, apply } = this._rules[name];
+    for (const name in this._rulesByIndex) {
+      if (!this._rulesByIndex.hasOwnProperty(name)) continue;
+      const { rule, format, apply } = this._rulesByIndex[name];
 
       if (!dataFormat || format.indexOf(dataFormat) !== -1) {
```
casserni: Don't really like matching formats by using indexOf, I think it should be an exact match so `format === dataFormat`

src/index.ts
```diff
@@ -94,118 +92,103 @@ export class Spectral {
...
-      for (const ruleName of this._paths[path]) {
-        const { rule, apply, format } = runRules[ruleName];
+  private runAllLinters(opts: IRunOpts): types.IRuleResult[] {
+    return flatten(
+      compact(
+        values(this._rulesByIndex).map((ruleEntry: IRuleEntry) => {
+          if (
+            !ruleEntry.rule.enabled ||
+            (opts.type && ruleEntry.rule.type !== opts.type) ||
+            ruleEntry.format.indexOf(opts.spec) === -1
```
casserni: Same as above using indexOf to match is kind of weird

src/index.ts
```diff
@@ -57,13 +60,8 @@ interface IRunOpts {
 }
 
 export class Spectral {
-  // paths is an internal cache of rules keyed by their path element and format.
-  // This is used primarily to ensure that we only issue one JSON path query per
-  // unique path.
-  private _paths: object = {};
-
   // normalized object for holding rule definitions indexed by name
```
casserni: indexed by `${format}-${name}`

src/__tests__/spectral.test.ts
```diff
@@ -15,6 +15,7 @@ describe('spectral', () => {
     expect(results.length).toBeGreaterThan(0);
   });
 
+  // Assures: https://stoplightio.atlassian.net/browse/SL-786
```
casserni: I like this assures comment, could be good to add to our best practices

src/__tests__/spectral.test.ts
```diff
@@ -50,6 +51,85 @@ describe('spectral', () => {
...
+    ];
+
+    const s = new Spectral({ rulesets });
+
+    expect(s.getRules('oas2')).toMatchInlineSnapshot(`
+Array [
+  Object {
+    "apply": [Function],
+    "format": "oas2",
+    "name": "oas2-ruleName1",
```
casserni: We should probably remove the format from the name since format is included in the object and it seems repetitious

casserni:
Merging despite the differences in code climate. There are a few other spectral updates comming that should improve those

src/__tests__/spectral.test.ts
marbemac: Yeah except this is intended to be a community project, and folks outside stoplight can't access that URL. Perhaps replace with a brief summary of the thing it assures instead of the link?

PR #56, opened 2018-12-03
fix: allowing passing in ruleStore to run instead of using this._

casserni:
[SL-788](https://stoplightio.atlassian.net/browse/SL-788)
[SL-789](https://stoplightio.atlassian.net/browse/SL-789)


src/index.ts
```diff
@@ -178,15 +183,39 @@ export class Spectral {
     return ruleEntry.apply(opt);
   }
 
-  private _parseRuleDefinition(name: string, rule: types.Rule, format: string): IRuleEntry {
+  private _parseRuleSets(
+    rulesets: types.IRuleset[],
+    opts: { includeCurrent: boolean } = { includeCurrent: true }
+  ): { rulesets: types.IRuleset[]; functionStore: IFunctionStore; ruleStore: IRuleStore } {
+    const rSets = merge([], rulesets);
```
marbemac: Would a shallow copy work here? Merging an array feels gross.

const rSets = [...rulesets];

src/__tests__/spectral.test.ts
```diff
@@ -51,6 +51,129 @@ describe('spectral', () => {
     expect(expectedCustomRuleSet).toEqual(givenCustomRuleSet);
   });
 
+  // Assures: https://stoplightio.atlassian.net/browse/SL-789
```
chris-miaskowski: I appreciate you like the convention :)

src/__tests__/spectral.test.ts
```diff
@@ -51,6 +51,129 @@ describe('spectral', () => {
...
+      [
+        {
+          rules: {
+            format: {
+              rule1: false,
+            },
+          },
+        },
+      ],
+      { includeCurrent: true }
```
chris-miaskowski: That'll do the trick!

What you did here reminds me something widely known as the [Flag Argument](https://martinfowler.com/bliki/FlagArgument.html). 

I'm curious what you think about another approach to solving that type of logic. 

One idea, that I'm sure you must have thought about, would be to split the implementation and expose a second function called: `updateRules`. Such method would merge with existing ruleset whereas the `setRules` would override it. 

One `if` statement less :)

With regards,
Chris The If Statement Slayer

src/__tests__/spectral.test.ts
```diff
@@ -177,6 +300,73 @@ Array [
...
+        },
+      },
+    ];
+
+    const s = new Spectral({ rulesets });
+
+    // run again with an override config
+    const run1 = s.run({ target: spec, spec: 'format', rulesets: overrideRulesets });
+
+    expect(s.getRules('format')).toMatchInlineSnapshot(`
```
chris-miaskowski: You may want to consider checking this:

```suggestion
    const s = new Spectral({ rulesets });
    const originalRules = merge({}, s.getRules('format'));
    
    const run1 = s.run({ target: spec, spec: 'format', rulesets: overrideRulesets });
    
    expect(s.getRules('format')).toEqual(originalRules);
```

Would you agree that such check would express the intention (of not overriding the original ruleset) more immediately?

src/index.ts
```diff
@@ -89,26 +89,31 @@ export class Spectral {
     return rules;
   }
 
-  public setRules(rulesets: types.IRuleset[]) {
-    this._rulesets = merge([], rulesets);
-    this._functions = this._rulesetsToFunctions(this._rulesets);
-    this._rulesByIndex = this._rulesetsToRules(this._rulesets);
+  public setRules(
+    rulesets: types.IRuleset[],
+    opts: { includeCurrent: boolean } = { includeCurrent: true }
```
chris-miaskowski: See my comment in one of the test and let me know what you think about breaking that up into two separate, public methods:
- `updateRules`
- `setRules`

src/index.ts
chris-miaskowski: Hi Marc, this actually solves a bug with mutating a deeply nested object. Imagine you pass a ruleset from outside the class and our class mutates on of the deeply nested objects. As a client of the class you may (will) not expect mutations.

Please see https://github.com/stoplightio/spectral/blob/develop/src/__tests__/spectral.test.ts#L19 if you need more details.

As a rule of thumb my opinion is that we should never mutate input parameters because it causes bugs that are hard to spot and results in an unpredictable API. 

Merge is one but not the most elegant solution of course. How about we use `immutablejs`. 
^ @marbemac @casserni?

src/index.ts
```diff
@@ -178,15 +183,39 @@ export class Spectral {
     return ruleEntry.apply(opt);
   }
 
-  private _parseRuleDefinition(name: string, rule: types.Rule, format: string): IRuleEntry {
+  private _parseRuleSets(
+    rulesets: types.IRuleset[],
+    opts: { includeCurrent: boolean } = { includeCurrent: true }
+  ): { rulesets: types.IRuleset[]; functionStore: IFunctionStore; ruleStore: IRuleStore } {
```
chris-miaskowski: How about elevating this to an interface for improved readability?

src/__tests__/spectral.test.ts
casserni: I considered having two separate functions as well, but the internal logic in them felt repetitious. It's actually exactly the same except one passes includeCurrent to _parseRulesets, while one does not

casserni: Separate function could still be a good idea for a clear understanding of what the functions do based on naming convention. And to have a clearer differentiation of responsibility. @chris-miaskowski @marbemac thoughts?

chris-miaskowski: @casserni how about exposing two different public method and using a shared private function to implement them?

src/__tests__/spectral.test.ts
casserni: Agreed

src/index.ts
casserni: Updated

src/index.ts
casserni: If we integrate something like immutable, I think that will be a larger task outside the scope of this one (may require some refactoring to using sets/gets instead of equals everywhere)

PR #60, opened 2018-12-04
SO-19, SO-20 refactor: reuse IValidationResult

chris-miaskowski:
SO-19

BREAKING CHANGE: IRuleResult extends IValidationResult which has a required `severityLabel` field.

I have updated the code to replace IRuleResult with IValidationResult. I left the original interface because it adds its own properties (and because the impact was much lower).

### DIFF
--- a/src/rulesets/oas/index.ts
+++ b/src/rulesets/oas/index.ts
@@ -1,4 +1,5 @@
-import { IRuleDeclaration, IRuleset, RuleFunction, RuleSeverity, RuleType } from '../../types';
+import { ValidationSeverity } from '@stoplight/types/validations';
+import { IRuleDeclaration, IRuleset, RuleFunction, RuleType } from '../../types';
 
 export const operationPath = "$..paths.*[?( name() !== 'parameters' )]";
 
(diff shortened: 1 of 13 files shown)


src/rulesets/oas/index.ts
```diff
@@ -87,7 +88,7 @@ export const commonOasRules = (): IRuleDeclaration => ({
     type: RuleType.VALIDATION,
     summary: 'Path parameters are correct and valid.',
     enabled: true,
-    severity: RuleSeverity.ERROR,
+    severity: ValidationSeverity.Error,
```
casserni: using `ValidationSeverity` is a little weird since the enums resolve to numbers. So really this is like saying `severity: 50` (which doesn't make too much sense to me), maybe we should update `IRuleDeclaration` to have `severity: ValidationSeverityLabel`?

marbemac: Numeric severity values are more useful from a programmatic perspective.

```ts
if (validation.severity >= RuleSeverity.ERROR) {
    // sound the alarm!
}
```

Human facing stuff should now use `severityLabel`.

PR #65, opened 2018-12-07
SO-23: /api rewrite

chris-miaskowski:
### TODO

0. [x] Update readme examples (Marc has done this).
1. [x] remove:
    - [x] Spectral constructor argument 
    - [x] `IRuleset` 
    - [x] `IRuleEntry.format`
2. [x] update `Spectral.run` signature

    ```ts
    public run(target: object, opts?: IRunOpts): IRunResult;

    // returning an object is better than an array, so that we can add properties later without breaking API
    interface IRunResult {
      results: types.IRuleResult[];
    }
    ```

3. [x] simplify `IRunOpts`
  
    ```ts
    interface IRunOpts {
      /**
       * The fully-resolved version of the target object.
       *
       * Some functions require this in order to operate.
       */
      resolvedTarget?: object;
    }
    ```

4. [x] move spectral class to `spectral.ts`
5. [x] move the types defined in `spectral.ts` to the types folder
6. [x] rename `IRuleStore` to `IRuleCollection` (or `IRuleset` now that that name is open?)
7. [x] rename `Spectral.setRules` to `Spectral.addRules`, update signature to take `IRuleCollection`. This does an Object.assign, overwriting existing rules with the same key. 

**Note:** my implementation differs. I don't think Object.assign is the best idea here.

8. [x] rename `Spectral.updateRules` to `Spectral.mergeRules`, update signature to take `IRuleCollection`. This does a deep merge, modifying existing rules with the same key.
9. [x] add `Spectral.addFunctions`
10. [x] update `IRule`

- [x] introduce 'given'
- [x] introduce 'then.function', 'then.functionOptions'
- [x] add 'then.field'
- [ ] ~add 'then.pattern'~
- [x] add 'when.field'
- [x] add 'when.pattern'

    ```ts
    export interface IRule<O = any> {
      // Defaults to RuleType.STYLE
      type?: RuleType;

      // A short summary of the rule and its intended purpose
      summary: string;

      // The severity of results this rule generates
      severity?: ValidationSeverity;
      severityLabel?: ValidationSeverityLabel;

      // A long-form description of the rule formatted in markdown
      description?: string;

      // Tags attached to the rule, which can be used for organizational purposes
      tags?: string[];

      // set to false to disable the rule (enabled by default)
      enabled?: boolean;

      // Filter the target down to a subset[] with a JSON path
      given: string;

      when?: {
        // the `path.to.prop` to field, or special `@key` value to target keys for matched `given` object
        // EXAMPLE: if the target object is an oas object and given = `$..responses[*]`, then `@key` would be the response code (200, 400, etc)
        field: string;

        // a regex pattern
        pattern: string;
      };

      then: {
        // the `path.to.prop` to field, or special `@key` value to target keys for matched `given` object
        // EXAMPLE: if the target object is an oas object and given = `$..responses[*]`, then `@key` would be the response code (200, 400, etc)
        field?: string;

        // name of the function to run
        function?: string;

        // Options passed to the function
        functionOptions?: O;
      };
    }
    ```

11. [x] rules should be enabled by default

12. [x] update custom function signature. no more `ensureRule` needed, functions should return void or an array of result objects that will be transformed into complete result objects by spectral internally.

    ```ts
    export type IFunction<O = any> = (targetValue: any, options: O, paths: IFunctionPaths, otherValues: IFunctionValues) => void | IFunctionResult[];

    export interface IFunctionPaths {
      given: Path;
      target: Path;
    }    

      original: any;
      resolved?: any;
      given: any;
    }

    export interface IFunctionResult {
      message: string;
      path?: Path; // NOTE: Path should come from @stoplight/types
    }
    ```

13. [ ] ~remove Pattern rule.~

14. [x] Final pass through codebase to simplify! The above changes should allow us to simplify a lot of the internals and organization of oas rulesets. The included "rulesets" are now really just a collection of rules and functions that anybody can import and add to their spectral instance via `Spectral.addRules` and `Spectral.addFunctions`. See suggested export structure for rulesets in `## Linting an OAS 2 document:` section of readme.


marbemac:
Ok most of the functionality is in now - still not sure on the `then` property name. Since `given` is now optional, `then` feels weird when used standalone. `assert` might be a better option?

Also, we really should add a pair of tests (failing object and passing object) for each rule in the oas2/oas3 rulesets.

marbemac:
Also, lots of tests added but still net ~500 lines less code. New API is both simpler and more powerful 💯.

screen shot 2018-12-08 at 8 00 15 pm

chris-miaskowski:
@marbemac so is it the code done? Can we merge?

Btw. why is `given` optional?

marbemac:
> so is it the code done? Can we merge?

I think we need a basic pass/fail test for each of the rules in https://github.com/stoplightio/spectral/blob/1-0-umbrella/src/rulesets/oas/index.ts that is not using one of the custom functions (which are already tested). If you have time left if your day any help there would be great - otherwise I'll try to find some time today.

> why is given optional?

It felt unnecessary for all of the rules that target the root object - but thinking about it more I think it's probably best we just keep it required. I'll update that.

marbemac:
Ok, what's left after our call:

- [ ] make given required (marc)
- [ ] simplify `mergeRules` (marc)
- [ ] add `applyRuleDeclarations`, only support boolean for now (marc)
- [ ] enrich `IFunctionResult` with extra props, single test for result structure (marc)
- [ ] snapshot tests instead of length checks?? (marc)

- [ ] pass/fail tests for oas rules. each test should load in just the specific rule it is testing, and define the minimal object required to test the rule (nick)

casserni:
tests are running locally, not sure why they timeout in circle

PR #68, opened 2018-12-11
test(rules): update failing rules to use snapshots

casserni:
Testing update

no


casserni:
@marbemac can we merge this?

PR #80, opened 2019-01-16
fix: a typo and a probable cut-and-paste error

MikeRalphson:
- n/a Docs have been added / updated (for bug fixes / features)

Bug fix

* Typo in comment (harmless). 
* OpenAPI3 test for present and non-empty `servers` array uses the word `string` instead of `array`. Associated tests therefore expect this summary value too.

The correct summary is given for the rule `api-servers`.

Possibly, if users are dependent on value of `summary` property. Users may need to update their harness code.

### Other information

src/rulesets/oas3/index.ts and associated tests:

'specifcication' -> 'specification' in a comment
`servers` must be, and is tested against, being an `array` not a `string`.

### DIFF
--- a/src/rulesets/oas3/__tests__/__snapshots__/api-servers.ts.snap
+++ b/src/rulesets/oas3/__tests__/__snapshots__/api-servers.ts.snap
@@ -10,7 +10,7 @@ Array [
     ],
     "severity": 40,
     "severityLabel": "warn",
-    "summary": "OpenAPI \`servers\` must be present and non-empty string.",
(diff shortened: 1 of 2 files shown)


tbarn:
@MikeRalphson I'm looking into what's going on with CircleCI and getting that worked out. Then I'll merge.

tbarn:
Hey @MikeRalphson, we are pretty sure we got it fixed now. There was an issue with the setting when it runs from forks. But in order to retrigger it again, we believe you'll need to open a new PR. We can't find a way for it to pick up this one. Could you do that?

MikeRalphson:
Not a problem.

Issue #88, opened 2019-02-04
Receiving warning: oasOpParams expects a resolved object


cbetta:
### What is the current behavior?

I am currently migrating my API specification and running into this warning:

```
oasOpParams expects a resolved object, but none was provided. Results may not be correct.
oasPathParam expects a resolved object, but none was provided. Results may not be correct.
```

There are no `$ref` mentions in my spec, so I am confused as to why I am getting this error. The code is as follows.

### What is the expected behavior?

For this not to error

### Please tell us about your environment:

  - Version:   `@stoplight/spectral": "^1.0.1"`
  - Framework: Node v11.7.0
  - Language: Node v11.7.0

### Other information

My code:

```js
// Load Spectral and its OAS3 validations
const { Spectral } = require('@stoplight/spectral')
const { oas3Functions, oas3Rules } = require('@stoplight/spectral/rulesets/oas3')

// Find all OpenAPI files
const fs = require('fs')
const glob = require('glob')

const lint = async () => {
  const fileNames = await new Promise((resolve, reject) => {
    glob('./v2.0/*.openapi.json', {}, (error, match) => {
      if (error) { reject(error) }
      else { resolve(match) }
    })
  })
  
  // Initialize Spectral
  const spectral = new Spectral()
  spectral.addFunctions(oas3Functions())
  spectral.addRules(oas3Rules())
  
  fileNames.forEach(fileName => {
    let specification = JSON.parse(fs.readFileSync(fileName))
    const results = spectral.run(specification)
    // console.dir(results)
  })  
}

lint()
```

marbemac:
Hi @cbetta - great question, and this is something that we could certainly be more clear about (will open a separate PR for readme).

Certain rules (like those param rules you mentioned above) need both the resolved and unresolved document in order to function. Since you don't have any $refs in your document, you should be able to supply it as the resolved version like so:

```js
const results = spectral.run(specification, { resolvedTarget: specification });
```

You can find the run signature here: https://github.com/stoplightio/spectral/blob/next/src/spectral.ts#L21.

I'm now wondering if we shouldn't just default resolvedTarget to the passed in document if resolvedTarget is not explicitly set.

Let me know if the above works for you for now!

cbetta:
Ah yes that works. The documentation wasn't quite up to date on this indeed.

tbarn:
I'm adding a FAQ for this right now.

cbetta:
Did I miss something? I don't see anything added.

[...]

PR #99, opened 2019-03-13
[SL-2024] Use json-ref-parser by default to provide dereferenced object

philsturgeon:

This feature will allow referencing to be done by Spectral, to stop the user needing to figure that out too.

Users of the JS library are forced to pass in the specification and an object with resolvedTarget:

```js
const results = spectral.run(specification, { resolvedTarget: specification });
```

This was causing confusion for users (https://github.com/stoplightio/spectral/issues/88), who were not even using $ref.

If spectral is told to resolve things, it will.

```js
const results = await spectral.run(specification, { resolve: true });
```

The `spectral.run()` method now returns a `Promise<object>` instead of `object`. The format of the object is the same.


marbemac:
Ok good start but a few concerns:

**One**

Before we had resolved and original objects. By operating only on the resolved value, it will report many issues for json paths that don't actually exist on the original document. For example:

1. imagine a rule that checks for camelCase
2. there is a model in definitions that breaks this rule
3. that model is $ref'd 50 times throughout the oas document
4. be operating on the resolved value, we end up reporting 50 issues, and most of those issues with have json paths pointing to non-existent spots in original document

**Two**

Do we want to for this to be async? The only reason it needs to be now is because of the resolving. I imagine that in many cases the end user might already have the resolved value, so resolving again is just extra work (such as our Studio use case). We could have a sync `run` method that allows you to optionally pass in the already resolved val (similar to what we do now but perhaps better worded?), and an async `resolveAndRun`?

**Three**

If we are going to resolve for the end user, we might as well resolve files and urls as well. See json-ref-resolver repo for examples of how to do that.

---

All that said, a final alternative is to just take the simplest approach for right now, and update the two rules that require a "resolved" value. This would look like:

1. remove `resolvedTarget` options from run, no longer needed and can add later in a non breaking change if needed
2. create the simplest of "lookup" functions that takes a local pointer like `#/parameters/userId` and looks the value up
3. update `oasOpParams` and `oasPathParam` functions to use the function in number 2 if the val is a $ref that starts with `#`.
4. If val is $ref that does not start with `#`, warn. This is the only use case this solution would not support - if end user is using remote $refs in the parameters object. We could document this and address later?

philsturgeon:
I replied to one on Jira, but 2 and 3:

*Two*

Absolutely, right now it is only async because of the resolver, but when I asked the guys here in Poland the consensus was that it would probably need to become async at some point, and better to do it now instead of after v1.0.0. I don't know if you agree with this, but it seemed like a reasonable assumption,

**Three**

Seeing as it is using json-ref-resolver, I was under the impression it was already handling externals, but I did not test that functionality as it was not a priority for this issue. If we use json-ref-resolver then absolutely, but maybe we could get resolver to just resolve and not also dereference. That would solve a lot of the confusion over dereferencing vs resolving.

philsturgeon:
Ok we chatted about **One** elsewhere and @marbemac made some great suggestions. 

> Resolving remote values would indeed be nice, however it would introduce a dependency on fs in order to resolve files. We should probably make that optional. Could have a resolvers directory with three ready to go versions of our resolver - one that does local only, one that does local and http, and one that does local, http, and file? Then end user could import the one they want and pass that as an argument when creating the spectral instance? Our ref resolver also has a great cache, so by allowing end user to pass in a resolver instance it will really help us in studio since we already have one we can pass to spectral.
> 
> Since we don’t bundle spectral, as long as the resolvers are in their own folder and not re-exported, should be ok to use in browser.
> 
> Example:

```
const { resolver: httpResolver } = require('@stoplight/spectral/resolvers/http');
const { resolver: httpAndFileResolver } = require('@stoplight/spectral/resolvers/all');

// passing in a resolver is optional - by default, spectral uses a resolver that resolves/dereferences local $refs
// pass in one of these (or your own json-ref-resolver instances!) to also handle remote $refs
// resolver should be an instance of https://github.com/stoplightio/json-ref-resolver/blob/master/src/resolver.ts#L11
const spectral = new Spectral({ resolver: httpAndFileResolver });

const results = await spectral.run({
  name: 'helloWorld',
});
```

> This allows us to:
> 
> a) get rid of confusing resolvedTarget thing
> 
> b) sane default where it just resolves (or dereferences if you prefer :) local refs
> 
> c) nice support for more complex resolution use cases
> 
> d) studio / graphite can pass in the custom resolver instance we already have to take advantage of the caching
> 
> e) we can add the validations from resolver to the spectral diagnostics, so invalid $refs are reported!
> 
> f) Running multiple times will take advantage of the built in resolver cache.

So, I've made what I think are the appropes changes to this PR. 

@marbemac Please let me know how it looks now. Also do you have any feedback to my replies RE points 2 and 3?

marbemac:
Much better, can add some ready to go resolvers in a separate PR (or perhaps we add those to json-ref-resolver repo and simply re-export here?).

Note I pushed a couple of tiny commits here to fix the tests.

> would probably need to become async at some point, and better to do it now instead of after v1.0.0. I don't know if you agree with this, but it seemed like a reasonable assumption

I agree! Let's do it. Note that this is a breaking change, and spectral is technically already at `v1.2.0` since we are using semantic release. No worries, we'll just have to make sure to document the breaking changes when this PR is merged so that it bumps to v2.0.0.

> Seeing as it is using json-ref-resolver, I was under the impression it was already handling externals, but I did not test that functionality as it was not a priority for this issue. If we use json-ref-resolver then absolutely, but maybe we could get resolver to just resolve and not also dereference. That would solve a lot of the confusion over dereferencing vs resolving.

By default `json-ref-resolver` only resolves local references e.g. `#/definitions/user`. This is a sane default for spectral I think, and folks can pass in a ref resolver instance that includes remote resolution targets if they wish.

--- 

A couple of asks:

1. Update docs
2. Add some tests that verify usage of resolved is working. We only use it in `src/rulesets/oas/functions/oasOpParams` and `src/rulesets/oas/functions/oasPathParam`, and I just realized we do not have any tests covering the case where the parameter is a $ref.

After that I think we're in pretty good shape.

philsturgeon:
Updating docs: I'm leaving this as a hidden feature until we actually have resolvers to pick from, that should be another later change I think. I should definitely update this to mention that run is async now! 

Tests on oasOpParams & oasPathParam: Yep! Wanted to make sure this was about right before going the whole hog on other tests. 👼

philsturgeon:
@marbemac ok one more thing, whilst I'm making breaking changes to run(), can I get rid of the`{ results: [] }` wrapper? Is there a strong reason why its there? I guess its for future proofing, but I think we always want to return results or rejections with an error, so there should not be a need for this? 

It'll make for writing nice async tests too, which should not be a driving force for code design, but a nice side benefit.

philsturgeon:
@marbemac ok one more thing, whilst I'm making breaking changes to run(), can I get rid of the`{ results: [] }` wrapper? Is there a strong reason why its there? I guess its for future proofing, but I think we always want to return results or rejections with an error, so there should not be a need for this? 

It'll make for writing nice async tests too, which should not be a driving force for code design, but a nice side benefit.

src/types/function.ts
```diff
@@ -15,7 +15,7 @@ export interface IFunctionPaths {
 export interface IFunctionValues {
   original: any;
   given: any;
-  resolved: any;
+  resolved?: any;
```
philsturgeon: It shouldnt be considered optional anymore as we _always_ pass it in.

marbemac:
> ok one more thing, whilst I'm making breaking changes to run(), can I get rid of the{ results: [] } wrapper?

Yup, future proofing - I generally avoid returning arrays directly. However, feels pretty safe to offer up a nicer dx and return the array directly - up to you! I'm ok either way.

Otherwise I think we're good to go - can you create a new v2 branch off of master and target that w this PR?

philsturgeon:
@marbemac done and done. You got a thumb for me?! 🥂

PR #100, opened 2019-03-14
[SL-2190] Removed schema-items-is-object rule 

philsturgeon:
Probably related to #99, and involves us thinking about how and when we care about refs.

Bug fix

Currently some rules, like schema-items-is-object, will only work if 
$ref is not used. This suggests that the specification needs to be
entirely dereferenced for this rule to work, which doesnt match current
expectations.


marbemac:
I forget if we got this rule from speccy or from a customer, but it feels a bit odd in general. What is this really trying to check? Should it just check `'$..schema.items'` or should we be checking all json schema models e.g. `'$..schema.items'` AND `'$.definitions.*.items'` etc. Curious if the rule is even worth keeping around in the first place?

philsturgeon:
Yeah it should either look at both `'$..schema.items'` and `'$.definitions.*.items'`, or neither. This is something @MikeRalphson could potentially explain a little better as I think it was actually a layover from swagger2openapi, not something I wrote into speccy.

marbemac:
I'd be OK getting rid of it as part of v2. Does not seem terribly useful. In fact, there might be a few more questionable rules in the standard sets? Not sure.

chris-miaskowski:
> This suggests that the specification needs to be
entirely dereferenced for this rule to work, which doesnt match current
expectations.

Why is that an issue though? Someone may simply not use $refs in their spec (which is effectively equal to being dereferenced), right?

philsturgeon:
Old rule copied from Speccy, not wildly useful as our schema validation rule should figure out if the OpenAPI is valid or not, without needing an extra rule to assert that again.

chris-miaskowski:
> Old rule copied from Speccy, not wildly useful as our schema validation rule should figure out if the OpenAPI is valid or not, without needing an extra rule to assert that again.

Ok :+1: makes sense!

PR #101, opened 2019-03-15
feat(oas): check examples are valid against their schema

philsturgeon:
If an example is provided, make sure it is valid against the schema for that property.

SL-2065

Feature

Example is pretty much ignored.

Now examples with nonsense values, formats that are not matched, etc, will product validation errors. 

This will create a breaking change if you have invalid examples, so you will want to fix the examples.

### DIFF
--- a/src/functions/schema-path.ts
+++ b/src/functions/schema-path.ts
@@ -1,0 +1,36 @@
+/*
+ * Schema Path is a variant of the normal Schema rule with a slight twist.
+ * Instead of passing the rule a schema to validate, there is already a schema
+ * somewhere inside the target, and we are just selecting it, and applying it to
+ * a given field.
+ *
+ * The primary use case for this was validating OpenAPI examples
+ * against their schema, but this could be used for other things.
+ */
+import { IFunction, ISchemaPathOptions } from '../types';
+import { schema } from './schema';
+
+import jp = require('jsonpath');
+
+export const schemaPath: IFunction<ISchemaPathOptions> = (targetVal, opts, paths, otherValues) => {
+  if (!targetVal) return [];
+  let value = targetVal;
+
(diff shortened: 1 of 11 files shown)


src/functions/schema-path.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (opts.field && typeof targetVal === 'object') {
+    value = targetVal[opts.field];
+  }
+
+  if (!value) return [];
+
+  // Hunt in the original targetVal for the schema we are going to apply
+  let schemaObj;
+  try {
+    schemaObj = jp.value(targetVal, opts.schemaPath);
```
casserni: Should we remove the field value we are testing on out of the `schemaObj`?

So if we are testing the `example` property in a schema should we remove that property from the schema

casserni: `delete schemaObj[opts.field]`

src/types/rule.ts
```diff
@@ -82,3 +82,10 @@ export interface ISchemaOptions {
   schema: object;
 }
 export type SchemaRule = IRule<RuleFunction.SCHEMA, ISchemaOptions>;
+
+export interface ISchemaPathOptions {
+  schemaPath: string;
+  // the `path.to.prop` to field, or special `@key` value to target keys for matched `given` object
+  field?: string;
```
casserni: Can field also be a path? So we aren't limited to top level properties?

src/functions/schema-path.ts
philsturgeon: It doesnt hurt having it there. It's being ignored by the validator, which is why we need this rule. :)

src/types/rule.ts
casserni: TODO/feature request so resolving

PR #102, opened 2019-03-15
A few more tests for valid-example

philsturgeon:

More tests. 

Chatting to @casserni we had a few concerns but it turns out everything is ooookay. These tests should live in here anyhow.

### DIFF
--- a/src/rulesets/oas2/__tests__/valid-example.ts
+++ b/src/rulesets/oas2/__tests__/valid-example.ts
@@ -85,14 +85,13 @@ describe('valid-example', () => {
     expect(results.results).toHaveLength(1);
   });
 
-  test('will error with totally invalid input', () => {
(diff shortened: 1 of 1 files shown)


src/rulesets/oas2/__tests__/valid-example.ts
```diff
@@ -85,14 +85,13 @@ describe('valid-example', () => {
     expect(results.results).toHaveLength(1);
   });
 
-  test('will error with totally invalid input', () => {
+  test('will only fail for deeply nested invalid item, not valid parents', () => {
```
casserni: should this say "invalid parents"?

philsturgeon: > will only fail for deeply nested invalid item, *but* not *for* valid parents

PR #104, opened 2019-03-18
feat: new shape of validation results [SL-2123]

P0lip:
TODO:

- [x] Switch to npm version of `@stoplight/types` once released
- [x] Make IDiagnostic['range'] optional? `parseWithPointers` function exposed by [JSON](https://github.com/stoplightio/json/pull/12)/[YAML](https://github.com/stoplightio/yaml/pull/6)/[Markdown](https://github.com/stoplightio/markdown/pull/5) parsers will be capable of providing the required information, but Spectral would need to make use of it for text files (when CLI mode is added to Spectral). Not sure about object literals (current usage via API), as they aren't expressed in text format.
I guess we could simply treat each property as it would be placed on a separate line and indentation could be based on actual nesting depth (BASE * depth).

Feature (breaking). 

The shape of a single validation object is different.

Yes, as stated above, the shape of single validation is different, so expect some properties to be missing, new ones to be added or their values to be changed.

#### What changes might users need to make in their application due to this PR?

Use new properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,7 @@ console.log(JSON.stringify(results, null, 4));
 //   {
 //     "name": "snake_case",
 //     "message": "must match the pattern '^[a-z]+[a-z0-9_]*[a-z0-9]+$'",
-//     "severity": 40,
-//     "severityLabel": "warn",
+//     "severity": 1,
 //     "path": [
 //       "name"
 //     ]
@@ -134,8 +133,7 @@ console.log(JSON.stringify(results, null, 4));
 //   {
 //     "name": "openapi_not_swagger",
 //     "message": "Use OpenAPI instead of Swagger!",
-//     "severity": 40,
-//     "severityLabel": "warn",
+//     "severity": 1,
 //     "path": [
 //       "description"
 //     ]
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "dependencies": {
     "@stoplight/json": "1.x.x",
     "@stoplight/json-ref-resolver": "^1.2.2",
-    "@stoplight/types": "3.x.x",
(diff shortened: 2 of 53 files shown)


marbemac:
👍 a couple of thoughts:

- will need some way to get `summary` for given `code` (should be easy lookup on rule)
- can we figure out a way to not include `range` at all for right now (without changing @stoplight/types)? can add range later once we have a proper solution

P0lip:
> can we figure out a way to not include range at all for right now (without changing @stoplight/types)? can add range later once we have a proper solution

I believe the easiest way would be the following
```diff
-export interface IRuleResult extends IDiagnostic {
+export interface IRuleResult extends Omit<IDiagnostic, 'range'> {
  path: JSONPath;
}
```

P0lip:
> will need some way to get summary for given code (should be easy lookup on rule)

Doesn't `message field serve that purpose?

marbemac:
> Doesn't `message field serve that purpose?

Message is the actual error message, summary is the rule summary. We use summary in the UI so that end user understands what the rule is trying to do.

P0lip:
> Message is the actual error message, summary is the rule summary. We use summary in the UI so that end user understands what the rule is trying to do.

Shall we keep it exclusive to `IRuleResult`? Don't really see how they would be useful for json/yaml parsing errors.

philsturgeon:
Can you change the base branch for this PR to release/2.0 please?

marbemac:
> Shall we keep it exclusive to IRuleResult? Don't really see how they would be useful for json/yaml parsing errors

Yeah `summary` should remain specific to Spectral so can just add back here.

Otherwise looks good!

PR #106, opened 2019-03-22
Positional info [SL-2197]

P0lip:

Range is now attached to a single validation object, therefore it's possible to see where the issue/error actually occurs in a file (when using Spectral using API)
Moreover, CLI makes use of them to display line and column as shown in the screenshot below

image

Range is not included (API usage) and positional info is not displayed (CLI usage).

image

See the screenshot above.

No changes required!

### Other information


philsturgeon:
RE the chat we had elsewhere about showing file path or JSON path: I think we should include both in JSON, and only show file path in the CLI. Most users will be really excited to be able to CTRL/CMD+Click  the file path and open it right up. If folk need the path information it will be available in JSON and other formats as we grow them out, and if users request we find a way to include it in stylish we can do that later.

Does that sound good?

P0lip:
@philsturgeon 
Sounds good to me!
I'm adding filenames at the very moment :)

philsturgeon:
Awesome! This means v2.0 is very close.

src/cli/commands/lint.ts
```diff
@@ -84,7 +85,7 @@ async function lint(name: string, flags: any, command: Lint) {
 
   let results = [];
   try {
-    results = await spectral.run(obj.data);
+    results = await spectral.run(obj.data, { uri: join(process.cwd(), name), ...obj });
```
philsturgeon: Can we avoid join cwd and use `path.resolve()` instead? We had this issue in prism where it would append onto _anything_ you passed in, which broke URLs, absolute paths, etc.

src/spectral.ts
```diff
@@ -25,9 +26,9 @@ export class Spectral {
     this.resolver = opts && opts.resolver ? opts.resolver : new Resolver();
   }
 
-  public async run(target: object): Promise<IRuleResult[]> {
+  public async run(target: object, parserMeta?: IParserMeta): Promise<IRuleResult[]> {
```
P0lip: `parserMeta` is needed to calculate lines and columns for given JSON Path.

Ideally, we could have a single argument accepting `IParserResult`, but that would be breaking and would force users to use `parseWithPointers` functions.

If `parserMeta` is provided, range property (lines and columns) is attached to a single result and hence can be displayed in the output (CLI) or consumed programmatically (API).

@philsturgeon @tbarn your thoughts on that?
We will need to document it for sure.

philsturgeon: What is this thing for?

P0lip: Also, `IRange` (range property) follows [LSP](https://microsoft.github.io/language-server-protocol/specification#range) standard and hence lines and characters (columns) are 0-based.

P0lip: @philsturgeon It's needed by `getJsonPathForPosition`

tbarn:
I'm getting an error when I do `yarn build`:

```
➜  spectral git:(feat/positional-info) yarn build
yarn run v1.13.0
$ sl-scripts build && oclif-dev manifest && node ./scripts/prepare-cli
building......
src/linter.ts:6:10 - error TS2305: Module '"../../../../../Users/taylorbarnett/Stoplight/spectral/node_modules/@stoplight/yaml"' has no exported member 'getLocationForJsonPath'.

6 import { getLocationForJsonPath } from '@stoplight/yaml';
           ~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
building...... done
error Command failed with exit code 2.
```

P0lip:
@tbarn make sure to run `yarn` before (dependencies have been updated).

tbarn:
@P0lip I did

tbarn:
@P0lip I did but I missed that it had an error too: 

`error sane@4.1.0: The engine "node" is incompatible with this module. Expected version "6.* || 8.* || >= 10.*". Got "9.11.2"`

P0lip:
@tbarn 
Either 8 or 10 (recommended as it's the latest LTS).

In any case, I don't see a reason not to support Node 9, therefore, the issue you expected should be fixed.
It seems like we might need to find out why `sane` doesn't accept Node 6 and node 9.

tbarn:
@P0lip I probably set my nvm to v9 to test something else. I just want to make sure we have these kinds of things recorded in the contributing doc for others. 

I got it running, tested out a few different spec documents, and it looks great to me.

P0lip:
@tbarn 
Did `yarn --ignore-engines` work for you?
If so, it's a good candidate to be listed in some troubleshooting section until we fix the issue.

tbarn:
Shared elsewhere, but I wanted to share here too. The bump on the stoplight/yaml package fixed the problem on v9.

README.md
```diff
@@ -152,6 +152,33 @@ console.log(JSON.stringify(results, null, 4));
 // ]
 ```
 
+### Positional information
```
P0lip: @philsturgeon @tbarn 
Added some piece of code to explain how positional information can be obtained.
Let me know if it's all clear for you.

There is no reasonable way to produce ranges (lines and columns) for already parsed content / object literals, e.g.
```js
const myObj = {
  foo: 'bar',
};
```

We cannot really produce anything meaningful in such case, therefore no range will be attached unless all the steps described below are taken.

src/spectral.ts
P0lip: I know the above looks slightly weird, but it preserves the old behavior.
It simply degrades gracefully - if no `parserMeta` is provided, all validation results will still have all the properties except range (and source - name of the file).

We could have a single argument, but we'd force user to use `parseWithPointers` function in such case, which is something we most likely want to avoid.
The other possible solution is to have another method, i.e. `runParsed` or similar, but IMHO this would introduce even bigger confusion to end-users.

@tbarn @philsturgeon @marbemac thoughts?
Note, The solution is also briefly described in the readme above.
Let me know what your thoughts are.

marbemac: I'd vote for keeping run as is, with a single argument, and simply `parseWithPointers(safeStringify(target))`. Basically re-stringify it if an object is passed in. Slight trade off re a potential extra parse but not a huge deal for the linting use case imho.

Thoughts?

README.md
marbemac: See https://github.com/stoplightio/spectral/pull/106#discussion_r268898760, I think we should consider just re-parsing if an object is passed in.

src/spectral.ts
P0lip: Yeah, sounds good to me.
What about S^2 (ya know what I mean I hope)? I'd like to avoid reparsing as much as possible in this particular case as this may degrade performance.

P0lip: Or did you mean to have the following implementation

```
run(target: IParserAstResult | object) {
  // some check here to see if we deal with parsed content or not.
}
```

marbemac: What about something like this?

Not sure we even need to document the `IParserAstResult` right now - can use internally and from studio for now.

```
run(target: IParserAstResult | object | string) {
  let parsed = target;
  if (!isParserAstResult(target)) {
    parsed = parseWithPointers(safeStringify(target));
  }

  // ...resolve parsed, etc
}
```

P0lip: The above looks good to me.

philsturgeon: Yeah having it there but hidden from docs is fine with me.

package.json
```diff
@@ -61,10 +61,10 @@
   "dependencies": {
     "@oclif/command": "^1",
     "@oclif/plugin-help": "^2",
+    "@stoplight/json": "1.8.0",
     "@stoplight/json-ref-resolver": "^1.2",
-    "@stoplight/json": "1.x.x",
     "@stoplight/types": "4.x.x",
-    "@stoplight/yaml": "1.2.x",
+    "@stoplight/yaml": "2.1.0",
```
philsturgeon: Can we start using `^2.1.0`so that we allow 2.1.1 and 2.1.2 without needing to changew the package file?

src/spectral.ts
```diff
@@ -25,9 +28,16 @@ export class Spectral {
     this.resolver = opts && opts.resolver ? opts.resolver : new Resolver();
   }
 
-  public async run(target: object): Promise<IRuleResult[]> {
-    const resolvedTarget = (await this.resolver.resolve(target)).result;
-    return runRules(target, this.rules, this.functions, { resolvedTarget });
+  public async run(target: IParserMeta | object | string): Promise<IRuleResult[]> {
```
philsturgeon: Instead of messing with the target argument, could we possibly add a second optional argument for this stuff? 

then we can throw warnings saying "We cannot give you any positional information because you didnt give us enough information" and the output can be ?? or something? Does that make any sense?

P0lip: Yeah, we had that done this way initially as discussed here https://github.com/stoplightio/spectral/pull/106#discussion_r268676451

philsturgeon: Ah, my comment saying `Yeah having it there but hidden from docs is fine with me.` was written from mobile and I hadn't seen marcs suggestion.

Ok, so the name IParserMeta is whats throwing me. Marcs example makes it clear that it can either be a parsed AST object, or a generic object, and if its the generic object then its not gonna work with line numbers and stuff. That's all fine if we use a name a bit more like his. 

Metadata is usually a different thing to data, so if it has meta in the name I'd expect it to be a different argument.

P0lip: @philsturgeon 
Yeah, naming is certainly not my strongest asset :laughing:

I can name it `IParsedAst` if it makes everything more clear.

philsturgeon:
Perfect!

src/spectral.ts
P0lip: Renamed to `IParsedResult` since it also has some different properties than just `ast`, so this could be confusing.
Let me know what you think @philsturgeon

marbemac: > and if its the generic object then its not gonna work with line numbers and stuff

It should still work if go with original suggestion. Basically, if it's a plain object we stringily it and parse it with our parser to get positional info:

```ts
run(target: IParserAstResult | object | string) {
  let parsed = target;
  if (!isParserAstResult(target)) {
    parsed = parseWithPointers(safeStringify(target));
  }

  // ...resolve parsed, etc
}
```

P0lip: Yeah, we reparse it therefore some ranges are attached. Initially, when we didn't reparse the object, ranges were missing.

P0lip: Yeah, we reparse it therefore some ranges are attached. Initially, when we didn't reparse the object, ranges were missing.

PR #114, opened 2019-03-27
Docs/2.0 update

tbarn:

Docs Update

Different organization

Reorganizes the docs and puts the more important info first with links to the more advanced stuff

Nope

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ const { Spectral } = require('@stoplight/spectral');
 const { oas3Functions, oas3Rules } = require('@stoplight/spectral/rulesets/oas3');
 
 // an OASv3 document
-var myOAS = {
+const myOAS = {
   // ... properties in your document
   responses: {
     '200': {
--- a/README.md
+++ b/README.md
@@ -128,10 +128,6 @@ spectral.addRules({
   },
 });
 
-const results = await spectral.run({
-  name: 'helloWorld',
-});
-
 // run!
 spectral.run({name: 'helloWorld',}).then(results => {
   console.log(JSON.stringify(results, null, 4));
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ A flexible JSON object linter with out of the box support for OpenAPI Specificat
 
 - Create custom rules to lint _any JSON object_.
 - Use JSON paths to apply rules / functions to specific parts of your JSON objects.
-- Built-in set of functions to help build custom rules. Functions include pattern checks, parameter checks, alphabetical ordering, a specified number of characters, provided keys are present in an object, etc.
-- Create custom functions for advanced use cases.
+- Built-in set of functions to help [build custom rules](#creating-a-custom-rule). Functions include pattern checks, parameter checks, alphabetical ordering, a specified number of characters, provided keys are present in an object, etc.
+- [Create custom functions](#creating-a-custom-function) for advanced use cases.
 - Optional ready to use rules and functions to validate and lint OpenAPI Specification (OAS) 2 _and_ 3 documents.
 - Validate JSON with [Ajv](https://github.com/epoberezkin/ajv).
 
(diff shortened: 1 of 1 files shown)


tbarn:
One problem, the examples no longer just work™️ because they are not in async functions. Need to figure out how to fix that in the least complicated way possible.

tbarn:
I'm not going to merge until I get the code examples fixed, trying to figure out the most basic example right now.

P0lip:
@tbarn 
There are 3 ways to fix the examples:
- wrap the code with async function:

```ts
(async () => {
  const results = await spectral.run(myOAS);
  console.log(JSON.stringify(results, null, 4));
})();
```

- 'consume' promise using then rather than `await`

spectral.run(myOAS).then(results => {
 console.log(JSON.stringify(results, null, 4));
});

- wait until top-level await is added to spec - there is a proposal for this, but I doubt it lands soon.

README.md
```diff
@@ -230,12 +231,18 @@ With Spectral, lint rules can be applied to _any_ JSON object. Speccy is designe
...
 
+### Example Implementations
+
+- [Stoplight's Custom Style and Validation Rules](https://docs.stoplight.io/modeling/modeling-with-openapi/style-validation-rules) uses Spectral to validate and lint OpenAPI documents on the Stoplight platform
+- [Spectral GitHub Bot](https://github.com/tbarn/spectral-bot), a GitHub pull request bot that lints your repo's OpenAPI document that uses the [Probot](https://probot.github.io) framework, built by [Taylor Barnett](https://github.com/tbarn)
+- [Spectral GitHub Action](https://github.com/XVincentX/spectral-action), a GitHub Action that lints your repo's OpenAPI document, built by [Vincenzo Chianese](https://github.com/XVincentX/)
+
 ## Helpful Links
 
 - [JSONPath Online Evaluator](http://jsonpath.com/), a helpful tool to determine what `path` you want
```
P0lip:
```diff
- [stoplightio/yaml](https://github.com/stoplightio/yaml), a library of useful functions for when working with YAML, including parsing YAML into JSON with a source map that includes JSONPath pointers for every property in the result
+ [stoplightio/yaml](https://github.com/stoplightio/yaml), a library of useful functions for when working with YAML, including parsing YAML into JSON, and a few helper functions such as `getJsonPathForPosition` or `getLocationForJsonPath`.

Comment (P0lip) [reply README.md]: ^ we don't include the list of pointers anymore, therefore the current description is no longer relevant.

Comment (tbarn) [reply README.md]: good point. fixing...

Comment (P0lip) [reply README.md]: It's been changed very recently, at the end of last week.
`@stoplight/json` package exposes these methods as well. :)

**Comment by tbarn** [comment]
I've personally moved more towards using the second way, do you think it is good universally?

**Comment by wmhilton** [comment]
I think the 2nd way is probably the best for short examples. The first way starts becoming appealing to me after about the fifth `.then()` in a chain, if you aren't using the results. For example

```js
spectral.init()
.then(() => {
  return spectral.start();
}).then(() => {
  return spectral.run();
}).then(() => {
  return spectral.finish();
}).then(() => {
  return spectral.getResults();
}).then(results => {
  console.log(results);
})
```
^ 4 fluffy lambdas `() => { }`

vs

```js
(async () => {
  await spectral.init();
  await spectral.start();
  await spectral.run();
  await spectral.finish();
  let results = await spectral.getResults();
  console.log(results);
})();
```
^ 1 fluffy async IIFE `(async () => { })();`

where by fluffy I mean they just add fluff. fluff that is necessary to run the code maybe, but adds no real value or meaning

P0lip:
We could also have an async function and call it, i.e.

```
async function main() {
  const results = await spectral.run(myOAS);
  console.log(JSON.stringify(results, null, 4));
}

main();
```

tbarn:
@wmhilton over here making 🐑 puns lol

I'm changing all them to the second way right now

tbarn:
Done! @P0lip I ran each example (with `node test.js`) from the built v2.0 package so it should be all good now.

README.md
```diff
@@ -85,7 +132,10 @@ const results = await spectral.run({
   name: 'helloWorld',
 });
 
-console.log(JSON.stringify(results, null, 4));
+// run!
+spectral.run({name: 'helloWorld',}).then(results => {
```
P0lip: There is a `await` statement above. We need to remove it.

README.md
```diff
@@ -85,7 +132,10 @@ const results = await spectral.run({
   name: 'helloWorld',
```
P0lip: To be removed :) We have a new code below.

README.md
```diff
@@ -45,9 +45,56 @@ Other options include:
...
+Spectral includes a number of ready made rules and functions for OpenAPI v2 and v3 documents.
+
+This example uses the OpenAPI v3 rules to lint a document.
+
+```javascript
+const { Spectral } = require('@stoplight/spectral');
+const { oas3Functions, oas3Rules } = require('@stoplight/spectral/rulesets/oas3');
+
+// an OASv3 document
+var myOAS = {
```
P0lip: Very very minor comment, but should we use `const`?
It doesn't change anything at all yet iit seems like most of the examples use `const`, so might be worth consistent to avoid any confusion?
```suggestion
const myOAS = {
```

README.md
tbarn: Lol dang it, you beat me before I could push that fix. :P

tbarn:
@P0lip I'll get this documented in a future PR after release.

PR #144, opened 2019-04-16
Feature/sl-2295/enable custom ruleset via cli

chris-miaskowski:

Feature

Currently, the CLI doesn't support any kind of custom ruleset

Adds the ability to reference a custom ruleset with:

`spectral file.yaml --ruleset=path/to/ruleset.yaml`

@chris-miaskowski any breaking changes? 

### Other information


tbarn:
@chris-miaskowski I'd love to check this out. How do I run it before it is merged?

philsturgeon:
This is going to need a rewrite sorry, we chatted  about it on Slack. It’s basically letting users define custom rules in a config file, which is basically what rulesets are. Instead we need to add a CLI option to enable already defined rulesets. 

Similar but different. :)

chris-miaskowski:
@philsturgeon I appreciate your feedback but I completely don't get it.

The story says:

Selection_004

> As a Spectral user, I want to be able to **specify my own rulesets**, which **may contain custom rules**, and this should all work via the command line.

My story solves exactly that issue. User selects a file with rules (and only rules) and Spectral loads it (on top of the default rulesets - I could skip loading default).

> spectral lint foo.yaml --ruleset=something.yaml --ruleset=http://example.com/other.yaml

Isn't that line clearly stating that the ruleset option is used to pass in **a file**?  

> Must support both local files and URLs

If not, then this is for sure. I'm pretty confused to be honest... I don't see how I missed the point in this PR.

> This is going to need a rewrite sorry, we chatted about it on Slack. It’s basically letting users define custom rules in a config file, which is basically what rulesets are.

Well yes, it's letting users specify which rulesets they can load by allowing them to specify a **ruleset file** (not **config file** btw, I haven't implemented loading config files).

> Instead we need to add a CLI option to enable already defined rulesets.
> _(from Slack conversation)_ They define the rulesets in the rulesets file (already done) then they enable them in the CLI (this new feature)

Maybe that's where my confusion comes from the most. What do you mean "they define ruleset in rulesets file (already done)"?

How is that done?  I guess I'm missing something but currently the way we load specs in CLI is the following:

```ts
  if (spec.data.swagger && spec.data.swagger === '2.0') {
    command.log('OpenAPI 2.0 (Swagger) detected');
    spectral.addFunctions(oas2Functions());
    spectral.addRules(oas2Rules());
  } else if (spec.data.openapi && typeof spec.data.openapi === 'string' && spec.data.openapi.startsWith('3.')) {
    command.log('OpenAPI 3.x detected');
    spectral.addRules(oas3Rules());
  } else {
    throw new Error('Input document specification type could not be determined');
  }
```

I don't see how we store rulesets in "ruleset files". We merely export some rules collections in some TS files (but it's nothing close to supporting "ruleset files").

Also, I don't understand "Instead we need to add a CLI option to enable already defined rulesets.". What do you mean? We only have three rulesets "defined": common, oas2 and oas3. But they don't have _names_, they are simply collections of rules and we arbitrarily call them oas2, oas3 etc. 

Can you please give specific examples?

I'm happy to rewrite if letting users specify custom rulesets was not the intention but we need to clearly define what the intention is first then.

chris-miaskowski:
Had a long call with @philsturgeon and we've cleared up the requirements.
Good news is: this does not need a rewrite but some additional features. Most of the confusion was coming from naming and dynamic VS static loading.

**Definitions**

**ruleset** - a structure holding a collection of rules **(rules collection)** and **functions** definitions.
A ruleset can define either **rules collection**, **functions** or both.

**Not in scope**

What @philsturgeon wanted to do on top of loading static _yaml/json_ rulesets is loading _ts/js_ files.
For now we will limit the scope to _yaml/json_ because the only advantage of a _typescripe rule collection_ is the fact it can use Enums rather than strings.

Loading _functions_ will also not be in the scope of this story.

**TODO**

1. Keep the static json/yaml files (rulesets)
2. Implement **require** function - ruleset A can **require** ruleset B (and merge)
3. Move oas2/oas3.ts collections into yaml/json
4. **Do not** autodetect oas2/3 if CLI option provided
5. Add definitions to README
6. Rename "config file" to "ruleset file"

philsturgeon:
Great summary @chris-miaskowski!

To add a bit of context to "Do not autodetect oas2/3 if CLI option provided". 

In Spectral 2.0 we autodetect which ruleset to apply, and thats done in the code Chris posted a few comments ago: 

```
if (spec.data.swagger && spec.data.swagger === '2.0') {
```

When we add this new `--ruleset=` option in 2.1.0 we want to avoid automatically shoving our opinions in as well as the opinions of the user defined in the ruleset files they are adding in. When we convert the existing oas2 and oas3 rulesets to YAML and add the `require` feature to these rulesets, users will be able to define custom rulesets which include our oas2 or oas3 rulesets _if they want to._

The only downside to this is that they will not be able to load both oas2 and oas3 (schema validation would clash) and they will lose the autodetect logic. This _might_ be fine, we will have to see how our users feel about it when they get the new feature. 

Autodetect will remain by default, so this is not a breaking change. When users start using `--ruleset` they will then lose the autodetect functionality, and this will be documented clearly. 

It's possible we can follow up with a 2.2 feature later which merges oas2 and oas3 into just "oas" and there is some sort of filter in the rules. We can talk about this on #145.

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
+# Configuring Spectral
+
+## Example config file (spectral.yaml)
+
+```yaml
+rules:
+    - rule-name:
```
XVincentX: Why is this an array objects? I'm assuming the `rule-name` is unique, so we can do something like

```yml
rules:
  rule-name:
    option1: true
  rule-name2:
    option1: true
```

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+        given: $..parameters[*]
+        then:
+            field: description
+            function: truthy        
+```
+
+## Rules
+
+Rules are highly configurable. There are only few required parameters but the optional ones gives powerful flexibility.
```
XVincentX: If the required parameters are `few` I'd call them out here.

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+    field: string;
+
+    // a regex pattern
+    pattern?: string;
+  };
+
+  then: IThen<T, O> | Array<IThen<T, O>>;
+}
+
+export interface IThen<T, O> {
```
XVincentX: Rant: I really do not like this rule that all the interface should start with `I`.

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+  when?: {
+    // the `path.to.prop` to field, or special `@key` value to target keys for matched `given` object
+    // EXAMPLE: if the target object is an oas object and given = `$..responses[*]`, then `@key` would be the response code (200, 400, etc)
+    field: string;
+
+    // a regex pattern
+    pattern?: string;
+  };
+
+  then: IThen<T, O> | Array<IThen<T, O>>;
```
XVincentX: I'd leave this as just an array and let `ajv` or somebody else coerce it as an array for us.
```suggestion
  then: Array<IThen<T, O>>;
```

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Rules
+
+Rules are highly configurable. There are only few required parameters but the optional ones gives powerful flexibility.
+
+TBD: replace TS interfaces with a nice table.
+
+```ts
+export interface IRule<T = string, O = any> {
+  type?: RuleType;
```
XVincentX: Why is this optional?

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+  // A short summary of the rule and its intended purpose
+  summary?: string;
+
+  // A long-form description of the rule formatted in markdown
+  description?: string;
+
+  // The severity of results this rule generates
+  severity?: DiagnosticSeverity;
```
XVincentX: Why is this optional?

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+  // Tags attached to the rule, which can be used for organizational purposes
+  tags?: string[];
```
XVincentX: I'd rather *not* make it optional. Provide `[]` if you do not want to provide any tag?

docs/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+  // should the rule be enabled by default?
+  enabled?: boolean;
```
XVincentX: I know I always say the same things, but a boolean is true or false. Nothing else.

src/cli/commands/lint.ts
```diff
@@ -70,20 +78,15 @@ linting ./openapi.yaml
...
+  const spec: IParserResult = await readParsable(name, flags.encoding);
 
   const spectral = new Spectral();
-  if (obj.data.swagger && obj.data.swagger === '2.0') {
+  if (spec.data.swagger && spec.data.swagger === '2.0') {
     command.log('OpenAPI 2.0 (Swagger) detected');
     spectral.addFunctions(oas2Functions());
     spectral.addRules(oas2Rules());
-  } else if (obj.data.openapi && typeof obj.data.openapi === 'string' && obj.data.openapi.startsWith('3.')) {
+  } else if (spec.data.openapi && typeof spec.data.openapi === 'string' && spec.data.openapi.startsWith('3.')) {
```
XVincentX: You can use `parseInt` instead of string comparison, it's going to detect anything `3.x.y`

```bash
> parseInt('3.1.2')
3
```

src/rulesets/spectral/ruleSchema.ts
```diff
@@ -0,0 +1,35 @@
...
+      array(
+        object({
+          field: optional(primitive('string')),
+          function: primitive('string'),
+          functionOptions: optional(primitive('any')),
+        })
+      ),
+    ],
+  },
+};
```
XVincentX: Is it possible to clarify what is going on here?

Apparently we're rolling out our own validation mechanism instead of using a JSON Schema for this?

docs/configuration.md
chris-miaskowski: Yeah, this readme is BS :) I have to update it. You're completely right about it, it's never even meant to be an array, it's a dictionary.

docs/configuration.md
chris-miaskowski: I will prepare a table describing Rule (similar to what you can find in OAS documentation)

docs/configuration.md
chris-miaskowski: What can I say :) "de gustibus non est disputandum" ;)

docs/configuration.md
chris-miaskowski: That's actually a copy paste from the types.ts file. I don't think there is much AJV can help us with here.

docs/configuration.md
chris-miaskowski: Out of scope, copy paste from types.ts

I'm happy to discuss this but it's not the main purpose of this PR

src/rulesets/spectral/ruleSchema.ts
chris-miaskowski: Yup, I thought it would be fun to lint spectral rules using spectral rules :P 
I'm cool switching to AJV (or any other validator) though

src/fs/reader.ts
```diff
@@ -0,0 +1,27 @@
...
+    const result = await fetch(name);
+    return parseWithPointers(await result.text());
+  } else if (existsSync(name)) {
+    try {
+      return parseWithPointers(readFileSync(name, encoding));
+    } catch (ex) {
+      throw new Error(`Could not read ${name}: ${ex.message}`);
+    }
+  }
+  throw new Error(`${name} does not exist`);
```
mallachari: How about check for valid yaml file and one more exception thrown? Parser should return diagnostics array with exceptions.

chris-miaskowski:
> @chris-miaskowski I'd love to check this out. How do I run it before it is merged?

Hey @tbarn, try this

1. clone repository and switch to this branch
2. `yarn && yarn build && npm link`
3. `spectral file.yaml --ruleset=path/to/ruleset.yaml`

__mocks__/fs.ts
```diff
@@ -1,6 +0,0 @@
-const fs = require.requireActual('fs');
-
-module.exports = {
-  ...fs,
-  writeFile: jest.fn((source: unknown, opts: unknown, cb: Function) => cb(null)),
```
P0lip: any reason to remove the mock? I don't see `fs.writeFile` mocked anymore and we do make such an operation in CLI related tests (output)

src/types/rule.ts
```diff
@@ -62,6 +62,11 @@ export interface ILengthRuleOptions {
 }
 export type LengthRule = IRule<RuleFunction.LENGTH, ILengthRuleOptions>;
 
+export interface IEnumRuleOptions {
+  values: Array<'string' | number>;
```
P0lip: Shouldn't this be `string`?
```suggestion
  values: Array<string | number>;
```

src/rulesets/oas/__tests__/__snapshots__/parameter-description.ts.snap
```diff
@@ -0,0 +1,87 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`parameter-description return errors if operation level parameter description is missing 1`] = `
+Array [
```
P0lip: Snapshots are a nightmare in Spectral.
When I was changing the diagnostics I had to update about ~200 (or more) snapshots...
I know all the other rules have similar assertions, so I don't oblige you to change anything, but I just wanted to point it out.

src/fs/reader.ts
P0lip: I don't think it's needed given the fact we expose parsing errors in both CLI mode and programmatic.
IHMO we shouldn't bail out if there is a syntax error, but we def need to display error messages in some way - as mentioned, it's already implemented, so I think we are good.

__mocks__/fs.ts
chris-miaskowski: it was a mistake

docs/configuration.md
chris-miaskowski: that's done

docs/configuration.md
chris-miaskowski: Not in scope, sorry

src/rulesets/spectral/ruleSchema.ts
chris-miaskowski: done

src/types/rule.ts
chris-miaskowski: good point!

src/rulesets/oas/__tests__/__snapshots__/parameter-description.ts.snap
chris-miaskowski: Good point but I don't think it has anything to do with snapshot technically. If we didn't have snapshots you'd need to update 200 tests. I'm not sure why but I'm guessing that's because we are not mocking stuff in unit tests (or write too many integration tests?)

src/fs/reader.ts
chris-miaskowski: Yup, I agree with JR :)

docs/configuration.md
```diff
@@ -0,0 +1,165 @@
...
+/rules/rule-with-invalid-enum/type 	 should be equal to one of the allowed values 
+```
+
+**Note for developers**
+
+Supporting YAML and JSON file validation doesn't come free. 
+We need to maintain schema files that mirror IRule and IRuleset types (see `/src/meta/*.schema.json`).
+Ideally, we would have a script that converts TS type to JSON Schema and keeps the meta files up to date. As of now we have a helper that partially automates the work.
+
+Execute `yarn schema.  update` to recreate the `/src/meta/rule.schema.json`.
```
philsturgeon: Is there some funky formatting happening here?

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -54,7 +57,49 @@ describe('lint', () => {
...
+      .command(['lint', validSpecPath, '-r', 'non-existent-path'])
+      .exit(2)
+      .it('outputs "does not exist" error');
+
+    test
+      .stdout()
+      .command(['lint', validSpecPath, '-r', invalidRulesetPath])
+      .exit(2)
+      .it('outputs "invalid ruleset" error', ctx => {
+        expect(ctx.stdout).toMatchSnapshot();
```
philsturgeon: Hey this might just be one persons opinion but I really don’t like match snapshot in 99% of cases. If it’s small can we use inline snapshots or better yet just use expects?

src/cli/commands/lint.ts
```diff
@@ -1,20 +1,14 @@
...
-import { existsSync, readFileSync, writeFile } from 'fs';
+import { getLocationForJsonPath } from '@stoplight/yaml';
 import { resolve } from 'path';
-import { promisify } from 'util';
-
-// @ts-ignore
-import * as fetch from 'node-fetch';
-
-import { json, stylish } from '../../formatters';
+import { readRuleset } from '../../config/reader';
```
philsturgeon: Is the folder /config/ maybe a layover from the old naming? Can this be rulesetReader or something?

src/cli/commands/lint.ts
```diff
@@ -69,33 +78,32 @@ linting ./openapi.yaml
...
-  } else if (obj.data.openapi && typeof obj.data.openapi === 'string' && obj.data.openapi.startsWith('3.')) {
+  } else if (parseInt(spec.data.openapi) === 3) {
     command.log('OpenAPI 3.x detected');
     spectral.addFunctions(oas3Functions());
     spectral.addRules(oas3Rules());
   } else {
     throw new Error('Input document specification type could not be determined');
   }
 
+  if (customRuleset) {
```
philsturgeon: Can we think of ruleset a more like “did they tell us what ruleset to use? No? Fine, we’ll take a guess”? Right now this is doing both. Remember if they provide a ruleset we don’t want to add oas2/3 automatically (they can require that later).

docs/configuration.md
```diff
@@ -0,0 +1,165 @@
+# Configuring Spectral
```
philsturgeon: Can we make it clear that this is about writing ruleset and not config? That’s a different thing to be done later.

src/cli/utils/output.ts
```diff
@@ -0,0 +1,25 @@
+import { writeFile } from 'fs';
+import { promisify } from 'util';
```
philsturgeon: It feels a little funny to make these generic functions when they are related to the lint command only. They had generic names because they didn’t need to be called formatLintOutput. 

Could we either put them back, or instead putting these in a util/ folder we put them in lint/?

src/cli/commands/lint.ts
```diff
@@ -50,16 +44,31 @@ linting ./openapi.yaml
...
 
   public async run() {
     const { args, flags } = this.parse(Lint);
+    const { ruleset } = flags;
+    let ruleCollection;
+
+    if (ruleset) {
+      try {
+        this.log(`Reading ruleset`);
+        ruleCollection = await readRuleset(ruleset, this);
```
philsturgeon: RuleCollection can just be rules we don’t need to introduce this new term. It’s just a dictionary of rules.

docs/configuration.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+### Ruleset validation
+
+We use JSON Schema & AJV to validate your rulesets file and help you spot issues early.
+
+**Example output**
+
+```bash
+spectral lint some-oas.yaml --ruleset invalid-ruleset.json
```
philsturgeon: This example could be “acme-company.yaml” to make it clear what it’s for. Rulesets for rulesets sake won’t excite people.

src/config/validation.ts
```diff
@@ -0,0 +1,15 @@
+import AJV = require('ajv');
```
philsturgeon: Another “config” layover. This PR should not have the term “config” in it _anywhere_. 😅

src/rulesets/oas/__tests__/__snapshots__/parameter-description.ts.snap
philsturgeon: Maybe maybe not. Snapshots contain everything, where as assertions only contain what you care about asserting. If an arbitrary property changes in 200 snapshots that sucks, but you might not have had to update 200 tests.

See other comment about snapshots. 
1. We don’t want to hide our assertions from the test file it makes it hard to read 
2. We don’t want our tests relying on a file which can change and accidentally get committed in a huge PR leading to false positives

Let’s just use expects :)

docs/configuration.md
chris-miaskowski: yup, see the preview or raw, it's fine

src/cli/commands/lint.ts
chris-miaskowski: Nope, I intentionally left it in `config` folder. My opinion is that loading rulesets is part of "configuring" spectral. The `reader` file can be extended with `function readConfigFile(` for instance. 

We don't really follow any file naming convention at the moment so both `rulesetReader/read` and `config/readRuleset` are equally good but judgemental.

src/cli/commands/lint.ts
chris-miaskowski: Yup, I remember, I wanted to do it in next PR though. Sounds good?

docs/configuration.md
chris-miaskowski: My intention was to store all that stuff in this file.
Setting custom rulesets _is_ imo configuration. On top of that there will be other stuff like config file etc. I tried to split this file into sections to make that clear. Is it not?

src/cli/utils/output.ts
chris-miaskowski: oh yeah, sorry I had some code before that used this and I extracted it :P I'm going to move it back.

src/cli/commands/lint.ts
chris-miaskowski: It's not a new term ;) It's exactly how the type we return is called, no? I'm not attached to this name thought, I'm happy to change it if bothers you.
`export type RuleCollection = Dictionary<Rule, string>;`

docs/configuration.md
chris-miaskowski: lol yeah good idea :)

src/config/validation.ts
chris-miaskowski: Like I said: I think setting custom rulesets _is configuration_. What do you want me to change? folder name? To what? :) `rulesets` is already taken ;)

src/rulesets/oas/__tests__/__snapshots__/parameter-description.ts.snap
chris-miaskowski: Right, it looks like it's a very delicate subject matter and mostly subjective ;)

I'm on the edge with snapshots personally. I try to avoid them to (like you said) increase visibility, but at the same time I disagree that they 'assert too much'. I actually think the opposite: 'expects.equalsTo' asserts too little.

E.g. 

>  If an arbitrary property changes in 200 snapshots that sucks,

You say it sucks, I say it's a bug that we missed to cover ;)

Anyway, I will see if I can reasonably ditch some of the snaphots but I'd like us to have a proper discussion about this - we need clear guidelines otherwise we'll keep arguing  about this in every PR ;)

chris-miaskowski:
@philsturgeon , @P0lip , @XVincentX , @mallachari thanks for great review guys! Really appreciate the effort :) I will address the comments and get back.

Can everyone leave their _final_ review before I start addressing? Will save me a couple of roundtrips ;)

@philsturgeon see my comments about `config` and let's decide.

src/rulesets/oas/__tests__/__snapshots__/parameter-description.ts.snap
philsturgeon: A bug that we missed to cover? No, im saying that if a test cares about only part of the response, and a completely unrelated property changes, then that is not something which should be causing a failure.

An assertion should only be asserting what it cares about. Instead of saying "I am interested in seeing that A, B and C are included in this object" you say "I can about this object matching a snapshop" and that giant snapshot has "A, B, C, D, E, F, G, H, I, 345, Tuesday, Ice Cream...." then you have a test that is 1.) unreadable, 2.) vague to the point of being confusing.

You can use https://jestjs.io/docs/en/expect.html#expectobjectcontainingobject to make awesome readable tests that are clear what they are asserting. MatchSnapshot is a nuclear option. 

We will make style guides one day, but for now I am just seeing a lot of people saying please stop using them, so that doesn't seem like much of an issue, lets just stop using them. :)

src/cli/commands/lint.ts
philsturgeon: Conceptually rulesets and config have no relationship or connection to each other at all. Config is just another form of input like CLI options. Please can we keep the two ideas separate.

src/cli/commands/lint.ts
philsturgeon: Thew require is slated to be done later absolutely, but I am talking about the fact that right now it is autoloading oas2/oas3 and then also `--ruleset=foo`. 

We should not add a feature to master and then change it later. The precedence is:
1. Load the rulesets if provided
2. Else autodetect

You just need to move the autodetect a few lines down.

src/cli/commands/lint.ts
philsturgeon: Generally we don't seem to be using Hungarian notation so it seems a bit out of place. It's not like we have rulesetDictionary and nameStr. 😅

src/config/validation.ts
philsturgeon: Already taken? That just means you can put the files inside without having to make a new directory.

src/rulesets/oas/__tests__/__snapshots__/parameter-description.ts.snap
XVincentX: Me and @P0lip, during a personal conversation on Slack, came out with the perfect implementation of snapshots:

carbon

P0lip: I'd say snapshots are only fine if the content is not meant to change, i.e. you have a parser, babel plugin etc. You usually don't expect the content to change much + everything is very important in such case + it's almost impossible (well, it's just unreadable) to use some regular assertion.

We def should come up with some guideline for that as indeed these discussions take place quite often and we've never reached any agreement.

chris-miaskowski: Cool, I will take a look and minimize snaphots ;)

src/cli/commands/lint.ts
chris-miaskowski: Coolio, yup 
> We should not add a feature to master and then change it later. 
you bought me with this argument. Will fix.

src/cli/commands/lint.ts
chris-miaskowski: lol I didnt realize that it's actually a THING :D 
Glad to learn that it's called hungarian notation :) 

will change!

src/cli/commands/__tests__/lint.test.ts
chris-miaskowski: done already, marking as resolved

src/cli/commands/lint.ts
chris-miaskowski: will move everything currently stored in `config/` folder into `ruleset/` folder.
I want to keep it separate from `rulesets/` (plural) since the latter is used for storing actual rulesets (oas2/3) , rather than ruleset loading logic

chris-miaskowski: done `rulesets`

chris-miaskowski:
:hankey: I've just realized that I've been addressing all requested changes in my `feature/sl-2295/sl-2470/add-extend-ruleset-feature` branch...

Meh. I guess I'm too lazy to revert this so I'll just merge everything here and we'll have one big-ace pull request.

chris-miaskowski:
@philsturgeon @XVincentX @mallachari 
everything addressed
I have merged the `extends` feature in so please re-review.

src/cli/commands/lint.ts
chris-miaskowski: done

docs/rulesets.md
```diff
@@ -0,0 +1,168 @@
...
+/rules/rule-with-invalid-enum/type 	 should be equal to one of the allowed values 
+```
+
+**Note for developers**
+
+Supporting YAML and JSON file validation doesn't come free. 
+We need to maintain schema files that mirror IRule and IRuleset types (see `/src/meta/*.schema.json`).
+Ideally, we would have a script that converts TS type to JSON Schema and keeps the meta files up to date. As of now we have a helper that partially automates the work.
+
+Execute `yarn schema.  update` to recreate the `/src/meta/rule.schema.json`.
```
philsturgeon: This yarn schema.   update still looks funny to me, what is it?

src/rulesets/oas/__tests__/parameter-description.ts
```diff
@@ -78,7 +78,26 @@ describe('parameter-description', () => {
         },
       },
     });
-    expect(results).toMatchSnapshot();
+    expect(results).toEqual([
+      {
+        code: 'parameter-description',
```
philsturgeon: This is great! If you wanted to make this a bit lighter to read (and going back to the conversation about only asserting the important and relevant bits) then you could do:

expect(results).toEqual([expects.objectContaining({

then keep the code, message, path, etc and maybe kill the range and anything else you don’t really care about. Not a blocked and if you decide you care about them all them keep them all, just letting you know it’s a thing. I was so happy when jakub showed me this 😬

docs/rulesets.md
chris-miaskowski: It's a helper. It will take the `IRule` type from `types.ts` file and automatically update `rule.schema.json` file (the JSON Schema represantation I use to validate rulesets). It can be done manually but this script is pretty helpful. There shouldn't be a whitespace here though.

src/rulesets/oas/__tests__/parameter-description.ts
chris-miaskowski: Sure, will do. The reason I didn't remove the `range` is that I think it's all important data to check (e.g. someone may rely on this information).

philsturgeon: Right thats what im saying, include the things you want, if you want specifically range then put that in the spec title, right? The test is asserting that when a problem happens, it will contain a range with these values. If thats what you want then thats the test. Anything you dont care about, get it outta there. :)

src/rulesets/reader.ts
```diff
@@ -0,0 +1,36 @@
+import merge = require('lodash/merge');
+import Lint from '../cli/commands/lint';
+import { formatAjv } from '../formatters/ajv';
```
philsturgeon: Is this added as a formatter because people are expected to also maybe use it like `spectral lint foo.yaml -f ajv` or this just for us? 

If it is just for us, can it not be a formatter? This seems to be slightly confusing the purpose of what formatters are currently doing.

src/rulesets/validation.ts
```diff
@@ -0,0 +1,15 @@
...
+import { ErrorObject } from 'ajv';
+import ruleSchema = require('../meta/rule.schema.json');
+import rulesetSchema = require('../meta/ruleset.schema.json');
+import { IRuleset } from '../types/ruleset';
+
+const ajv = new AJV({ allErrors: true, jsonPointers: true });
+const validate = ajv.addSchema(ruleSchema).compile(rulesetSchema);
+
+export function validateRuleset(ruleset: IRuleset): ErrorObject[] {
+  if (validate(ruleset)) {
```
philsturgeon: This seems to be doing three things? If its valid, then return [], otherwise try and return errors but also maybe return []. How is something going to be invalid but not have any errors? Is this just some hackery to make TS happy?

chris-miaskowski: > Is this just some hackery to make TS happy?

yup, errors can be `undefined`

src/rulesets/reader.ts
chris-miaskowski: Gotcha, it's formatting the `ruleset` errors output. I don't mind renaming it. Any suggestions?

src/rulesets/validation.ts
philsturgeon: Wait why?!

chris-miaskowski: Because that's the return type :)

chris-miaskowski:
@philsturgeon 
hmm I think it wont trigger the semantic release: https://github.com/stoplightio/spectral/commit/4670f54f988b41a33a897308dfff33561d49bb3c

look at the commit title `Feature/sl-2295/enable custom ruleset via cli (#144)`
should be `feat: something`, right?

philsturgeon:
It's fine we don't need this released, you have further improvement PRs coming right?

PR #154, opened 2019-04-18
Feature/sl-2295/sl-2470/add extend ruleset feature

chris-miaskowski:

### Other information


src/cli/commands/__tests__/__fixtures__/ruleset-extends-invalid.yaml
```diff
@@ -0,0 +1,11 @@
+extends:
```
philsturgeon: Extends is a bit overloaded with expectations from OOP, could we stick with "require" like we discussed in the issues and everywhere?

chris-miaskowski: Hey, I'm merging this to the `--ruleset` branch.
Let's continue discussion over there. I'm not sure about require. If you look at the implementation I actually do extend (merge) rulesets. So `require` is a bit misleading (isn't it?).

PR #157, opened 2019-04-19
Feature/SO-106/add config file support

mallachari:
Work in progress PR adding config file support.
Open for suggestions.


package.json
```diff
@@ -62,8 +62,10 @@
     "@stoplight/json-ref-resolver": "1.x.x",
     "@stoplight/types": "4.1.x",
     "@stoplight/yaml": "2.x.x",
+    "@types/js-yaml": "^3.12.1",
```
P0lip:
```suggestion
```

package.json
```diff
@@ -62,8 +62,10 @@
...
     "ajv": "6.x.x",
     "chalk": "^2.4.2",
+    "js-yaml": "^3.13.1",
```
P0lip:
```suggestion
```
I believe we can use `parseWithPointers` function exposed by `@stoplight/yaml` package. It's already installed and used by Spectral.

src/cli/commands/lint.ts
```diff
@@ -150,3 +167,19 @@ async function readInputArguments(name: string, encoding: string) {
   }
   throw new Error(`${name} does not exist`);
 }
+
+function mergeConfig(config: IConfig, flags: any): IConfig {
+  return merge(
```
P0lip: If I follow the code correctly the object is very likely to be 'flat', so we can using `Object.assign` or object spread properties might be a more suitable solution here.

```ts
return {
  ...config,
  ...omitBy<IConfig>(
      {
        encoding: flags.encoding,
        format: flags.format,
        output: flags.output,
        maxResults: flags.maxResults,
        verbose: flags.verbose,
      },
      isNil
  ),
};
```

src/config/configLoader.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (existsSync(filePath)) {
+    return yaml.safeLoad(readFileSync(filePath, 'utf8')) || {};
+  }
+  throw new Error(`${filePath} does not exist`);
+};
+
+const loadConfig = (filePath: any): IConfig => {
+  let config;
+
+  switch (path.extname(filePath)) {
```
P0lip: As stated earlier, we can utilize `parseWithPointers` function (it handles JSON just fine).

```ts
import { parseWithPointers } from '@stoplight/yaml';

const loadConfig = (filePath: string): IConfig => {
  try {
    const { data } = parseWithPointers(readFileSync(filePath, 'utf8'));
    // validate data if needed and throw if it's entirely iirrelevant
    return data;
   } catch (ex) {
     // rethrow some common error here, like InvalidConfig or whatever is already defined.
   }
}
```

package.json
mallachari: Yeah, I can go with it, thanks!

src/cli/commands/lint.ts
mallachari: Right, so far it's flat and spread could be better. I just need to be sure that it being flat is what is expected. @philsturgeon

philsturgeon: Flat sounds good, we can get fancy later

src/cli/commands/lint.ts
```diff
@@ -50,16 +51,32 @@ linting ./openapi.yaml
...
   };
 
   public static args = [{ name: 'source' }];
 
   public async run() {
     const { args, flags } = this.parse(Lint);
+    const { config: configFileFlag } = flags;
+    let config: IConfig = mergeConfig(createEmptyConfig(), flags);
+
+    const configFile = configFileFlag || getDefaultConfigFile(process.cwd()) || null;
```
P0lip: I'd say we should follow ESLint / TSLint and general ecosystem (i.e. Babel) a bit more here.
tl;dr

- package.json field, i.e. `spectral`
- `.spectralrc`, `.spectral.yaml`, `.spectral.json`
- js files should be supported for easier merging - environment based configuration is way nicer to implement
- file relative configuration is optional, we don't need it now

https://eslint.org/docs/2.0.0/user-guide/configuring#using-configuration-files
https://babeljs.io/docs/en/config-files#configuration-file-types

@philsturgeon thoughts?

philsturgeon: We don't want to rely on package.json as this is usable via a binary (not always installed via npm or yarn).

We can only use .spectralrc if its some special file type, so lets stick with `.spectral.yaml` and `.spectral.json`. 

"js files should be supported for easier merging" What do you mean about merging? There should not be any merging in config files.

P0lip: > We don't want to rely on package.json as this is usable via a binary (not always installed via npm or yarn).

Yeah, that's why it'd be optional. There are plenty of projects that tend to configure tools using package.json, so this would be nice to have, but we can skip it for now.

> "js files should be supported for easier merging" What do you mean about merging? There should not be any merging in config files.

True, I wasn't clear enough.
In our case, it would be something like

```ts
verbose: !!process.env.CI
```

Nothing crucial at the very moment, since our configuration isn't very complex yet, but that would be a nice addition in the future once we add more options.

philsturgeon: Gotcha oin the package stuff, but right now one way is ok, don't need to make two ways. We are just making a config file.

I think that JS way might be overthinking things a bit, people can combine a config file with CLI options, they would just use `lint --verbose` in their CI script.

P0lip: Sounds good!
In general, this is more food for thoughts. Our configuration is still fairly limited, so we can simply revisit the above ideas if we need to make it more robust one day.

src/config/configLoader.ts
P0lip: Are you still planning to address this, ya'all?
@mallachari @philsturgeon

philsturgeon: Sure lets do that.

package.json
```diff
@@ -63,8 +63,10 @@
     "@stoplight/json-ref-resolver": "1.x.x",
     "@stoplight/types": "4.1.x",
     "@stoplight/yaml": "2.x.x",
+    "@types/js-yaml": "^3.12.1",
```
chris-miaskowski:
```suggestion
    "@types/js-yaml": "3.12.x",
```

;) that's the SL convention

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -177,4 +180,63 @@ describe('lint', () => {
...
+          expect.any(Function) // callback, util.promisify handles it for us
+        );
+      });
+  });
+
+  describe('when using config file and command args', () => {
+    test
+      .stdout()
+      .command(['lint', invalidSpecPath, '-c', validConfigPath, '-m', '1'])
+      .it('outputs warnings in json format', ctx => {
```
chris-miaskowski: How about?
```suggestion
      .it('given maxResults set to 1 outputs warnings in json format', ctx => {
```

src/cli/commands/lint.ts
chris-miaskowski: I'm happy with keeping it simple for now, one default config file path should be enough for this story, we can add support to other paths later.

src/cli/commands/lint.ts
```diff
@@ -69,9 +74,21 @@ linting ./openapi.yaml
...
 
+    let config: ILintConfig = mergeConfig(createEmptyConfig(), flags);
+
+    const configFile = configFileFlag || getDefaultConfigFile(process.cwd()) || null;
+    if (configFile) {
+      try {
+        const loadedConfig = await loadConfig(configFile);
+        config = mergeConfig(loadedConfig, flags);
+      } catch (ex) {
+        this.error('Cannot load provided config file');
```
chris-miaskowski: I think it would be helpful to actually log the `ex` in there. That could give the developer a hint of what's happening.
```suggestion
        this.error('Cannot load provided config file', ex);
```

package.json
mallachari: Good to know, anyway it is not needed anymore so I removed it.

src/cli/commands/__tests__/lint.test.ts
mallachari: Much better

src/cli/commands/lint.ts
mallachari: done

mallachari:
Thanks for reviews! 
I have just added final ruleset support for config files. It can go either as a string or an array of paths.
Please take a look and let me know if you notice anything to correct.

src/config/configLoader.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (!isEmpty(diagnostics)) {
+        throw new Error('Parsing exception');
+      }
+      return data;
+    default:
+      throw new Error('Unknown file');
+  }
+};
+
+const extend = async (config: IConfig, filePath: string): Promise<IConfig> => {
```
philsturgeon: Hey what’s this extends stuff about? The only extends I know about is for rulesets.

src/config/__tests__/__fixtures__/config.invalid.yaml
```diff
@@ -0,0 +1,6 @@
+---
+{
```
philsturgeon: I guess this is double invalid, its mising the lint _and_ bad syntax. Is that intention?

src/config/configLoader.ts
mallachari: That's extension that was there since previous PR. I asked later if I should remove it but don't remember clear response. But I can get rid of it easily, just let me know.
One other thing that might be interesting for last call is the default file in cwd. Currently it recognizes only spectral.yml. I can add also json an yaml.

philsturgeon: Ok, let’s get rid of it. We gotta be careful about accidentally adding features without planning them, or we end up in tricky situations because somebody will use whatever is added and then we’re stuck with it. :)

mallachari: It's removed :)

src/config/__tests__/__fixtures__/config.invalid.yaml
mallachari: Right. I split it in separate test cases. I've also added validation for empty config and missing command.

PR #182, opened 2019-05-07
Feature/SO-170/convert oas2 and 3 rulesets to yaml

chris-miaskowski:

Feature

When using CLI we automatically detect oas2/3 files and add standard Stoplight rules based on the file type. We also allow users to provide their own ruleset files that can extend other ruleset files. However, our standard rules come in TypeScript file only and can't be extended from using the CLI ruleset feature.

This PR introduces three ruleset `yaml` files (oas.ruleset.yaml, oas2.ruleset.yaml, oas3.ruleset.yaml) that can be generated and kept up to date using the `yarn build.rulesets` command. These files mirror our TypeScript rules.

No


scripts/generate-rulesets.ts
```diff
@@ -0,0 +1,22 @@
+import { writeFileSync } from 'fs';
```
philsturgeon: This looks like we are keeping them as TS but converting to YAML programatically instead of just actually using the new YAML format. Is there a technical reason we cannot use the new ruleset format we made?

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -167,17 +170,46 @@ describe('lint', () => {
...
           expect(ctx.stdout).toContain('No errors or warnings found!');
         });
     });
+
+    describe('when a standard oas3 ruleset provided through option', () => {
+      test
+        .stdout()
+        .command(['lint', invalidOas3SpecPath, '-r', standardOas3RulesetPath])
+        .it('outputs warnings in default format', ctx => {
+          expect(ctx.stdout).toContain('Applying custom rules. Automatic rule detection is off.');
```
philsturgeon: I am seeing the word custom a lot and it feels funny. We do not want to give people the feeling that there is "normal" and "custom", there are just rulesets and rules. If you do not specify a ruleset, we will detect a default, but we don't need to keep saying "Woah there's a custom thing, this is pretty different!" 

Speccy would just tell you what rulesets it was trying to load and what rules it found, but didnt ever care if they were core or third party https://github.com/wework/speccy/blob/master/lib/loader.js#L133

scripts/generate-rulesets.ts
chris-miaskowski: Yes, it's basically convenience. 
I imagine that other teams who build their custom rulesets would do the same. They could use TS/JS/Java or whatever to build their rulesets. What matters is the end result, not the means IMO.

TS is convenient because it allows us to reuse variables and functions. It also makes conditional rendering possible. 

In the end we should not care how the yaml file is generated (think about an http yaml file that is automatically generated based on your cookie or whatever :) ).

src/cli/commands/__tests__/lint.test.ts
chris-miaskowski: Sure, I'm open to suggestions. Does `Applying third-party rules` sound good?

scripts/generate-rulesets.ts
chris-miaskowski: @XVincentX that's a good point. I think I will just remove the TS files and make sure the yaml files are in the build output.

src/cli/commands/lint.ts
```diff
@@ -105,16 +105,24 @@ async function lint(name: string, flags: any, command: Lint, customRules?: RuleC
...
   const spectral = new Spectral();
+  command.log('Setting up functions...');
+  if (parseInt(spec.data.swagger) === 2) {
+    command.log('Adding OpenAPI 2.0 (Swagger) functions');
+    spectral.addFunctions(oas2Functions());
+  } else if (parseInt(spec.data.openapi) === 3) {
+    command.log('Adding OpenAPI 3.x functions');
+    spectral.addFunctions(oas3Functions());
+  }
+
```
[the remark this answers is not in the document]
chris-miaskowski: Because you can't import functions using the `yaml` file. Imagine someone imported the oas3 ruleset using the `--ruleset` param. In such case we disabled the automatic detection and wouldn't import oas3 functions (w/o functions that file would fail to execute).

This fixes that issue (we always detect functions automatically but only detect rules if a ruleset file was not provided). We will be able to remove this code when we extend the `yaml` files with `functions` functionality.

scripts/generate-rulesets.ts
chris-miaskowski: @philsturgeon @XVincentX 
Ok after considering once again I think we should **not** remove the typescript files. We should provide YAML files and make sure they are up to date with the TS file.

Why? Because for maintainers of the repository it will be much much easier to modify TS files than the generated YAML files. 

Take the `oas3.ruleset.yaml` for instance. Even with the "common.ruleset.yaml" rules removed the file has ~2000 LOC compared to 100 LOC of the typescript file.

Most of the bloat in the generated YAML/JSON comes from the `schema` object (but not only). See below TS example that's converted to hundreds of lines of code.

```
        then: {
          function: RuleFunction.SCHEMA,
          functionOptions: {
            schema,
          },
        },
```

I think leaving **just yaml/json** is simply non-maintainable. Let's stick the TS files and generate json/yaml in build time.

chris-miaskowski:
For some reason (maybe a failed merge) it suddenly started complaining about tslint errors... I'm fixing all in this PR.

scripts/generate-rulesets.ts
XVincentX: Yeah, I think this should be approach!

philsturgeon: > Why? Because for maintainers of the repository it will be much much easier to modify TS files than the generated YAML files.

If this is how we feel about our YAML ruleset format then it means our YAML ruleset format is not very good. If we can't be bothered to use it, how can we expect others to use it? 

Maybe we need to dogfood, and that will help us improve our format?

chris-miaskowski: > If this is how we feel about our YAML ruleset format then it means our YAML ruleset format is not very good. If we can't be bothered to use it, how can we expect others to use it?

It's not much about the format itself but about the volume of the output file (it's a 3000 LOC file). 

It's doing a good job when passed into a CLI or when being extended (`extends: node_modules/@stoplight/oas2.ruleset.yaml`) but it's **not meant to be editable**. 

I agree with @XVincentX that this is basically a build artifact distributed in an npm package. What is editable is our TS files.

philsturgeon: I disagree that these yaml files should be considered a read only srtifact I believe they should be the source of truth and there are ways to make that happen. 

I’m not sure why these files are 3000 lines long? Are you bundling the common oas stuff into both oas2 and oas3? Speccy has “named rukesets” which lived in the core. You could “require: strict” and it would look in the core of speccy itself, then paths or URLs. This would solve our duplication issues if both just extends: oas, right?

chris-miaskowski: > I’m not sure why these files are 3000 lines long? Are you bundling the common oas stuff into both oas2 and oas3?

@philsturgeon it's not a matter of bundling oas.common into oas3 (I have done both: bundled and extended and each results if very long file). Like I said in one of my previous comments the problem with oas3 yaml file (for instance) comes from the fact we include the whole `schema` in the ruleset. Remember this? : 

```
functionOptions: {
            schema,
          },
```

It looks innocent but the `schema` is a big-fat object.

> I disagree that these yaml files should be considered a read only srtifact I believe they should be the source of truth and there are ways to make that happen.

I don't agree with you disagreement ;P To me it's like telling: code in assembler language because it's doable. Of course it's doable, but harder. If one (let's say a Google contributor) wants to keep their rules in SQL and generate their rules.json file using Kotlin.

The thing I don't understand is why should the yaml not be considered an artifact? Are you worried about usability? Stability? I don't see any issues here. And not saying that there isn't a problem here. Maybe there is and I don't see it, but the "I believe" argument doesn't tell me much (not being sarcastic here).

Maybe, just maybe (different topic) we should have allowed people to include TS files (vs YAML) in the first place but I still hold the opinion that YAML files are eventually much easier for the majority of people (e.g. yaml/json files don't need a transpiler).

>  Speccy has “named rukesets” which lived in the core. You could “require: strict” and it would look in the core of speccy itself, then paths or URLs.

That's a related but different topic, we don't have that functionality yet and this PR doesn't address that. We can discuss that later but that's outside of the scope of this story.

chris-miaskowski: @philsturgeon @XVincentX need you thoughts on that, let's keep in simple though because the discussion is becoming much longer and more complex than the implementation. 

My opinion for now is that we should keep generating YAML files from TS because maintaining 2000-3000 LOC files will be a problem. This PR solves the original request to make the rules "importable", so it meets the AC (and AFAIK Dojo waits for this feature, right?). We can debate the implementation but at this point I think it's better to 'test the hypothesis' rather than to try to anticipate the outcome.

philsturgeon: RE named rulesets, I only brought those up as a potential solution to avoid duplicate content in oas2 and oas3 rulesets, but if thats no an issue we don't need a solution for that issue. 

> This PR solves the original request to make the rules "importable", so it meets the AC

There are a whole bunch of reasons behind switching to using YAML for our core rules, one of them yes is making them importable, but they also serve as an example for others to work with. 

I think your comparison to assembly is a great example of the difference in our thinking. I do not want our YAML ruleset to be like assembly, I want it to be the language people want to work with. If its too hard to work with, then we have to make it easier. 

This point cannot be stressed enough as far as I'm concerned. If we are finding issues with our ruleset format then there is a problem with our ruleset format. We can't say it's fine, yet also have concerns that it is overly verbose due to common things like use JSON Schemas due to them needing to be inlined.

Do we need to create more functionality for our JSON/YAML rulesets like supporting $ref so people can reference schemas from files? If we can see a path to improving our ruleset format where we will no longer need to transpile, then I am ok with using transpiling as a temporary solution until the other functionality is complete. 

Btw there is not much need to test the hypothesis, speccy had YAML rulesets and a whole boatload of people were using those. Nobody ever asked for JS/TS transpiling. This feature is just meant to be recreating that functionality in Spectral so people can migrate to using our tool instead of that stagnant one. :)

chris-miaskowski: @philsturgeon ok, so let's settle it. You would rather have everything in large YAML files and then improve on them iteratively than to have them mixed?

I'm fine with that if you agree to have 2k LOC files but adding $refs and all that good improvements need to be put aside for now and we'll need to get back to it later. Does that sound good to you?

chris-miaskowski: Instead YAML I will use JSON because it's natively supported by TS and we can still include our rules with import (w/o having to setup webpack!)

philsturgeon: It's hard to agree to the massive YAML files when I can't even see them. I woud hope that they aren't so massive.

Goals: https://github.com/wework/speccy/blob/master/rules/default.yaml

The end goal for me is that we are using is the same approach that we are asking people to use, so that we avoid a double standard. 

If what you do in this PR is something we can document and recommend to our users, great!

If what you do is in this PR is a stepping stone towards that and we make some stories/issues to help track progress towards that, fair enough.

chris-miaskowski: Pasting from Slack:

Here is the caveat I've mentioned. See that when I include the json file (which is ok) TS "compiler" complains about types :disappointed:

Selection_386

And there reason it complains matters I think.
```
Types of property 'type' are incompatible.
      Type 'string' is not assignable to type 'RuleType | undefined'
```
So because `json` is just strings but the actual IRule has all that good stuff like enums, numbers etc we won't be able to use types anymore to make sure our rules are correct.
If you think that's ok and then I can remove TS files but probably I need to make the types less strict (e.g. replace enums with strings)

chris-miaskowski: Here is an example json file https://drive.google.com/open?id=1Nb6qjddIHLsaNAhNtNhfFCOYxhuzHD5n

philsturgeon: The example file is only 118 lines when you remove the inline JSON Schema, which is perfectly manageable. Seems like we just need to allow JSON Schema files to be referenced by path instead of inlined, which could be this PR or another.

chris-miaskowski: @philsturgeon 

> The example file is only 118 lines when you remove the inline JSON Schema, which is perfectly manageable. Seems like we just need to allow JSON Schema files to be referenced by path instead of inlined, which could be this PR or another.

I have already commented on the size of the file and we are on agreement that we can implement referencing schemas etc. See my other comment (the one about types).

XVincentX: I want to be part of this discussion. I just need to find the time to look into.

chris-miaskowski: @philsturgeon 

There is a quick way to implement it using dynamic `import` introduced in TypesScript 2.4
The caveat is that TS is a JS in the runtime :) so in the CLI we can import JS files.

It's a pretty simple change if we can agree to support JS files.

Selection_387

The above allows us to do:

```bash
spectral lint src/__tests__/__fixtures__/petstore.oas2.json --ruleset `pwd`/dist/rulesets/oas2/index.js
```

chris-miaskowski: Just verified it would also allows us to use in in TS code

Selection_388

philsturgeon: Right yeah we can just support JS and let folks know they should give us their built TS if they wanna use TS, fine with that.

Just checking that oas2Rules is part of the proof of concept only, and not part of the suggested API :D 

The JS file should be returning an array, or defining a function which returns an array, generally trying to look as similar to the YAML as possible.

chris-miaskowski: > Just checking that oas2Rules is part of the proof of concept only, and not part of the suggested API :D

lol yeah :) it's a PoC

chris-miaskowski: :hankey: :hankey: :hankey: 
I keep finding bumps I haven't thought about before all the time :rage: 

@philsturgeon @XVincentX 
So, when I was sure that using `import` will solve all our problems I realised it has a major flaw as soon as I started testing it...

He is the simplest case that shows that importing `js` (or `ts`) files will be tough.

Selection_389

Selection_390

So...
- importing JS files is easy when done locally, but difficult when done over the wire
- we can't simply say "yeah let's support local files only for now", because people need to be able to use our `extend: xyz` with remote files as well.
- we could make our `js` artifacts bundle all dependencies (but, yeah... no)

Ok @philsturgeon, back to your idea of **named rulesets**! :P 
Plan:
- we hardcode two named rulesets `@stoplight/oas2` & `@stoplight/oas3`, exact match them in the ruleset reader & **load the TS files**.
- we either ditch our YAML rulesets OR still include them in the `dist` package (so that people could tweak it (?)

package.json
```diff
@@ -39,10 +39,11 @@
     "node": ">=8.3.0"
   },
   "scripts": {
-    "build": "sl-scripts build && node ./scripts/prepare-cli",
+    "build": "sl-scripts build && node ./scripts/prepare-cli && yarn copy.rulesets --verbose",
```
XVincentX: Whenever is possible, separate the scripts; use `postbuild`.

src/cli/commands/lint.ts
```diff
@@ -109,28 +107,50 @@ linting ./openapi.yaml
   }
 }
 
+async function handleReaderError(command: Lint, reader: Function) {
```
XVincentX: I think this is not a great name; it's not handling the error, but it's rather trying to read the value and eventually handle the error.

```suggestion
async function tryReadOrLog(command: Lint, reader: Function) {
```
or something like that

src/rulesets/oas2/__tests__/api-schemes.ts
```diff
@@ -1,11 +1,11 @@
...
-import { oas2Rules } from '../index';
-
-const ruleset = { rules: oas2Rules() };
+import * as ruleset from '../ruleset.json';
 
 describe('api-schemes', () => {
   const s = new Spectral();
   s.addRules({
+    // @ts-ignore
+    // @ts-ignore
```
XVincentX: Double ts-ignore
```suggestion
```

yarn.lock
```diff
@@ -6152,7 +6447,121 @@ npm@6.5.0:
     worker-farm "^1.6.0"
     write-file-atomic "^2.3.0"
 
-"npmlog@0 || 1 || 2 || 3 || 4", "npmlog@2 || ^3.1.0 || ^4.0.0", npmlog@^4.0.2, npmlog@~4.1.2:
+npm@^6.8.0:
```
XVincentX: This should not be here I'd say

chris-miaskowski:
@XVincentX thanks for the comments, really appreciate. If you haven't read through the whole thing can you do your second round before I address your requests? Commenting -> addressing -> commenting -> addressing will take too much time. Throw all you got at me and I will implement everything in one go.

chris-miaskowski:
@XVincentX addressed everything.

Btw guys I hope you don't mind if I replace the ts-ignore with `as Rule` some other time? It's a refactoring that's nice to have but I'd rather boyscout it later.

src/__tests__/functions.test.ts
```diff
@@ -4,6 +4,7 @@ import { IRuleResult, Rule, RuleFunction } from '../types';
 const applyRuleToObject = async (r: Rule, o: object): Promise<IRuleResult[]> => {
   const s = new Spectral();
   s.addRules({
+    // @ts-ignore
```
XVincentX: This does not seem to be required.

image
```suggestion
```

src/__tests__/spectral.test.ts
```diff
@@ -36,6 +36,7 @@ describe('spectral', () => {
       const s = new Spectral();
 
       s.addRules({
+        // @ts-ignore
```
XVincentX: This is not required 
```suggestion
```

src/__tests__/spectral.test.ts
```diff
@@ -75,6 +76,7 @@ describe('spectral', () => {
         const s = new Spectral();
 
         s.addRules({
+          // @ts-ignore
```
XVincentX: This is not required
```suggestion
```

src/rulesets/oas/__tests__/info-contact.ts
```diff
@@ -1,11 +1,10 @@
 import { Spectral } from '../../../spectral';
-import { commonOasRules } from '../index';
-
-const ruleset = { rules: commonOasRules() };
+import * as ruleset from '../ruleset.json';
 
 describe('info-contact', () => {
   const s = new Spectral();
   s.addRules({
+    // @ts-ignore
```
XVincentX: So apparently the reason why Typescript is complaining is because of the `type` property, which is an enum, while the JSON file is kind of arbitrary.

There are two solutions for that; the first one is cheating a little bit on typescript:

```ts
  s.addRules({
    'info-contact': Object.assign(ruleset.rules['info-contact'], {
      enabled: true,
      type: RuleType[ruleset.rules['info-contact'].type]
    }),
  });
```

The second one is how we should start doing things with external input:

```ts
describe('info-contact', () => {
  const s = new Spectral();

  RTV.decode(ruleset.rules['info-contact'].type).map(p => {
    s.addRules({
      'info-contact': Object.assign(ruleset.rules['info-contact'], {
        enabled: true,
        type: RuleType[p]
      }),
    })
  }).mapLeft(e => expect(e).not.toBeDefined())
```

Let's go with the former for now, but let's think about the latter as well.

chris-miaskowski: Thanks! Some regexp wizardy made the refactoring easy, I will post an update in a sec.

chris-miaskowski:
@XVincentX I'm done, rereview pls

src/cli/commands/lint.ts
```diff
@@ -109,28 +107,50 @@ linting ./openapi.yaml
...
+    command.log('Adding OpenAPI 3.x functions');
+    spectral.addFunctions(oas3Functions());
+  }
+
+  command.log('Setting up rules...');
+  if (rules) {
     if (flags.verbose) {
       command.log(`Found ${Object.keys(rules).length} rules`);
     }
+    command.log('Applying rules. Automatic rule detection is off.');
```
philsturgeon: Can we get rid of this? It's taken care of in 146, and is wrapped in verbose flags.

src/cli/commands/lint.ts
```diff
@@ -109,28 +107,50 @@ linting ./openapi.yaml
...
   } else {
     if (flags.verbose) {
       command.log('No rules loaded, attempting to detect document type');
     }
     if (parseInt(spec.data.swagger) === 2) {
       command.log('OpenAPI 2.0 (Swagger) detected');
-      spectral.addFunctions(oas2Functions());
-      rules = oas2Rules();
+      rules = await tryReadOrLog(command, async () => await oas2Rules());
```
philsturgeon: Do we want to let spectral continue if the rules cannot be loaded? 

Should it error if it has 1 of 2 rulesets loaded? 
Should it error if it has 0 of 2 rulesets loaded?

I'd say yes for both. Not just a command.error but a process.exit too.

src/rulesets/reader.ts
```diff
@@ -1,34 +1,46 @@
...
 import { readParsable } from '../fs/reader';
 import { RuleCollection } from '../types';
 import { IRulesetFile } from '../types/ruleset';
 import { formatAjv } from './ajv';
 import { resolvePath } from './path';
 import { validateRuleset } from './validation';
 
-export async function readRulesets(command: Lint, ...files: string[]): Promise<RuleCollection> {
-  const rulesets = await Promise.all(files.map(file => readRuleset(command, file)));
+export async function readRulesFromRulesets(...files: string[]): Promise<RuleCollection> {
```
philsturgeon: Is this a more appropriate name? It is reading the full contents of the ruleset file, not just the rules section. It's reading the extend, and any other information we add to it.

package.json
```diff
@@ -40,9 +40,11 @@
   },
   "scripts": {
     "build": "sl-scripts build && node ./scripts/prepare-cli",
+    "postbuild": "yarn copy.rulesets --verbose",
     "build.binary": "yarn build && yarn build.manifest && cp yarn.lock dist && cd dist && yarn && pkg . --out-path ./bin",
     "build.docs": "sl-scripts build:typedoc",
     "build.manifest": "cd dist && oclif-dev manifest",
+    "copy.rulesets": "copyfiles -u 1 \"src/**/ruleset.json\" dist",
```
philsturgeon: Let's ditch these commands if we arent doing the TS>YAML build thingy.

src/cli/commands/lint.ts
philsturgeon: Ahh looks like this is the same behavior as before so this can be handled another time!

src/rulesets/oas/__tests__/contact-properties.ts
```diff
@@ -1,13 +1,12 @@
...
-const ruleset = { rules: commonOasRules() };
+import { RuleType, Spectral } from '../../../spectral';
+import * as ruleset from '../ruleset.json';
 
 describe('contact-properties', () => {
   const s = new Spectral();
   s.addRules({
     'contact-properties': Object.assign(ruleset.rules['contact-properties'], {
       enabled: true,
+      type: RuleType[ruleset.rules['contact-properties'].type],
```
philsturgeon: What is the cause of us needing this? It seems like a pain in the backside to need to do this? Is this some sort of TypeScript to force the type to be a valid RuleType? Is there any other way of doing this?

XVincentX: Yes, use a validation library (such as `io-ts`) or put `//ts-ignore` everywhere.

package.json
XVincentX: I think this is still required, we need to copy the rules in the destination directory we're publishing for npm, right? @chris-miaskowski

src/__tests__/functions.test.ts
XVincentX: I still see the comment here, we can remove this right?

src/rulesets/oas/functions/oasOpSecurityDefined/__tests__/index.ts
```diff
@@ -1,15 +1,18 @@
...
+const oas2Ruleset = { functions: oas2Functions(), rules: oas2Rules };
+const oas3Ruleset = { functions: oas3Functions(), rules: oas3Rules };
 
 describe('oasOpSecurityDefined', () => {
   describe('oas2', () => {
     const s = new Spectral();
     s.addFunctions(oas2Ruleset.functions || {});
     s.addRules({
+      // @ts-ignore
       'operation-security-defined': Object.assign(oas2Ruleset.rules['operation-security-defined'], {
```
XVincentX: You can avoid the tsignore using the same trick I described above, right?

src/rulesets/reader.ts
chris-miaskowski: But it returns rules, not the whole ruleset, so yes it's more appropriate imo.

package.json
chris-miaskowski: It's still required, TS does not copy JSON files for us, we need to do it manually.

src/rulesets/oas/__tests__/contact-properties.ts
chris-miaskowski: Marking as resolved, was using `@ts-ignore` before and we agreed with @XVincentX this TS trick is better for now.

package.json
philsturgeon: OOOOH because we put them in src, right.

src/cli/commands/lint.ts
```diff
@@ -109,13 +107,36 @@ linting ./openapi.yaml
...
+}
+
 async function lint(name: string, flags: any, command: Lint, rules?: RuleCollection) {
   if (flags.verbose) {
     command.log(`Linting ${name}`);
   }
   const spec: IParserResult = await readParsable(name, flags.encoding);
   const spectral = new Spectral();
-  if (rules !== undefined) {
+  command.log('Setting up functions...');
```
philsturgeon: Sorry I commented about this on other lines not each specific line, but can we make sure anything we log by default is truly crucial? This doesn’t seem like it’s adding enough value to be output by default, I’m not sure I’d even want it if —verbose was on.

P0lip: I have similar observations. We seem to log an insane amount of stuff. Most of that is irrelevant for end-user

chris-miaskowski:
Thanks for helping me resolve the conflicts Jakub!

src/cli/commands/lint.ts
chris-miaskowski: Fixed that. Some of the logs were there for a reason: namely we used those logs in tests (which I know is a bit silly).

philsturgeon:
Ugh. Ship. 😅

chris-miaskowski:
> Ugh. Ship.

Freudian slip! :P

PR #158, opened 2019-04-20
feat(cli): skip rules from cli

cg-cnu:
Added a flag "-s", which will ignore the given rules.

### Please check if the PR fulfills these requirements
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

Tests & Docs are pending. Waiting on initial input.

### What kind of change does this PR introduce?

feature

### What is the current behavior?

https://github.com/stoplightio/spectral/issues/147

### If this is a feature change, what is the new behavior?
You can use the flag -s valid-rule,invalid-rule to skip rules.

### Does this PR introduce a breaking change?
No.

PR #188, opened 2019-05-10
CLI Skip Rule

### LINKED ISSUE #147: Feature : Skip rule
### **I'm submitting a...**
  - [ ] bug report
  - [x] feature request

You will get all errors for everything in whatever ruleset, and there is no way to disable it. 

As a Spectral CLI user, sometimes I like a whole ruleset but a specific rule is really annoying me for some reason. I would like to be able to pass `--skip=valid-example` and spectral would not even bother running that. 

Plenty of folks have pointed out that the default ruleset is a bit more opinionated than they might like, so turning off a few errors temporarily whilst their team catch up on fixing them would help.

philsturgeon:

Fixes #147
Closes #158

Adds a new -s/--skip-rule switch that will allow you to ignore rules. 

It deprecates maxResults and removes it from documntation opting for --max-results instead. We should use kebab-case for-all-options instead of camelCase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Other options include:
   -h, --help                   show CLI help
   -m, --maxResults=maxResults  [default: all] maximum results to show
   -o, --output=output          output to a file instead of stdout
+  -s, --skip=skip              skip rules. Provide multiple rules seperated by ","
   -v, --verbose                increase verbosity
 ```
 
--- a/README.md
+++ b/README.md
@@ -79,13 +79,16 @@ spectral lint petstore.yaml
 Other options include:
 
 ``` text
-  -e, --encoding=encoding      [default: utf8] text encoding to use
-  -f, --format=json|stylish    [default: stylish] formatter to use for outputting results
+  -c, --config=config          path to a config file
+  -e, --encoding=encoding      text encoding to use
+  -f, --format=json|stylish    formatter to use for outputting results
   -h, --help                   show CLI help
-  -max=max                     [default: all] maximum results to show
+  -m, --maxResults=maxResults  deprecated: use --max-results instead
   -o, --output=output          output to a file instead of stdout
-  -s, --skip=skip              skip rules. Provide multiple rules separated by ","
+  -r, --ruleset=ruleset        path to a ruleset file (supports remote files)
+  -s, --skip-rule=skip-rule    ignore certain rules if they are causing trouble
   -v, --verbose                increase verbosity
+  --max-results=max-results    [default: all] maximum results to show
 ```
 
 > Note: The Spectral CLI supports both YAML and JSON.
--- a/README.md
+++ b/README.md
@@ -82,9 +82,9 @@ Other options include:
   -e, --encoding=encoding      [default: utf8] text encoding to use
   -f, --format=json|stylish    [default: stylish] formatter to use for outputting results
   -h, --help                   show CLI help
-  -m, --maxResults=maxResults  [default: all] maximum results to show
+  -max=max                     [default: all] maximum results to show
   -o, --output=output          output to a file instead of stdout
-  -s, --skip=skip              skip rules. Provide multiple rules seperated by ","
+  -s, --skip=skip              skip rules. Provide multiple rules separated by ","
   -v, --verbose                increase verbosity
 ```
 
--- a/src/cli/commands/lint.ts
+++ b/src/cli/commands/lint.ts
@@ -58,6 +58,10 @@ linting ./openapi.yaml
       description: 'path to a ruleset file (supports remote files)',
       multiple: true,
     }),
+    skip: flagHelpers.string({
+      char: 's',
+      description: 'skip rules. Provide multiple rules separated by ","',
+    }),
   };
 
   public static args = [{ name: 'source' }];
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -82,8 +82,9 @@ Other options include:
   -e, --encoding=encoding      [default: utf8] text encoding to use
   -f, --format=json|stylish    [default: stylish] formatter to use for outputting results
   -h, --help                   show CLI help
-  -m, --maxResults=maxResults  [default: all] maximum results to show
+  -max=max                     [default: all] maximum results to show
```
P0lip: I believe `maxResults` was a better choice, as `max` is too ambiguous and may become troublesome in future.
For instance, quite a few linters provide a way to set a maximum amount of warnings allowed.
Examples:

P0lip: I believe maxResults was a better choice, as max is too ambiguous and may become troublesome in future.
For instance, quite a few linters provide a way to set a maximum amount of warnings allowed.
Here is an example of such option
https://eslint.org/docs/user-guide/command-line-interface#--max-warnings

src/cli/commands/lint.ts
```diff
@@ -41,23 +45,28 @@ linting ./openapi.yaml
       char: 'o',
       description: 'output to a file instead of stdout',
     }),
+    max: flagHelpers.integer({
+      description: '[default: all] maximum results to show',
+    }),
+    // @deprecated in 2.2, remove in 3.0
     maxResults: flagHelpers.integer({
```
P0lip: If we decide it to drop it eventually, we should show a deprecation notice now.

src/cli/commands/lint.ts
```diff
@@ -100,26 +109,38 @@ linting ./openapi.yaml
...
-  } else if (parseInt(spec.data.swagger) === 2) {
-    command.log('OpenAPI 2.0 (Swagger) detected');
-    spectral.addFunctions(oas2Functions());
-    spectral.addRules(oas2Rules());
-  } else if (parseInt(spec.data.openapi) === 3) {
-    command.log('OpenAPI 3.x detected');
-    spectral.addFunctions(oas3Functions());
-    spectral.addRules(oas3Rules());
+  if (rules !== undefined) {
+    const numRules = new Array(rules).length;
```
P0lip: Does our collection have a `length` or `size` property or do we need to instantiate an array?

src/cli/commands/lint.ts
```diff
@@ -100,26 +109,38 @@ linting ./openapi.yaml
...
+    command.log(`Found ${numRules} rules`);
```
P0lip: I'd say it should be only printed when `verbose` is enabled.
Doesn't feel like it's very important info.

src/cli/commands/lint.ts
```diff
@@ -100,26 +109,38 @@ linting ./openapi.yaml
...
+      rules = oas2Rules();
+    } else if (parseInt(spec.data.openapi) === 3) {
+      command.log('OpenAPI 3.x detected');
+      spectral.addFunctions(oas3Functions());
+      rules = oas3Rules();
+    }
+  }
+
+  if (flags.skip) {
+    rules = skipRules(rules, flags, command);
```
P0lip: Suggest we make a shallow copy of rules (`{...rules}` should be fine) given the fact we mutate it in `skipRules` function.

README.md
philsturgeon: I was struggling with getting kebab case to work and angrily went with single words instead. I think seeing as this is in the context of lint, we do not have much room for confusion?

src/cli/commands/lint.ts
philsturgeon: I was getting errors when I used length.

src/cli/commands/lint.ts
philsturgeon: Done!

src/cli/commands/lint.ts
philsturgeon: When we have more usage I will put more effort into this, but for now a changelog will suffice.

src/cli/commands/lint.ts
mallachari: As I can see that Dictionary is just an object without any length or size properties. 
If so then wouldn't creating an Array from it always generate one element array regardless of how many properties the object has? Or do I see it wrong?

P0lip: > If so then wouldn't creating an array from it with new Array() always generate one element array regardless of how many properties the object has? Or do I see it wrong?

Yeah, it would.
```suggestion
    const numRules = Object.keys(rules).length;
```

README.md
P0lip: Yeah, I think it should be fine.

philsturgeon: I've got it cooking with kebabs.

P0lip: Great!

src/cli/commands/lint.ts
```diff
@@ -41,23 +45,28 @@ linting ./openapi.yaml
       char: 'o',
       description: 'output to a file instead of stdout',
     }),
+    'max-results': flagHelpers.integer({
```
P0lip: Should we add it back to readme?

philsturgeon: fixed!

philsturgeon:
Thanks for the awesome feedback and catching that new Array thing. I keep forgetting im not writing Ruby.

PR #203, opened 2019-05-14
feat: add support node 12

philsturgeon:

Recently pkg was updated to support node 12, so this PR makes sure our tests
pass in node 12 too. 

Node 12 support uncertain.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,27 @@
-version: 2
+version: 2.1
 
-jobs:
-  test_node_8:
-    docker:
-      - image: circleci/node:8
+commands:
+  cached-dependencies:
+    steps:
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install Dependencies
+          command: yarn install --frozen-lockfile
+      - save_cache:
+          name: Save Yarn Package Cache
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -22,25 +38,53 @@ jobs:
...
+    steps:
+      - install-and-test
+
   release:
     docker:
       - image: circleci/node:8
     steps:
       - checkout
-      - run: yarn
+      - cached-dependencies
```
philsturgeon: This makes our test suite moar faster.

.circleci/config.yml
```diff
@@ -22,25 +38,53 @@ jobs:
           command: |
             ./cc-test-reporter after-build --coverage-input-type lcov --exit-code $?
 
+jobs:
+  test-node-latest:
+    docker:
+      - image: circleci/node:latest
```
XVincentX: Are you sure you want to test on latest? It's kind of a tag pointing somewhere and when they release 13 I think it's ok that tests do not pass; this might be preventing the releases.

philsturgeon: This is just running tests to let us know if things work or not, we do not need to require this job to pass on GitHub. It will not block anything unless we configure it to do so.

XVincentX: Ah nice, I didn't know it was a possibility. Has it been already configured to be in such way? (not blocking the merges)

Issue #187, opened 2019-05-09
OpenAPI-specific Formats not supported


akrabat:
### **I'm submitting a...**
  - [x] bug report
  - [ ] feature request

### What is the current behavior?

Setting the `format` for an integer doesn't work.

test.yaml:
```
openapi: "3.0.2"
info:
  title: "Test"
  description: "Test spec"
  version: "1.0.0"
  contact:
    name: "Rob"
tags:
  - name: tag1
servers:
  - url: http://example.com

paths:
  /:
    get:
      operationId: root
      tags:
        - tag1
      summary: root
      description: The root node
      requestBody:
        content:
          application/x-www-form-urlencoded:
            schema:
              type: object
              properties:
                ip_address:
                  description: IP address (as a `long int`)
                  type: integer
                  format: int64
                  example: 2886989840
      responses:
        '200':
          description: Nothing returned
```

Run:
```
$ spectral lint test.yaml
Linting test.yaml
OpenAPI 3.x detected
Encountered error when running rule 'valid-example' on node at path '$,paths,/,get,requestBody,content,application/x-www-form-urlencoded,schema,properties,ip_address':
Error: unknown format "int64" is used in schema at path "#"
No errors or warnings found!
```
### What is the expected behavior?

Spectral should not display:

```
Error: unknown format "int64" is used in schema at path "#"
```

Also, it should not display _No errors or warnings found!_ when the line above starts with the word `Error` !

### What is the motivation / use case for changing the behavior?

[The Spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#data-types) says that `int64` is a valid `format` and the petstore-expanded example shows [it in use](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/petstore-expanded.yaml#L132-L135).

### Please tell us about your environment:

```
$ spectral --version
@stoplight/spectral/2.1.1 darwin-x64 node-v8.12.0
```

philsturgeon:
A few bits of debug here, from the chat with @akrabat in slack. 

The problem is with valid-example, it is validating the contents of your example and doesn't like it for some reason. We need to figure out why that's happening, but the 2nd issue you mention:

> Also, it should not display No errors or warnings found! when the line above starts with the word Error !

https://github.com/stoplightio/spectral/blob/master/src/functions/schema-path.ts#L37

[...]

Issue #116, opened 2019-03-28
Schema Rules have unhelpful summary in default CLI output


philsturgeon:
Linting this document:

```yaml
openapi: 3.0.8
info:
  title: occaecat ex exercitation consequat
  version: sunt Duis aliqua qui
x-testcase: 'should not work without required property: paths'
```

Results in stylish format output:

```
linting ../openapi3-examples/fail/fuzz1/331be1bf-781d-407f-93d6-1f4b390ae32b.yaml
OpenAPI 3.x detected

/home/mike/node/openapi3-examples/fail/fuzz1/331be1bf-781d-407f-93d6-1f4b390ae32b.yaml
 1:1    error  oas3-schema       Validate structure of OpenAPIv3 specification
 1:1  warning  api-servers       OpenAPI `servers` must be present and non-empty array
 2:6  warning  info-contact      Info object should contain `contact` object
 2:6  warning  info-description  OpenAPI object info `description` must be present and non-empty string

✖ 4 problems (1 error, 3 warnings, 0 infos)
```

But the json format output gives the more helpful message for the error:

```json
	{
		"code": "oas3-schema",
		"path": [],
		"message": "should have required property 'paths'",
		"severity": 0
	},
```

_Originally posted by @MikeRalphson in https://github.com/stoplightio/spectral/pull/112#issuecomment-477554960_

philsturgeon:
Yeah this is a tough one. Most of the time the summary is a bit more concise but for oasX-schema based rules you really want the message. Id rather not make the formatters be aware of such things, so we need to find somewhere slightly higher up to make that change perhaps.

MikeRalphson:
> Id rather not make the formatters be aware of such things

Understood, just to clarify, my concern was that a program consuming the json format output cannot obtain the same error message as is displayed by the stylish formatter currently.

philsturgeon:
Oh thats the inverse of what I said! Ha. We can sort that at the same time.

philsturgeon:
@MikeRalphson great so @P0lip just added summary to JSON in #117 so we'll sort the bit I said out later.

P0lip:
Fixed in 2.4.0

PR #215, opened 2019-05-17
feat: improve our AJV usage

P0lip:

Feature (fewer messages?) + fixes for https://github.com/stoplightio/spectral/issues/187 https://github.com/stoplightio/spectral/issues/116

image

image

Not really :)

### Other information

My initial intention was to make CLI messages more flexible, but this turned out to be not the greatest idea as we actually need meaningful messages only for stylish CLI format. API usage and JSON CLI format should be untouched, so decided to add the simple condition, so that both worlds are satisfied.

### DIFF
--- a/src/__tests__/linter.test.ts
+++ b/src/__tests__/linter.test.ts
@@ -97,7 +97,7 @@ describe('linter', () => {
       },
     });
 
-    expect(result[0]).toEqual({
+    expect(result[0]).toMatchObject({
       code: 'rule1',
       message,
       severity: DiagnosticSeverity.Warning,
@@ -206,19 +206,16 @@ responses:: !!foo
     expect(result).toEqual([
       expect.objectContaining({
         code: 'oas3-schema',
-        summary: 'Validate structure of OpenAPIv3 specification.',
         message: 'should NOT have additional properties: type',
         path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
       }),
       expect.objectContaining({
         code: 'oas3-schema',
-        summary: 'Validate structure of OpenAPIv3 specification.',
         message: 'should match exactly one schema in oneOf',
         path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
       }),
       expect.objectContaining({
         code: 'oas3-schema',
-        summary: 'Validate structure of OpenAPIv3 specification.',
         message: "should have required property '$ref'",
         path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
       }),
--- a/src/__tests__/linter.test.ts
+++ b/src/__tests__/linter.test.ts
@@ -207,16 +207,19 @@ responses:: !!foo
       expect.objectContaining({
         code: 'oas3-schema',
         message: 'should NOT have additional properties: type',
+        summary: 'should NOT have additional properties: type',
         path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
       }),
       expect.objectContaining({
         code: 'oas3-schema',
         message: 'should match exactly one schema in oneOf',
+        summary: 'should match exactly one schema in oneOf',
         path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
       }),
       expect.objectContaining({
         code: 'oas3-schema',
         message: "should have required property '$ref'",
+        summary: "should have required property '$ref'",
         path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
       }),
     ]);
--- a/src/__tests__/linter.test.ts
+++ b/src/__tests__/linter.test.ts
@@ -1,7 +1,15 @@
 import { DiagnosticSeverity } from '@stoplight/types';
+import * as fs from 'fs';
+import * as path from 'path';
 
 import { oas2Functions, oas2Rules } from '../rulesets/oas2';
+import { oas3Functions, oas3Rules } from '../rulesets/oas3';
 import { Spectral } from '../spectral';
+
+const invalidSchema = fs.readFileSync(
+  path.join(__dirname, './__fixtures__/petstore.invalid-schema.oas3.yaml'),
+  'utf-8',
+);
 
 const fnName = 'fake';
 const fnName2 = 'fake2';
@@ -187,6 +195,34 @@ responses:: !!foo
         },
       ]),
     );
+  });
+
+  test('should merge similar ajv errors', async () => {
+    spectral.addRules(oas3Rules());
+    spectral.addFunctions(oas3Functions());
+
+    const result = await spectral.run(invalidSchema);
+
+    expect(result).toEqual([
+      expect.objectContaining({
+        code: 'oas3-schema',
+        summary: 'Validate structure of OpenAPIv3 specification.',
+        message: 'should NOT have additional properties: type',
+        path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
+      }),
+      expect.objectContaining({
+        code: 'oas3-schema',
+        summary: 'Validate structure of OpenAPIv3 specification.',
+        message: 'should match exactly one schema in oneOf',
+        path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
+      }),
+      expect.objectContaining({
+        code: 'oas3-schema',
+        summary: 'Validate structure of OpenAPIv3 specification.',
+        message: "should have required property '$ref'",
+        path: ['paths', '/pets', 'get', 'responses', '200', 'headers', 'header-1'],
+      }),
+    ]);
   });
 
   describe('functional tests for the given property', () => {
--- a/src/functions/schema.ts
+++ b/src/functions/schema.ts
@@ -2,6 +2,7 @@ import { decodePointerFragment } from '@stoplight/json';
 import * as AJV from 'ajv';
 import * as jsonSpecv4 from 'ajv/lib/refs/json-schema-draft-04.json';
 const validator = require('ajv-oai/lib/format-validator');
+import { IFunction, IFunctionResult, ISchemaOptions } from '../types';
 
 const ajv = new AJV({
   meta: false,
@@ -21,9 +22,6 @@ ajv.addFormat('float', { type: 'number', validate: validator.float });
 ajv.addFormat('double', { type: 'number', validate: validator.double });
 ajv.addFormat('byte', { type: 'string', validate: validator.byte });
 
-import { AdditionalPropertiesParams } from 'ajv';
-import { IFunction, IFunctionResult, ISchemaOptions } from '../types';
-
 const formatPath = (path: string) =>
   path
     .split('/')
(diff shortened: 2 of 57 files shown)


src/functions/schema.ts
```diff
@@ -14,6 +15,12 @@ ajv._opts.defaultMeta = jsonSpecv4.id;
 // @ts-ignore
 ajv._refs['http://json-schema.org/schema'] = 'http://json-schema.org/draft-04/schema';
 
+ajv.addFormat('int32', { type: 'number', validate: validator.int32 });
+ajv.addFormat('int64', { type: 'number', validate: validator.int64 });
+ajv.addFormat('float', { type: 'number', validate: validator.float });
+ajv.addFormat('double', { type: 'number', validate: validator.double });
+ajv.addFormat('byte', { type: 'string', validate: validator.byte });
```
XVincentX: I had a discussion with @philsturgeon about this just yesterday and he was against this. Did you folks change your mind?

marbemac:
Another thought, might be worth evaluating https://github.com/atlassian/better-ajv-errors.

src/functions/schema.ts
```diff
@@ -1,6 +1,8 @@
 import { decodePointerFragment } from '@stoplight/json';
 import * as AJV from 'ajv';
 import * as jsonSpecv4 from 'ajv/lib/refs/json-schema-draft-04.json';
+const validator = require('ajv-oai/lib/format-validator');
```
philsturgeon:
```suggestion
const oasFormatValidator = require('ajv-oai/lib/format-validator');
```

src/functions/schema.ts
philsturgeon: Ok, this approach is a bit nicer than what I had in mind when we discussed using ajv-oas. I thought we'd be passing all validation through that function, instead we're kinda teaching our generic validation logic about some of the formats OpenAPI people are likely to use, and seeing as OpenAPI and JSON Schema are used interchangeably by many its not hurting to have these available in generic JSON Schema usage anyway. 

Let's go with this, and if we get pushback about having OpenAPI-specific formats enabled for all the things then... ok we can look into adding specific validation rules for specific rulesets, or autodetecting OpenAPI, or some other thing.

P0lip:
> I would like to see our summary generally be more useful, it should be a very quick hint as to what is going on, with a little information about the field or whatever. Those in the know will look at this short summary and know how to fix it, but others will want more information.

I'd drop summary entirely. Not really sure what's the point of that property.
Ideally, we should have something similar to what ESLint does. That is:
```

I'm not sure whether it makes sense to include property given the fact path is already present.
There is path available

**Comment by P0lip** [comment]
> I would like to see our summary generally be more useful, it should be a very quick hint as to what is going on, with a little information about the field or whatever. Those in the know will look at this short summary and know how to fix it, but others will want more information.

I'd drop summary entirely. Not really sure what's the point of that property.
Ideally, we should have something similar to the snippet below (inspired by ESLint)
```
{
    rule: "oas2-schema",
    message: "Expected something to be truthy",
    path: ['some','path']
    severity: 2,
    line: 100,
    column: 2
}
```
I'm not sure whether it makes sense to include property given the fact path is already present.
It'll also make things more difficult in Studio as we already do highlight the line and point to given line and column etc, so we would most likely want to avoid any long messages.
I'm open to discussion though. I do agree that the current solution is not ideal.

> This more information might have strings we've written or other variables involved, with maybe information on how to fix it, etc.

Agreed, would be nice to give some feedback. I'd say the best way to do it is to provide an extensive doc about what given rule does and give examples when it is met and when not.
Yet again, ESLint is a good example here
https://eslint.org/docs/rules/
https://eslint.org/docs/rules/eqeqeq
https://eslint.org/docs/rules/no-unexpected-multiline

File: src/functions/schema.ts (hunk 1 above)
Comment (XVincentX) [reply src/functions/schema.ts]: Hmm, ok. I fail to understand the difference with what was proposed last week though. It was the exact same solution.

**Comment by philsturgeon** [comment]
Literally putting everything through the os validator, vrs adding a few extra rules from that validator, is conceptually different enough that it doesn’t feel quite as unsettling.  

--
Phil Sturgeon
@philsturgeon

**Comment by marbemac** [comment]
> I'd drop summary entirely.

I agree, summary doesn't make much sense on the validation results (still makes sense on the rule definition). Instead, consumers could retrieve summary on demand, as needed. We already expose `get rules()` on the spectral instance, so with no changes a consumer could do this:

```ts
const summary = spectralInstance.rules[result.rule].summary;
```

> I'm not sure whether it makes sense to include property given the fact path is already present.

Yeah I agree in general, but there is a tricky use case - how to describe a missing property. Because the property does not exist in the source, we have nothing to highlight / point to inside of Studio, for example. So we end up putting the error on the parent object, and having a bit more information about the missing property in the message is helpful, e.g. "the description property is missing" versus "a property is missing" (and having user look at the path to see which property is missing).

> I would like to see our summary generally be more useful, it should be a very quick hint as to what is going on, with a little information about the field or whatever. Those in the know will look at this short summary and know how to fix it, but others will want more information.

Maybe we should more formally define what the goal of `message` (property on each validation result) and `summary` (property on the rule itself) is? As it stands right now, message is generated at runtime, during validation, and summary is hardcoded into the rule itself and never changes.

> That said, I'm still open to making messages more customizable.
We could implement a decent string interpolation.

We could run messages through https://mustache.github.io, and expose several variables to it, so that rule authors can generate more useful messages?

> Literally putting everything through the os validator

I agree - we should improve the messages in spectral itself (which will benefit all downstream consumers, such as Studio), versus cleaning things up in the CLI (assuming I'm understanding this PR correctly).

P0lip:
> Maybe we should more formally define what the goal of message (property on each validation result) and summary (property on the rule itself) is? As it stands right now, message is generated at runtime, during validation, and summary is hardcoded into the rule itself and never changes.

`summary` could be renamed to `description`. This would be less confusing.

> Yeah I agree in general, but there is a tricky use case - how to describe a missing property. Because the property does not exist in the source, we have nothing to highlight / point to inside of Studio, for example. So we end up putting the error on the parent object, and having a bit more information about the missing property in the message is helpful, e.g. "the description property is missing" versus "a property is missing" (and having user look at the path to see which property is missing).

This is indeed a valid case.

> We could run messages through https://mustache.github.io, and expose several variables to it, so that rule authors can generate more useful messages?

Frankly speaking, this what I wanted to initially implement in the scope of this PR.
I even implemented some code that should be sufficient to cover our needs.
```

**Comment by P0lip** [comment]
@chris-miaskowski you might want to take a look at this PR as well

**Comment by philsturgeon** [comment]
This PR is in response to https://stoplightio.atlassian.net/browse/SO-84 which is meant to formally define what summary and message are, because sometimes summary is useful and sometimes message is useful. Having the two has lead to inconsistencies.

I feel like we got to a good consensus here, lets drop summary entirely and just have message, which looks like this: 

image

Basically exactly what you've done, but everywhere, for all the formatters. 

To avoid needing to make this a breaking change and therefore v3.0, we can just make summary an alias of message and mark it as deprecated in code, or something like that?

**Comment by P0lip** [comment]
@philsturgeon 
Sounds like a good plan to me! Will do.

File: src/rulesets/oas3/__tests__/valid-example.ts [hunk 3]
```diff
@@ -3,6 +3,8 @@ import { oas3Rules } from '../index';
 
 const ruleset = { rules: oas3Rules() };
 
+declare var test: jest.It;
```
XVincentX: Why is this required?

src/__tests__/linter.test.ts
```diff
@@ -206,19 +206,16 @@ responses:: !!foo
     expect(result).toEqual([
       expect.objectContaining({
         code: 'oas3-schema',
-        summary: 'Validate structure of OpenAPIv3 specification.',
```
XVincentX: If `summary` is the same as `message`, wouldn't it be worth to still check its presence?

src/linter.ts
```diff
@@ -88,13 +88,25 @@ export const lintNode = (
...
-        const location = parsedResult.getLocationForJsonPath(parsed, result.path || targetPath, true);
+        const path = result.path || targetPath;
+        const location = parsedResult.getLocationForJsonPath(parsed, path, true);
 
         return {
           code: rule.name,
-          summary: rule.summary,
-          message: result.message,
-          path: result.path || targetPath,
+          get summary() {
```
XVincentX: Nitpick: new line

src/linter.ts
```diff
@@ -88,13 +88,25 @@ export const lintNode = (
...
+            return this.message;
+          },
+          message:
```
XVincentX: Nitpick: new line

src/linter.ts
XVincentX: I guess this is a property because we can't use `this` if not in a defined function, right?

src/linter.ts
XVincentX: Ok took me some time to get this — maybe a comment could help here

src/rulesets/oas3/__tests__/valid-example.ts
P0lip: We have `mocha` test runner included in Spectral and TS is confused and doesn't know which `it` function it should refer to.

philsturgeon: Can we get rid of mocha? I think oclif added jest support a while back. https://github.com/oclif/oclif/pull/219

P0lip: Sorry, was wrong. Mocha is not listed as a dependency, but TS still sees typings for it and prefers it over Jest.

P0lip:
```
# npm ls @types/mocha # gives the following
@stoplight/spectral@0.0.0 /home/Stoplight/spectral
└─┬ @oclif/test@1.2.4
  └─┬ fancy-test@1.4.3
    └── @types/mocha@5.2.6 
```
Seems like oclif/test requires types/mocha

src/__tests__/linter.test.ts
P0lip: Sounds reasonable.

src/rulesets/oas3/__tests__/valid-example.ts
P0lip: Well, they haven't released any newer version, so we'll need to stick to that unless you have a better workaround.
The reason I had to override the type is caused by the use of `test.each`. Plain `it` or `test` have a very similar, if not exact, signature.
Should I add a comment explaining that?
@XVincentX @philsturgeon

XVincentX: Ok figured it out in person. We can't do anything about this unless we decide to ditch the whole oclif testing framework.

src/rulesets/oas3/__tests__/valid-example.ts
```diff
@@ -3,6 +3,8 @@ import { oas3Rules } from '../index';
 
 const ruleset = { rules: oas3Rules() };
 
+// @oclif/test packages requires @types/mocha, therefore we have 2 packages coming up with similar typings
+// TS is confused and prefers the mocha ones, so we need to instrument it to pick up the Jest ones
```
XVincentX: Nitpick: newline.

P0lip:
@philsturgeon do you have any extra feedback or can I merge it?

@marbemac a few words on the package you mentioned - it seems truly great. I'm not sure what others think, but IMHO it might be a nice addition. Just wondering whether we should do it as a part of this PR or next one. Up to discussion.

marbemac:
Totally cool to do a second pass and separate PR to make the messages even better in the future! Don't need to block this one.

src/rulesets/oas/functions/oasPathParam/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -33,11 +31,7 @@ exports[`oasPathParam Error if duplicate path parameters with same name are used
 Array [
   Object {
     "code": "path-params",
-    "message": "The path \\"**/foo/{bar}/{bar}**\\" uses the parameter \\"**{bar}**\\" multiple times.
-
-Path parameters must be unique.
-
-To fix, update the path so that all parameter names are unique.",
```
philsturgeon: The old message said they should be unique and the new one doesn’t. Can we bring that back?

P0lip: Sure Phil, thanks for spotting this, I'll create a PR to address that soon (certainly today, this is 21.05).

PR #233, opened 2019-05-29
[WIP] fix: "data/required should be array" issue

lag-of-death:

### Other information

### DIFF
--- a/src/functions/example.ts
+++ b/src/functions/example.ts
@@ -1,0 +1,13 @@
+export function example(a: any, b: any, c: any, d: any) {
+  const original = d.original;
+
+  const type = (original.schema && original.schema.type) || original.type;
+
+  return type && typeof original.example !== type
+    ? [
+        {
+          message: `${original.example} is not of type ${type}`,
+        },
+      ]
+    : [];
+}
--- a/src/rulesets/oas3/ruleset.json
+++ b/src/rulesets/oas3/ruleset.json
@@ -2388,13 +2388,26 @@
         }
       }
     },
-    "valid-example": {
-      "summary": "Examples must be valid against their defined schema.",
+    "valid-schema-example": {
+      "summary": "Schema examples must be valid against their defined schema.",
       "message": "\"{{property}}\" property {{error}}",
       "type": "validation",
-      "given": "$..[?(@.example)]",
+      "given": "$..*[?(@.schema)][?(@.example)]",
       "then": {
         "function": "schemaPath",
+        "functionOptions": {
+          "field": "example",
+          "schemaPath": "$"
+        }
+      }
+    },
+    "valid-openapi-example": {
+      "summary": "Examples must be valid against ???",
+      "message": "{{error}}",
+      "type": "validation",
+      "given": "$..*[?(!@.schema)][?(@.example)]",
+      "then": {
+        "function": "example",
         "functionOptions": {
           "field": "example",
           "schemaPath": "$"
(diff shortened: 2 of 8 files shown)


src/rulesets/oas3/ruleset.json
```diff
@@ -2388,18 +2388,31 @@
...
       }
     },
-    "valid-example": {
-      "summary": "Examples must be valid against their defined schema.",
+    "valid-schema-example": {
+      "summary": "Schema examples must be valid against their defined schema.",
       "message": "\"{{property}}\" property {{error}}",
       "type": "validation",
-      "given": "$..[?(@.example)]",
+      "given": "$..*[?(@.schema)][?(@.example)]",
```
lag-of-death: So now, this rule only looks for `example`s inside `schema`s.

src/rulesets/oas3/ruleset.json
```diff
@@ -2388,18 +2388,31 @@
...
           "field": "example",
           "schemaPath": "$"
         }
       }
+    },
+    "valid-openapi-example": {
+      "summary": "Examples must be valid against ???",
+      "message": "{{error}}",
+      "type": "validation",
+      "given": "$..*[?(!@.schema)][?(@.example)]",
```
lag-of-death: This rule only looks for `example`s NOT inside `schema`s.

src/rulesets/oas3/ruleset.json
```diff
@@ -2388,18 +2388,31 @@
...
+      "then": {
+        "function": "example",
```
lag-of-death: I've created a separate function for this validation - `example`. But this is only for prototyping purposes.

src/functions/example.ts
```diff
@@ -0,0 +1,13 @@
+export function example(a: any, b: any, c: any, d: any) {
```
lag-of-death: So this might look weird. When there is an `x.example` field, that is not inside `x.schema` field (so not `x.schema.example`), then:
- define `type` as one of `x.schema.type` or `x.type`
- verify if `x.example` is of `type`
This is only an example and I'm not sure if we even want to validate `x.example` field. Do we? If we do, is this the way to go? Or should it be validated somehow else?

philsturgeon: This is very confusing, good job for battling through. 

We want to support:

``` yaml
      parameters:
        - in: path
          name: petId
          required: true
          schema:
            type: integer
          example: 3
```

``` yaml
      parameters:
        - in: path
          name: petId
          required: true
          schema:
            type: integer
            example: 3
```

It's a quirk in OpenAPI that these are both allowed. I don't know why they are. But they are. :(

We should use exactly the same logic to validate either, but the function parameters are a little different. 

Check out the valid-example rule:

```
        "function": "schemaPath",
        "functionOptions": {
          "field": "example",
          "schemaPath": "$"
        }
```

The schemaPath is saying "pass in this current object", which would look like this:

            type: integer
            example: 3

The field keyword is asking for the field we want to validate, which is example, so now we are validating `3` against the schema `type: integer`. Huge success, everything works, yay. Until...

```
            schema: { whatever }
            required: true
            example: 3
```

The only problem is that example is being spotted (outside of a schema) so now the `schemaPath: $` has a bunch of keywords which arent a schema, and that freaks AJV out.

So the schemaPath function is fine, and we wont need an example function because that is too specific. You did a good job making valid-schema-example more accurate, but that is the only change required for that rule. The valid-openapi-example should include:

```
        "functionOptions": {
          "field": "example",
          "schemaPath": "$.schema"
        }

or something. Basically if an example is sitting next to a schema, we want to pass that example to that schema. 

Confused? Good, so am I. 😭

lag-of-death: Hehe, yes, a bit confused ;) thank you @philsturgeon for the detailed description! I shall dig into it some more! ;)

Issue #223, opened 2019-05-22
Parameter with required and example fails valid-example rule


cbautista1002:
### **I'm submitting a...**
  - [x] bug report

### What is the current behavior?

If you have a parameter that has `required: true` and an example, the `valid-example` rule fails. See "Other information" for details.

### What is the expected behavior?

The `valid-example` rule executes successfully for a valid parameter example as it understands that required is not a list for parameters but a boolean.

### What is the motivation / use case for changing the behavior?

The `valid-example` rule is very helpful and it should work for parameters

### Please tell us about your environment:
```
$ spectral --version
@stoplight/spectral/3.0.0 linux-x64 node-v12.2.0
```

### Other information

Given this schema
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: Swagger Petstore
  license:
    name: MIT
  contact:
    name: Test
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets/{petId}:
    get:
      summary: Info for a specific pet
      operationId: showPetById
      description: Get a pet
      tags:
        - pets
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
          example: "1234567890"
      responses:
        '200':
          description: Expected response to a valid request
```

Running `spectral lint myschema.yml` results in:
```
Adding OpenAPI 3.x functions
OpenAPI 3.x detected
Encountered error when running rule 'valid-example' on node at path '$,paths,/pets/{petId},get,parameters,0':
Error: schema is invalid: data/required should be array
No errors or warnings found!
```

As an aside, an "error" was found but yet spectral says "No errors or warnings found!" and the return code is `0`. Why the discrepancy?

Thanks a lot. Let me know if I can help!

philsturgeon:
Hey, there is an error in your example, the example is invalid. If you fix the example this will go away. :)

Example errors are not related the same as other errors, so the word “error” is overloaded here and this looks funny for sure. We want to improve that so users know what’s up, but you know how to fix the thing in your spec for now. 

--
Phil Sturgeon
@philsturgeon

cbautista1002:
Hi @philsturgeon, thanks for responding, but I don't see what's wrong with my example. Do you mind pointing out what's wrong?

I tried another example (changed string to integer)
```yaml
      parameters:
        - in: path
          name: petId
          required: true
          description: The id of the pet to retrieve
          schema:
            type: integer
[...]

PR #243, opened 2019-06-03
Parameter with required and example fails valid-example rule

lag-of-death:

Bug fix.

https://github.com/stoplightio/spectral/issues/223

No.


src/functions/schema-path.ts
```diff
@@ -37,5 +37,11 @@ export const schemaPath: IFunction<ISchemaPathOptions> = (targetVal, opts, paths
     console.error(error);
   }
 
-  return schema(relevantObject, { schema: schemaObject }, paths, otherValues);
+  return schema(relevantObject, { schema: withoutRequiredIfNotArray(schemaObject) }, paths, otherValues);
 };
+
+function withoutRequiredIfNotArray(schemaObject: any) {
```
philsturgeon: Is this required or is this an older change that is left in here? 

This should not need to be done. If `required: true` is there it means the rule is operating on a parameter object, which with the jsonpath rules you've set up, it shouldnt ever be. 

The only way a schema object will have a required: true in it, is because the user has invalid API specifications, and that'll get caught earlier.

lag-of-death: hi @phil :)
so `oas3-schema` rule lints a spec with:
```yaml
     parameters:
        - name: petId
          in: path
          required: true
          schema:
            type: string
            example: "2323"
          example: "2323"
```
as OK. So, I'm assuming this is valid OAS3. Maybe it isn't?
Now:
1. `valid-openapi-example` will match `schema` and `schema.example`
2. `valid-example` will match `parameters[0]` and `parameters[0].example`

The problem is with `2`, since there is `required: true` on the object, that is later passed to https://github.com/stoplightio/spectral/blob/6590414fc69777e58c084463c83c11092aaccab6/src/functions/schema.ts#L61, and that's why I'm using `withoutRequiredIfNotArray`. Maybe `oas3-schema` should lint this case as OK ?

philsturgeon: this is entirely valid OAS3 yeah.

1. valid-schema-example should match example against the `schema` object
1. valid-openapi-example should match example against the current (`$.`) object

lag-of-death: @phil, done, `withoutRequiredIfNotArray` is removed now

philsturgeon:
This is pretty close but OpenAPI v2 and v3 are tricky beasts.  Let's do some TDD and I'll write up the tests for you to code into.

src/rulesets/oas3/__tests__/valid-example.ts
```diff
@@ -215,3 +216,95 @@ describe('valid-example', () => {
     },
   );
 });
+
+describe('valid-schema-example', () => {
```
philsturgeon: Looks like we've split this one rule into two but have left them both in the old file? The file names match the rules so we should have:

- src/rulesets/oas3/__tests__/valid-openapi-example.ts
- src/rulesets/oas3/__tests__/valid-schema-example.ts

src/rulesets/oas3/valid-example-ruleset.json
```diff
@@ -0,0 +1,17 @@
+{
```
philsturgeon: I've never seen a rule specific ruleset before. What is in the intention? Is this for testing?

lag-of-death: It's to move out this rule from `oas3 rules`. My understanding is that `spectral` can validate jsons that are not `oas2`/`oas3`. We could also move this rule to some other `not oas2/oas3` ruleset, but haven't found one (is there one existing?)

philsturgeon: functions are generic and can be used anywhere. We have `schemaPath` as a generic function, and this is the basis of the valid example rules.

The rules are different for oas2 and oas3 because the specification structures are different.

So we reuse the schemaPath function, but do not reuse the valid example stuff because we cannot.

lag-of-death: And that's why I have moved `valid-example` out of `ruleset.json` for oas3

src/rulesets/oas3/__tests__/valid-example.ts
lag-of-death: If not in `src/rulesets/oas3/ruleset.json`, then where I should put these 2 new rules? Thought `oas3/ruleset.json` is the place as the two rules belong to oas3 validators, no?

philsturgeon: The valid-example.ts is when there was one rule. You are splitting it into two rules. So make two files, the ones above, and tadaaaa you’re done. No new ruleset, that’s a total misuse of ruleset a.

1 rule becomes 2 rules.

src/rulesets/oas3/valid-example-ruleset.json
philsturgeon: Please remove this new file.

lag-of-death: @philsturgeon, if I remove, where I put `valid-example`? I'm kinda confused. I still have to test `valid-example` rule..

lag-of-death: @philsturgeon, I can move already existing tests from `src/rulesets/oas3/__tests__/valid-example.ts` to  `src/rulesets/oas2/__tests__/valid-example.ts`, since `valid-example` rule didn't change for `oas2`. So this way `src/rulesets/oas3/__tests__/valid-example.ts` and `src/rulesets/oas3/valid-example-ruleset.json` are removed. Would that be OK?

philsturgeon: So after seeing the talk about how many different types of param/header/schema example there is, maybe you could use that to create some tests which highlight each type of parameter and property example we looked at, then make sure this change works with each of those? 

The ruleset you added is not the way to go about DRYing things, although I do understand the urge. Mixing up OAS2 and OAS3 is also not the way to go, as the two specs are different. 

Seeing as you have split the `valid-example` rule in two, and the describe tests split in two, I was expecting you to just split the `valid-example` file in two. 

instead of having `describe('valid-schema-example', () => {` inside `src/rulesets/oas3/__tests__/valid-example.ts ` the convention is to match the test file with whatever you are describing, so if you are gonna have two different rules have two different test files.

Is that all starting to make sense?

lag-of-death:
Hi @phil. This is WIP, but I wanted to show you what's been done so far. Please mind that external examples (`.externalValue`) were not considered.

src/functions/example-type.ts
```diff
@@ -0,0 +1,37 @@
+import * as AJV from 'ajv';
```
philsturgeon: I don’t understand what this file is, could you give it a quick explanation?

lag-of-death: Sure. @philsturgeon, this file is used by `valid-type-example` rule: https://github.com/stoplightio/spectral/blob/06fcb9e4cadc78b5c4299abd28d8f327f913f0f1/src/rulesets/oas3/ruleset.json#L2432
This rule takes care of the cases where we have:
```
type: some_type
example: some example
```
So where the shape of an example is defined by `.type`, not by `schema`.

philsturgeon:
Work is now being done in #472

PR #245, opened 2019-06-04
feat: add default resolvers 

### LINKED ISSUE #134: Add optional support for external $ref in CLI
### **I'm submitting a...**
  - [ ] bug report
  - [x] feature request

The default behavior of Spectral v2.0 is to operate only on the original description object for 95% of the rules, and the uses the `resolvedTarget` in a few rules which specifically request that resolved version of the object (like oasOpsParams). 

Spectral CLI uses [json-ref-parser](https://github.com/stoplightio/json-ref-resolver) to created that resolvedTarget, and by default json-ref-parser only processes internal `$ref`. This is a safe default, because $ref can be potentially used maliciously if you are working with description files you don't control.  

This is unexpected for people transitioning from Speccy, which did external by default. See #108.

Maybe we could have a new CLI option for the `lint` command, maybe `--external`.

The main `Spectral` constructor takes an object with a `resolver` key, which defaults to a stock standard json-ref-parser instance. The lint command should take advantage of this and change the resolver based on CLI options. 

```
if (flags.external) {
    const spectral = new Spectral(resolver: <fancy-resolver-however-that-works>);
}
```

Or something like that.

Thoughts @tbarn @marbemac @MikeRalphson?

marbemac:
fixes #134 #171 

WIP - still needs tests and docs.

Bug fix / kind of a feature?

Will add description to readme.

No.

### Other information


src/resolvers/http.ts
```diff
@@ -0,0 +1,16 @@
+import { Resolver } from '@stoplight/json-ref-resolver';
+import unfetch from 'unfetch';
+
+export const httpReader = {
+  async read(ref: any) {
+    return unfetch(String(ref)).then(r => r.text);
```
P0lip: code-style: you could use `await` here.

src/spectral.ts
```diff
@@ -23,13 +25,13 @@ export * from './types';
...
 
-  private resolver: Resolver;
-  constructor(opts?: IConstructorOpts) {
-    this.resolver = opts && opts.resolver ? opts.resolver : new Resolver();
+  constructor(opts: IConstructorOpts = {}) {
+    this._resolver = opts.resolver ? opts.resolver : new Resolver();
   }
 
-  public async run(target: IParsedResult | object | string): Promise<IRuleResult[]> {
+  public async run(target: IParsedResult | object | string, opts: IRunOpts = {}): Promise<IRuleResult[]> {
```
P0lip: This looks a bit weird. I believe we should let resolver be configurable rather than providing its settings to `run` method.
Thoughts?

src/cli/commands/lint.ts
```diff
@@ -124,8 +126,11 @@ async function lint(name: string, flags: any, command: Lint, rules?: RuleCollect
   if (flags.verbose) {
     command.log(`Linting ${name}`);
   }
-  const spec: IParserResult = await readParsable(name, flags.encoding);
-  const spectral = new Spectral();
+
+  const targetUri = resolve(name);
+
+  const spec: IParserResult = await readParsable(targetUri, flags.encoding);
+  const spectral = new Spectral({ resolver: httpAndFileResolver });
```
P0lip: One more thought.
I'd say the above might be quite a significant change. It may slow down CI since Spectral will start making HTTP requests. We should provide a way to opt-out - a CLI flag here for resolver would be more than welcome.
It's a different story, though, so it can stay as is for now.

src/spectral.ts
marbemac: The `authority` is not a global resolver option, since it kind of depends on the file being run, so that option won't really work.

src/cli/commands/lint.ts
marbemac: It will only make http requests if there are http $refs in the spec, and if there are http $refs in the spec, we can't really validate the spec without resolving them.

src/resolvers/http.ts
marbemac: Yeah or we can just remove the async from line 5 :).

src/cli/commands/lint.ts
P0lip: I'm aware, but IMHO we should provide a way to opt-out.

src/resolvers/http.ts
P0lip: Yeah, you are also missing a function call 
```suggestion
    return unfetch(String(ref)).then(r => r.text());
```
Will fix it.

src/cli/commands/lint.ts
philsturgeon: What would that actually mean though, would it just fail for folks using the CLI if they have HTTP refs? I think then opt out is removing the HTTP refs from your spec right?

src/cli/commands/lint.ts
```diff
@@ -163,12 +169,17 @@ async function lint(name: string, flags: any, command: Lint, rules?: RuleCollect
...
-      source: resolve(process.cwd(), name),
+      source: targetUri,
       parsed: spec,
       getLocationForJsonPath,
     };
 
-    results = await spectral.run(parsedResult);
+    results = await spectral.run(parsedResult, {
+      resolve: {
+        authority: targetUri,
```
philsturgeon: Why is it called authority? It's not quite clicking for me yet.

src/cli/commands/lint.ts
P0lip: > What would that actually mean though, would it just fail for folks using the CLI if they have HTTP refs? I think then opt out is removing the HTTP refs from your spec right?

Opting-out would mean turning off http/https $ref resolution.
Resolving external resources may significantly slow down CI. Not a big deal though, we can leave it as is.

marbemac: Yeah opting out would mean it would fail for folks if they use http $refs. I can't see a use case for this - how about we re-consider if somebody brings it up later?

src/cli/commands/lint.ts
marbemac: Authority is just piggybacking off of the `authority` option in https://github.com/stoplightio/json-ref-resolver. I created json-ref-resolver quite a while ago, and in hindsight it's not the best name for that property. I have an open ticket to make an update to json-ref-resolver to use https://github.com/microsoft/vscode-uri rather than urijs (to fix a bunch of issues with windows style file paths), and was thinking about leveraging that breaking release to also update some of the names.

We could decide on a name right now, and use that in spectral so that we don't have a breaking change here later? The authority option of json-ref-resolver is described here: https://github.com/stoplightio/json-ref-resolver#example-resolving-relative-remote-references-with-the-authority-option.

Basically, it's the location of the document currently being resolved, so that relative $refs inside of it can be effectively resolved (solution to #171).

Open to suggestions, here are some:

- parentUri
- documentUri
- uri
- $id (from json schema, "It declares a base URI against which $ref URIs are resolved."). This is basically exactly how we're using it.
- documentId

package.json
```diff
@@ -71,7 +72,8 @@
     "node-fetch": "2.x.x",
     "strip-ansi": "^5.2.0",
     "text-table": "^0.2.0",
-    "typescript-json-schema": "^0.38.0"
+    "typescript-json-schema": "^0.38.0",
+    "unfetch": "^4.1.0"
```
philsturgeon: We've been using node-fetch which uses node Request and other core stuff internally, meaning we can use tools like Polly.js for testing. Would you mind switching to that?

src/cli/commands/lint.ts
philsturgeon: Yeah we've had a lot of complaints that it cant resolve and I cant imagine it going the other way. I agree we should reconsider if somebody asks for a way to turn it off. Maybe for security reasons or something else.

src/cli/commands/lint.ts
philsturgeon: OHHH i see! documentUri sounds great. Maybe we could switch to vscode-uri and do all the breaky things in one go?

marbemac: > I have an open ticket to make an update to json-ref-resolver to use https://github.com/microsoft/vscode-uri rather than urijs (to fix a bunch of issues with windows style file paths), and was thinking about leveraging that breaking release to also update some of the names.

:)

package.json
marbemac: Sure, good idea!

marbemac:
Alright we have a tricky question...

Should we report errors in $ref'd files? For example, in the below image the `library.yaml` file has an error (the example does not match the schema). `one.yaml` references the address model in `library.yaml`. If a user runs spectral on `library.yaml`, should the error in `one.yaml` be reported? If so, it brings up the tricky issue of how to report this error to the end user in a way that is not confusing (e.g. it should probably say its in a different file, with accurate line numbers relevant to that file, etc). This is non-trivial to do.

Screen Shot 2019-06-04 at 8 08 36 PM

P0lip:
@marbemac I think I've already asked you that question before and you said we wouldn't report such errors. Only errors relevant to linted files would be caught.

src/resolvers/http-and-file.ts
```diff
@@ -0,0 +1,37 @@
...
+            if (err) reject(err);
+            resolve(data);
+          });
+        });
+      },
+    },
+  },
+
+  parseAuthorityResult: async opts => {
+    const parts = opts.targetAuthority.path().split('.');
```
P0lip: Using path.extname is likely to be a better idea here.

philsturgeon: A yeah I can see this causing trouble.

marbemac:
Ok ready for another review. In answer to my question above, I think by default we just validate the provided file (not files that this file references). At a later date we could add an option to also validate $ref'd documents (would be off by default, so non breaking) - it will be tricky to get this UX right.

philsturgeon:
This didn’t update the README and after this discussion I’m not sure what actually changed. If we are not validating the contents of external refs then what are we doing?

philsturgeon:
@P0lip hey this still had "WIP - still needs tests and docs." on it, so we dont have tests, or docs. Could you maybe switch to doing this and put a pause on request maker?

P0lip:
@philsturgeon we wanted to have browser testing first, didn't we?

marbemac:
> This didn’t update the README and after this discussion I’m not sure what actually changed. If we are not validating the contents of external refs then what are we doing?

We're resolving file and http(s) $refs now, and reporting an error if they cannot be resolved. See this comment for the "do we validate things that this file references" question: https://github.com/stoplightio/spectral/pull/245#issuecomment-498899833

In general, that is a larger task and should probably be handled separately. It kind of changes an underlying assumption that we're validating one file at a time. We'll need to re-think how we track and report errors, since now errors (and their line numbers etc) can span across multiple files in a single spectral run.

cc @philsturgeon

src/cli/commands/lint.ts
```diff
@@ -137,8 +140,15 @@ async function lint(name: string, flags: any, command: Lint, rules?: RuleCollect
   if (flags.verbose) {
     command.log(`Linting ${name}`);
   }
-  const spec: IParserResult = await readParsable(name, flags.encoding);
-  const spectral = new Spectral();
+
+  let targetUri = name;
+  if (!startsWith(name, 'http')) {
```
philsturgeon: @P0lip looks like if name is `"httpModels/somefile.json" it will think it is a URL. Can we switch to`http(s)://` regex in the test PR you are working on? #264

philsturgeon:
Ok we've made a CLI specific issue for this. Thank you!

src/cli/commands/lint.ts
P0lip: @philsturgeon sure, I'll make that happen.

Issue #242, opened 2019-06-03
Can't use regex modifiers with the pattern function in YAML rulesets


arno-di-loreto:
### **I'm submitting a...**
  - [x] bug report
  - [ ] feature request

### What is the current behavior?

When using the pattern function in a YAML ruleset with spectral-cli, regex with modifiers (`/^.*dto$/i` for example) do not work.
They work when using the spectral module and defining ruleset in node.

YAML ruleset and swagger 2.0 YAML file to reproduce the bug can be found here: https://gist.github.com/arno-di-loreto/63c307e14d8a01928b98625191d59bb7

Not working in YAML ruleset
```
rules:
  schema-name-is-valid-not-working:
    summary: Definition name must not use dto suffix
    given: $.definitions
    severity: 0
    then:
      field: "@key"
      function: pattern
      functionOptions:
        notMatch: /^.*dto$/i
```

Working in node:
```
const rules = () => {
  return { 
    'definition-name-is-valid': {
      summary: 'Definition name must not use dto suffix',
      given: '$.definitions',
      type: RuleType.STYLE,
      severity: DiagnosticSeverity.Error,
      then: {
        field: '@key',
        function: RuleFunction.PATTERN,
        functionOptions: {
           notMatch: /^.*dto$/i
        }
      }
    }
}
```

### What is the expected behavior?

Being able to use regex with modifiers with the pattern function (for both match and notMatch options) in YAML rulesets.

### Please tell us about your environment:

  - spectral-cli -v: @stoplight/spectral/3.0.3 win32-x64 node-v8.11.3

### Other information

I think that when the ruleset is loaded, the /^.*dto$/i regex is loaded as a simple string and not a regex.

Proposed modification in  https://github.com/stoplightio/spectral/blob/master/src/functions/pattern.ts

- Analyzing the match and notMatch regexes to extract regex and modifiers
- Then use new Regex(regex, modifiers)

(Idea based on this stackoverflow response: https://stackoverflow.com/a/874742, I'm not sure about the regex to parse regex, maybe a simple substring would be simpler)

I think I can do it myself and propose a pull request if it is actually a bug (I may have miss something)

P0lip:
@arno-di-loreto 

It's a bug - we should be supporting all possible modifiers, as they are very commonly used. Unfortunately, JS regular expressions cannot be represented in any way in YAML or JSON format, we need to construct a regular expression during the runtime execution based on provided input. 
[...]

PR #251, opened 2019-06-07
fix(pattern.ts): enable use of regex flags for pattern in ruleset files

arno-di-loreto:
#242

Bug fix

See #242

None


src/functions/pattern.ts
```diff
@@ -1,5 +1,25 @@
 import { IFunction, IFunctionResult, IRulePatternOptions } from '../types';
 
+function test(value: string, regex: any) {
+  let re;
+  if (typeof regex === 'string') {
+    // regex in a string like {"match": "/[a-b]+/im"} or {"match": "[a-b]+"} in a json ruleset
+    // the available flags are "gimsuy" as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
+    const splitRegex = /^\/(.+)\/([gimsuy]*)$/.exec(regex);
```
P0lip: I wouldn't be less lenient about flags here.
```suggestion
    const splitRegex = /^\/(.+)\/([a-z]*)$/.exec(regex);
```
IMHO we don't have to be so defensive here. Furthermore, we won't need to adjust the code in the future when a new flag is introduced. (`s` is fairly new for instance, introduced in ES2018 IIRC)

Moreover, we don't seem to allow regular expressions without any flags specified, so I'd suggest tweaking the above regular expression further
```suggestion
    const splitRegex = /^\/(.+)\/([a-z]*)?$/.exec(regex);
```

src/functions/pattern.ts
```diff
@@ -1,5 +1,25 @@
 import { IFunction, IFunctionResult, IRulePatternOptions } from '../types';
 
+function test(value: string, regex: any) {
```
P0lip:
```suggestion
function test(value: string, regex: RegExp | string) {
```

src/functions/pattern.ts
```diff
@@ -1,5 +1,25 @@
...
+  let re;
+  if (typeof regex === 'string') {
+    // regex in a string like {"match": "/[a-b]+/im"} or {"match": "[a-b]+"} in a json ruleset
+    // the available flags are "gimsuy" as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
+    const splitRegex = /^\/(.+)\/([gimsuy]*)$/.exec(regex);
+    if (splitRegex) {
+      // with slashes like /[a-b]+/ and possibly with flags like /[a-b]+/im
+      re = new RegExp(splitRegex[1], splitRegex[2]);
```
P0lip: Syntax error caused by `/foo/iii` is uncaught here. Shall we wrap it with try...catch?
```suggestion
      try {
        re = new RegExp(splitRegex[1], splitRegex[2]);
      } catch {
        return false;
      }
```

arno-di-loreto: By silently catching errors, users wouldn't be aware that their regex is invalid. I would prefer to keep a behavior similar to what happens when you provide an invalid schema to the schema function (the exception is thrown, that really helped me 😅 ).
The following output shows how it looks when the schema and pattern function throw an exception:
```
Adding OpenAPI 2.0 (Swagger) functions
Encountered error when running rule 'invalid-regex-modifiers' on node at path '$,definitions':
SyntaxError: Invalid flags supplied to RegExp constructor 'izorglub'
Encountered error when running rule 'test-schema-error' on node at path '$,definitions,ValidSchemaName':
Error: schema is invalid: data/type should be equal to one of the allowed values, data/type should be array, data/type should match some schema in anyOf

/Users/arnaud/Dev/spectral-testing/samples/pattern-swagger.yaml
 23:24  error  schema-name-is-valid-not-working      Definitionname must not end with dto (any case)
 23:24  error  schema-name-is-valid-ulgy-workaround  Definitionname must not end with dto (any case)

✖ 2 problems (2 errors, 0 warnings, 0 infos)
```
If you're OK with that I just need to update the test to modify the one with an invalid flag. Can I do that on my branch and then use the `yarn commit` command without doing another pull request?

P0lip: > By silently catching errors, users wouldn't be aware that their regex is invalid. I would prefer to keep a behavior similar to what happens when you provide an invalid schema to the schema function (the exception is thrown, that really helped me sweat_smile ).

Agreed.

> If you're OK with that I just need to update the test to modify the one with an invalid flag. Can I do that on my branch and then use the yarn commit command without doing another pull request?

I'm totally fine with that. You can do it on your branch, sure.

src/functions/__tests__/pattern.test.ts
```diff
@@ -48,10 +48,14 @@ describe('pattern', () => {
...
     });
 
-    test('should return an array with a message when given value matches the given match string regex with slashes and modifier', () => {
-      expect(runPattern('aBc', { match: '/[abc]+/izorglug' })).toEqual([
-        { message: "must match the pattern '/[abc]+/izorglug'" },
-      ]);
+    test('should throw an exception when given string regex contains invalid flags', () => {
+      let exception = null;
+      try {
+        runPattern('aBc', { match: '/[abc]+/invalid' });
```
P0lip:
```ts
expect(() => runPattern('aBc', { match: '/[abc]+/invalid' })).toThrow(
  "Invalid flags supplied to RegExp constructor 'invalid'",
);
```

Could be simplified a bit :slightly_smiling_face:

P0lip:
I went ahead and made the change. Hope you don't mind.

PR #257, opened 2019-06-17
fix: switch to more lightweight jsonpath pkg

### LINKED ISSUE #214: Can not install due to jsonpath
### **I'm submitting a...**
  - [x] bug report
  - [ ] feature request

When trying to install spectral using `npm install @stoplight/spectral` I get the following error:

```
npm ERR! code 128
npm ERR! Command failed: C:\Program Files\Git\cmd\git.EXE clone --mirror -q https://github.com/stoplightio/jsonpath.git C:\Data\npm\cache\_cacache\tmp\git-clone-9c9fbadd\.git --config core.longpaths=true
npm ERR! warning: templates not found in C:\Users\USER\AppData\Local\Temp\pacote-git-template-tmp\git-clone-818f94da
npm ERR! fatal: unable to access 'https://github.com/stoplightio/jsonpath.git/': SSL certificate problem: self signed certificate in certificate chain
npm ERR!
```

I'm behind a corporate proxy which is why there is a self-signed certificate.

It installs without errors

Spectral is unusable for me without a fix. I literally can't download it via NPM.

  - Version: 1.x or above
  - OS: Windows 10
  - Package manager: NPM
  - Node version: 8.13.0

### Other information

Running things behind a corporate proxy is a known bad experience. In this case, things break because of the hard https reference to a forked repository.

Does this have to be a reference to the fork? What's wrong with `jsonpath@1.0.0`?

marbemac:
fixes #214 

The jsonpath implementation that we use right now (https://github.com/dchester/jsonpath) has a number of shortcomings. Most importantly, it does some weird things with esprima. This has caused some interop issues, and bloats the dependency size.

This PR switches the jsonpath implementation to https://github.com/s3u/JSONPath. Benefits:

- 13% the size when minified - 11kb versus 79kb.
- No more esprima or weird globals stuff.
- Significantly more performant - https://github.com/andykais/json-querying-performance-testing.


src/rulesets/oas2/__tests__/valid-example.ts
```diff
@@ -228,7 +228,7 @@ describe('valid-example', () => {
             "/pet",
             "post",
             "parameters",
-            0,
+            "0",
```
P0lip:
```suggestion
            "0",
```
This should be a number.

src/rulesets/oas3/__tests__/__snapshots__/server-trailing-slash.ts.snap
```diff
@@ -7,7 +7,7 @@ Array [
     "message": "Server URL should not have a trailing slash.",
     "path": Array [
       "servers",
-      0,
+      "0",
```
P0lip: Number, please. This might break `getJsonPathForLocation`.

src/rulesets/oas/__tests__/__snapshots__/tag-description.ts.snap
```diff
@@ -7,7 +7,7 @@ Array [
     "message": "Tag object should have a \`description\`.",
     "path": Array [
       "tags",
-      0,
+      "0",
```
P0lip: Number, please.

package.json
```diff
@@ -59,29 +59,28 @@
...
     "@oclif/config": "^1.12.11",
     "@oclif/plugin-help": "^2.0",
-    "@stoplight/json": "^2.1.0",
-    "@stoplight/json-ref-resolver": "^1.5.0",
-    "@stoplight/types": "^7.0.1",
-    "@stoplight/yaml": "^2.3.0",
+    "@stoplight/json": "2.x.x",
+    "@stoplight/json-ref-resolver": "1.5.x",
+    "@stoplight/types": "7.x.x",
+    "@stoplight/yaml": "2.x.x",
```
P0lip: I believe we'll want to switch to `^2.3.0`?
```suggestion
    "@stoplight/yaml": "^2.3.0",
```

package.json
```diff
@@ -59,29 +59,28 @@
     "@oclif/command": "^1.0",
     "@oclif/config": "^1.12.11",
     "@oclif/plugin-help": "^2.0",
-    "@stoplight/json": "^2.1.0",
-    "@stoplight/json-ref-resolver": "^1.5.0",
-    "@stoplight/types": "^7.0.1",
-    "@stoplight/yaml": "^2.3.0",
+    "@stoplight/json": "2.x.x",
```
P0lip: Ditto, let's use ^ syntax as Will described in Graphite PR.

src/rulesets/oas3/__tests__/server-not-example.ts
```diff
@@ -37,7 +37,7 @@ describe('server-not-example.com', () => {
       expect.objectContaining({
         code: 'server-not-example.com',
         message: 'Server URL should not point at `example.com`.',
-        path: ['servers', 0, 'url'],
+        path: ['servers', '0', 'url'],
```
P0lip:
```suggestion
        path: ['servers', 0, 'url'],
```

package.json
marbemac: That's a huge change that we should agree upon as a team, otherwise we'll end up being more inconsistent than we are now. Why ^ versus x? Can you link me to PR?

package.json
marbemac: Why? Only reason to do this is if we release breaking changes in stoplight/yaml without bumping the major version, which we should try not to do...

package.json
P0lip: https://github.com/stoplightio/graphite/pull/194#discussion_r294107274

src/rulesets/oas3/__tests__/__snapshots__/server-trailing-slash.ts.snap
P0lip: Fixed in https://github.com/stoplightio/json/pull/21

package.json
marbemac: Done

package.json
```diff
@@ -57,31 +57,30 @@
...
-    "ajv": "6.x.x",
-    "ajv-oai": "^1.1.1",
-    "chalk": "^2.4.2",
-    "jsonpath": "https://github.com/stoplightio/jsonpath#f1c0e9e634da2d45671b7639fa0a99afc807da17",
+    "@stoplight/json": "^2.2.2",
+    "@stoplight/json-ref-resolver": "^1.5",
+    "@stoplight/types": "^7.0",
+    "@stoplight/yaml": "^2.5",
+    "ajv": "^6",
+    "ajv-oai": "^1",
```
P0lip:
```suggestion
    "ajv-oai": "^1.1.1",
```
Suggest reverting this one as well.

marbemac: Got it - do we need something from ajv-oai 1.1.1? just curious, can make the change as nothing downstream will likely use this dep, so can be more specific with it.

package.json
```diff
@@ -57,31 +57,30 @@
...
+    "chalk": "^2.4",
```
wmhilton:
```suggestion
    "chalk": "^2.4.2",
```

package.json
P0lip: We need this commit https://github.com/BiteBit/ajv-oai/commit/6f7fa263b210a61db731d3b038f93ed0ccde369d.
It's on 1.1.0.

package.json
```diff
@@ -57,31 +57,30 @@
...
-    "@stoplight/json": "^2.1.0",
-    "@stoplight/json-ref-resolver": "^1.5.0",
-    "@stoplight/types": "^7.0.1",
-    "@stoplight/yaml": "^2.3.0",
-    "ajv": "6.x.x",
+    "@stoplight/json": "^2.2.2",
+    "@stoplight/json-ref-resolver": "^1.5",
+    "@stoplight/types": "^7.0",
+    "@stoplight/yaml": "^2.5",
+    "ajv": "^6",
```
philsturgeon: Is this intentional here? It would allow 7, 8 or 9 which could have breaking changes. We can avoid that with `^6.0`.

src/rulesets/oas/ruleset.json
```diff
@@ -321,7 +321,7 @@
     "operation-tags": {
       "summary": "Operation should have non-empty `tags` array.",
       "type": "style",
-      "given": "$..paths.*[?( name() === 'get' || name() === 'put' || name() === 'post' || name() === 'delete' || name() === 'options' || name() === 'head' || name() === 'patch' || name() === 'trace' )]",
+      "given": "$..paths.*[?( @property === 'get' || @property === 'put' || @property === 'post' || @property === 'delete' || @property === 'options' || @property === 'head' || @property === 'patch' || @property === 'trace' )]",
```
philsturgeon: This is a breaking change for anyone who has written their own styleguide, right?

philsturgeon:
Other than being a bit loose on AJV this looks good to me. I'm gld we have found an lternative to our fork! We have been talking about switching to [gitflow](https://stoplightio.atlassian.net/wiki/spaces/EN/pages/121602057/Open-Source) and now might be the time, as this is a very slight breaking change, and it might be best for us to put it together with a few other breaking changes for v4.0.

What do you folks think?

src/rulesets/oas/ruleset.json
P0lip: That's right.

P0lip:
Since we are very likely to have another breaking release, I'm removing the deprecated `maxResults` flag

philsturgeon:
@P0lip good catch, but I meant to suggest we make a develop branch now so we can avoid making v4 happen right now. We are smashing through major releases, about to launch v4 for no real reason, and we haven't even blogged about anything since v2.1. 

We should at least get #134 into v4 IMO.

P0lip:
> @P0lip good catch, but I meant to suggest we make a develop branch now so we can avoid making v4 happen right now. We are smashing through major releases, about to launch v4 for no real reason, and we haven't even blogged about anything since v2.1.

Yeah, I totally understand it and agree this PR should land in develop branch.
Having `breaking change` in the commit message will help us in the future, once we rebase master onto develop (never know whether one says develop onto master or master onto develop, but I mean `git rebase --onto master develop` :sweat_smile:).
I'm also not a fan of having releases at such pace. This might scare developers, as release frequency is sometimes treated as an indicator of package stability.

philsturgeon:
@marbemac I switched the base to develop, and sorry but there's loads of dependency PRs gone into develop. Hopefully no conflicts happen, if so I'll approve and this can go into our new develop branch. :)

philsturgeon:
Race conditions! I was just about to merge then this conflicted on the stoplight/types update haha.

philsturgeon:
@P0lip ok I fixed the conflicts and also changed the package.json from accidental major dependency upgrades. If this still looks ok, can you approve and merge?

P0lip:
@philsturgeon iiiiit's in!

philsturgeon:
Awesome work everyone! :D

package.json
marbemac: It actually only allows version 6 range (`^` = minor/patch can change). I recently found https://semver.npmjs.com/ which is awesome to understand this stuff.

philsturgeon: Stupid memory. I was mixing it up with ~ sorry.

PR #264, opened 2019-06-19
test: cover external $refs

P0lip:

Test

### Other information


src/cli/commands/__tests__/lint.test.ts
```diff
@@ -100,6 +101,18 @@ describe('lint', () => {
...
 
+  describe('when a spec contains external references', () => {
+    test
+      .stdout()
+      .command(['lint', oas2RefSpecPath])
+      .it('outputs invalid references', ctx => {
+        expect(ctx.stdout).toContain(
+          '"email" property can\'t resolve reference ./schema-ref.yaml#/definitions/email-d from id #',
+        );
+        expect(ctx.stdout).toContain("should have required property 'type'");
```
P0lip: @philsturgeon this assertion fails, we don't validate external schema.
This is unexpected, right? We'd like to have it validated as well.

philsturgeon: Yep this s outdated. You mentioned that #134 actually just hid these errors, and the end goal for v4 is to actually lint those external files, so either way this is not correct.

PR #271, opened 2019-06-25
feat: process referenced files

P0lip:

Feature

Currently, Spectral does not perform any extensive checks on `$ref` properties.
Although there are a few rules making use of resolved object literals, it's fair to say this is an exception and in general, we don't actually process external contents in any way. This behavior is not a bug yet is a result of a decision taken in the past, as the solution was considerably good and it was quite a fair compromise to make given we didn't resolve external refs by default. Now, when external refs are resolved by default, this became an issue, as we don't report any feedback...

... the purpose of the following PR is to change it and start reporting errors for referenced files as they were actually an integral part of a processed file.

Yeah, new resolver.

Use SpectralResolver rather than Resolver.

### Other information


marbemac:
Hmm, how can we get this done without requiring a customized instance of resolver (the SpectralResolver thing)?

Graphite/Studio have their standard own resolver instances that they pass into Spectral. They do unique things like read latest file contents from memory (so that can see spectral issues before saving the file), have all the lookups already cached, etc.

P0lip:
@marbemac 
I'm afraid it's impossible without touching json-ref-resolver. If `refMap` in resolved result contained nested refs and their contents, I could get rid of SpectralResolver.

marbemac:
Probably wouldn't be very difficult, I could get that done. What would `refMap` ideally look like? Could you share a small example with some realistic data?

P0lip:
@marbemac 

The biggest obstacle I faced with refMap is its lack of any support for nested refs.
At the moment, the scope of refMap seems to be limited to top-level' refs and does not contain any nested refs found in referenced files. That's why I have to follow them manually.
Obviously, the major showstopper is lack of resolved content in raw format - json-ref-resolver does not really need it, it's sufficient for it to have an object literal, but Spectral does need an actual raw format to map given error to a physical place in a file (line and character).

To sum up:
- refMap has to contain all refs & ideally point straight to final (tail) ref.
- unparsed content has to be exposed

P0lip:
@marbemac any update on that?

P0lip:
```
[...] Instead of having the registry I added, we could simply load the files manually in Studio and then have some interceptor in Spectral to use require rather than fs.readFile or any other method operating on fs.
```
I'll try to make that work right now.

marbemac:
Let's perhaps hop on a call tomorrow about this.

src/functions/schema.ts
```diff
@@ -60,11 +64,13 @@ export const schema: IFunction<ISchemaOptions> = (targetVal, opts, paths) => {
     const validator = validators.get(schemaObj);
     if (!validator(targetVal) && validator.errors) {
       results.push(
-        ...(betterAjvErrors(schemaObj, targetVal, validator.errors, { format: 'js' }) as IOutputError[]).map(
-          ({ error }) => ({
-            message: cleanAJVErrorMessage(error),
-            path,
-          }),
+        ...(betterAjvErrors(schemaObj, targetVal, validator.errors, { format: 'js' }) as IAJVOutputError[]).map(
+          ({ error, path: errorPath }) => {
```
P0lip: Note to self: add a few tests with `path` assertions. The vast majority of tests covering ajv-based validation are mainly oriented on checking messages, but a valid `path` is crucial as well.

src/rulesets/oas2/__tests__/valid-example.ts
```diff
@@ -154,8 +154,8 @@ describe('valid-example', () => {
     expect(results).toEqual([
       expect.objectContaining({
         code: 'valid-example',
-        message: '"halRoot" property type should be array',
-        path: ['definitions', 'halRoot'],
+        message: '"self" property type should be array',
+        path: ['definitions', 'halRoot', '_links', 'self'],
```
P0lip: Urgh, it was stupid of me. We did have a proper path here I erroneously changed in a different PR. It's now on par with [master](https://github.com/stoplightio/spectral/blob/master/src/rulesets/oas2/__tests__/valid-example.ts#L155L159) and should stay so.

P0lip:
@marbemac yeah, I've added an extra unit test concerning paths, but will try to toss one more in, though.

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -17,6 +17,8 @@ const validRulesetConfigPath = resolve(__dirname, '__fixtures__/config.ruleset.y
 const invalidRulesetConfigPath = resolve(__dirname, '__fixtures__/config.ruleset.invalid.yml');
 const standardOas3RulesetPath = resolve(__dirname, '../../../rulesets/oas3/ruleset.json');
 const standardOas2RulesetPath = resolve(__dirname, '../../../rulesets/oas2/ruleset.json');
+const draftRefSpec = resolve(__dirname, './__fixtures__/draft-ref.oas2.json');
```
philsturgeon: I don't understand what "draft" means in this context?

P0lip: I couldn't think of anything more meaningful.
It's just an almost empty spec, hence the draft in the name, but happy to rename it.

philsturgeon: ahh right. Most of these fixtures are bare minimal API descriptions so no need to differentiate any how.

src/spectral.ts
```diff
@@ -1,11 +1,20 @@
...
 import { Resolver } from '@stoplight/json-ref-resolver';
-import { IResolveError } from '@stoplight/json-ref-resolver/types';
-import { DiagnosticSeverity, IParserResult } from '@stoplight/types';
-import { getLocationForJsonPath, parseWithPointers } from '@stoplight/yaml';
-import { merge, uniqBy } from 'lodash';
-
+import { DiagnosticSeverity, Dictionary, IParserResult } from '@stoplight/types';
+import {
+  getLocationForJsonPath,
+  getLocationForJsonPath as getLocationForJsonPathYAML,
```
philsturgeon: What is all this about?  :D

P0lip: Oops, oversight, no need to double the imports.

PR #274, opened 2019-06-27
Upload artifacts on release

XVincentX:
This PR copies some automation I did for Prism to create and upload the artifacts on the GitHub release page and add an install.sh script for ease of consumption.

P.S: The Windows binary is broken. We need to use Azure Pipelines to produce a good one.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,23 @@ jobs:
       - cached-dependencies
       - run: yarn build
       - run: yarn release
+      - run: yarn build.binary
+      - persist_to_workspace:
+          root: ./dist
+          paths:
+            - bin
+  upload_artifacts:
+    docker:
+      - image: circleci/golang:1-stretch
+    steps:
+      - attach_workspace:
+          at: /tmp/
+      - run:
+          name: "Download GitHub Release Utility"
+          command: go get github.com/tcnksm/ghr
+      - run:
+          name: "Publish Release artifacts on GitHub"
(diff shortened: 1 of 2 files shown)


XVincentX:
I'm going to need somebody to notice me next time you're releasing on master so I can help/check out in the (rare) case something goes wrong.

philsturgeon:
I'm going to ta a beta very soon, as soon as I work out how we're doing that. Prism is lerna based so I have to figure out how to do it non-lerna. Lets chat on slack a bit.

philsturgeon:
oooo wait I found one incorrect thing. Only one though! 

After negotiator block it says "Success?" then the No goes up to "takes lowest 2xx" but 2xx is for success. What are we trying to say here?

.circleci/config.yml
```diff
@@ -86,3 +103,9 @@ workflows:
           filters:
             branches:
               only: master
+      - upload_artifacts:
+          filters:
+            branches:
+              only: master
```
philsturgeon: @XVincentX If we're only doing this on master, will `${CIRCLE_TAG}` have a value? Will ghr know what version number this is?

XVincentX: Oh this is a very good point — I think for the way semantic release works the tag is never tested on CircleCI, we're going to need to find another way around it. I think we can get away by doing a simple `git tag` and grab the latest one since the release has happened already.

philsturgeon: Latest is no good we might release v4.0 then 43.2.1

PR #275, opened 2019-06-27
feat: use latest openapi v3.0 metaschema

philsturgeon:

Kinda a feature I suppose. OpenAPI v3.0 has a proper metaschema now and have done for a few months, but we're still using the one grabbed from the PR a loooong time ago.

Validates OpenAPI 3.0 description files properly I hope.

It should validate openapi 3.0 :D

### DIFF
--- a/src/rulesets/oas3/schemas/main.json
+++ b/src/rulesets/oas3/schemas/main.json
@@ -1,6 +1,13 @@
 {
+  "id": "https://spec.openapis.org/oas/3.0/schema/2019-04-02",
+  "$schema": "http://json-schema.org/draft-04/schema#",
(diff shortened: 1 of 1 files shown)


src/rulesets/oas3/schemas/main.json
```diff
@@ -376,13 +402,21 @@
         },
         "enum": {
           "type": "array",
-          "items": {},
+          "items": {
+          },
           "minItems": 1,
-          "uniqueItems": true
+          "uniqueItems": false
```
marbemac: This is interesting - why would anybody ever want duplicate values in their enum?

philsturgeon: Great question, I'll ask!

PR #289, opened 2019-07-05
release: fixes

XVincentX:
1. Remove packages that do not make sense
2. Make sure pkg has all the data it needs to build a working binary
3. Do not recreate the package.json with weird stuff
4. Use `npm version from-git` so we just need a tag to release
5. Create a Docker image (this is significantly easier than Prism)


.gitignore
```diff
@@ -32,3 +32,5 @@
 *.log*
 *-debug.log*
 *-error.log*
+oclif.manifest.json
+binaries
```
P0lip:
```suggestion
/binaries
```

XVincentX:
@P0lip I wish — Prism uses kind of a different schema since it's a monorepo and follows different rules  . I've basically re-applied the knowledge I gained while automatic Express Gateway.

.circleci/config.yml
```diff
@@ -102,9 +103,13 @@ workflows:
       - release:
           filters:
             branches:
-              only: master
+              ignore: /.*/
+            tags:
+              only: /^v.*/
```
P0lip:
```suggestion
              only: /^v\.*/
```
Pretty sure you want to match a dot here.

.circleci/config.yml
```diff
@@ -102,9 +103,13 @@ workflows:
...
       - upload_artifacts:
           filters:
+            branches:
+              ignore: /.*/
             tags:
               only: /^v.*/
```
P0lip: Ditto?

PR #292, opened 2019-07-05
fix: add bin directory

XVincentX:


.circleci/config.yml
```diff
@@ -86,9 +86,9 @@ jobs:
       - run: "docker build -t stoplight/spectral:${CIRCLE_TAG} -t stoplight/spectral:latest --build-arg EG_VERSION=latest ."
       - run: docker push stoplight/spectral:${CIRCLE_TAG}
       - run: |
-          if [ $CIRCLE_TAG = *beta* ] ; then
+          if [[ "$CIRCLE_TAG" = *beta* ]] ; then
```
P0lip: How about the other condition you have in the file?

XVincentX: Yeah, good point

PR #298, opened 2019-07-05
fix: rule with human-readable severity should be considered as valid

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/src/meta/rule.schema.json
+++ b/src/meta/rule.schema.json
@@ -49,13 +49,20 @@
             "type": "string"
         },
         "severity": {
-            "enum": [
-                0,
-                1,
-                2,
-                3
-            ],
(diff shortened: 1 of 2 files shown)


src/meta/rule.schema.json
```diff
@@ -49,13 +49,20 @@
...
-                0,
-                1,
-                2,
-                3
-            ],
-            "type": "number"
+            "oneOf": [
+                {
+                    "enum": [
+                        0,
```
marbemac: Technically I guess it can be `-1` for `off`?

P0lip: Correct. I will add it.

PR #299, opened 2019-07-05
docs: SVG for the readme and a quick go at cli

philsturgeon:
This is not quite ready, we need to make sure there are docs for each line in the Examples bullet point list, but it took me too long to make that damn SVG. Can somebody else pick this up on Monday?


XVincentX:
@tbarn This should be good to review now, looks complete.

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
-        given: $..parameters[*]
-        then:
-            field: description
-            function: truthy
+  operation-2xx-response: warn
 ```
 
+The example above will run the recommended rules from the `spectral:oas2` ruleset, but report `operation-2xx-response` as a warning rather than as an error (as is the default behavior in the `spectral:oas2` ruleset).
+
+Available severity levels are `error`, `warn`, `info`, and `off`.
```
P0lip:
```suggestion
Available severity levels are `error`, `warn`, `info`, `hint` and `off`.
```

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+
 ## Rules
```
P0lip: https://github.com/stoplightio/spectral/blob/77cb641d7deea9c24dc672019efdd635f7ad3c88/docs/rulesets.md#rules it's updated, we should include human-readable severity levels (off, warn, error etc)

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
 
 Rules are highly configurable. There are only few required parameters but the optional ones gives powerful flexibility. Please see the following type tables for more information.
```
P0lip: I am not sure if that's on purpose or not, but we are missing an example with functionOptions.
```
rules:
  operation-2xx-response: [warn, { foo: bar }]
```
^ we should pass something more reasonable than this dummy object

tbarn:
I'll review this today. Doing it in context with what's going in the blog post so it wasn't a quick review.

README.md
```diff
@@ -5,117 +5,93 @@
...
 
-### Local Installation
-
-```bash
-npm install @stoplight/spectral
-```
+Spectral's CLI can be installed either via npm or yarn
 
-### Global Installation
+### npm
```
tbarn: this section is npm and yarn, but you changed it to npm 😛

README.md
```diff
@@ -5,117 +5,93 @@
...
-  -h, --help                   show CLI help
-  -m, --maxResults=maxResults  deprecated: use --max-results instead
-  -o, --output=output          output to a file instead of stdout
-  -q, --quiet                  no logging - output only
-  -r, --ruleset=ruleset        path to a ruleset file (supports remote files)
-  -s, --skip-rule=skip-rule    ignore certain rules if they are causing trouble
-  -v, --verbose                increase verbosity
-  --max-results=max-results    [default: all] maximum results to show
-```
+Spectral can be installed and used via your terminal. Take a look at the [CLI docs](./docs/cli.md).
```
tbarn: "terminal" -> "command-line"

README.md
```diff
@@ -5,117 +5,93 @@
...
+- **Rules** filter your object down to a set of target values, and specify the function that is used to evaluate those values.
+- **Functions** accept a value and return issue(s) if the value is incorrect.
 
-Think of a set of **rules** and **functions** as a flexible and customizable style guide for your JSON objects.
+Think of a **Spectral ruleset** as a flexible and customizable style guide for your JSON objects.
 
 ## Programmatic usage
 
-Spectral is written in TypeScript (JavaScript) and can be used directly for when you need to use Spectral programmatically. Take a look at our ["JavaScript API documentation"](docs/js-api.md).
+Spectral is written in TypeScript (the compiled to JavaScript) and can be used directly for when you need to use Spectral programmatically. Take a look at our ["JavaScript API documentation"](docs/js-api.md).
```
tbarn: "the" -> "then"

README.md
```diff
@@ -5,117 +5,93 @@
...
+#### I want to lint my OpenAPI documents but don't want to implement Spectral right now.
 
 No problem! A hosted version of Spectral comes **free** with the Stoplight platform. Sign up for a free account [here](https://stoplight.io/?utm_source=github&utm_campaign=spectral).
 
-**What is the difference between Spectral and [Speccy](https://github.com/wework/speccy)?**
+#### What is the difference between Spectral and [Speccy](https://github.com/wework/speccy)?
 
 With Spectral, lint rules can be applied to _any_ JSON object. Speccy is designed to work with OpenAPI v3 only. The rule structure is different between the two. Spectral uses [JSONPath](http://goessner.net/articles/JsonPath/) `path` parameters instead of the `object` parameters (which are OpenAPI specific). Rules are also more clearly defined (thanks to TypeScript typings) and now require specifying a `type` parameter. Some rule types have been enhanced to be a little more flexible along with being able to create your own rules based on the built-in and custom functions.
 
+### Executable binaries
```
tbarn: Why not include this with all the other installations?

docs/cli.md
```diff
@@ -0,0 +1,25 @@
+# Spectral CLI
+
+One installed, Spectral can be run via the command-line:
```
tbarn: "One" -> "After"

docs/cli.md
```diff
@@ -0,0 +1,25 @@
...
+``` text
+  -e, --encoding=encoding      text encoding to use
+  -f, --format=json|stylish    formatter to use for outputting results
+  -h, --help                   show CLI help
+  -o, --output=output          output to a file instead of stdout
+  -q, --quiet                  no logging - output only
+  -r, --ruleset=ruleset        path to a ruleset file (supports remote files)
+  -s, --skip-rule=skip-rule    ignore certain rules if they are causing trouble
+  -v, --verbose                increase verbosity
+  --max-results=max-results    [default: all] maximum results to show
```
tbarn: max-results is deprecated, right?

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
-- Documentation
-- Rulesets
----
+# Spectral Rulesets
 
-Rulesets are a container for collections of rules. These rules are essentially calling functions Spectral, and by taking parameters you can make these functions do whatever you want. 
+Rulesets are a container for collections of rules. These rules are taking parameters, and calling functions on certain parts of the JSON object being linted.
 
-First the rule filters your object (JSON/YAML file) down to a set of target values, and then list what function arguments should be passed in.
+## CLI Usage
```
tbarn: Maybe link to the CLI docs here too

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
 
-## Usage
+When you run the `spectral lint my-document.json` CLI command, Spectral will automatically apply the built in OpenAPI v2 or v3 ruleset if appropriate.
 
-```bash
-spectral lint foo.yaml --ruleset=path/to/acme-company-ruleset.yaml --ruleset=http://example.com/acme-common-ruleset.yaml
+To customize the rules that are applied, simply create a `spectral.yml` in the same directory that you are running the `spectral lint` command from and it will automatically be used.
```
tbarn: Remove the word "simply"

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+    given: $.tags[*]
+    severity: error
+    then:
+      field: description
+      function: truthy
+```
+
+The example above adds a single rule that checks that tags objects have a description property defined.
+
+Running `spectral lint` on the following object with the rulset above will result in an error being reported, since the tag does not have a description:
```
tbarn: "rulset" -> "ruleset"

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+
+```json
+{
+  "tags": [{
+    "name": "animals"
+  }]
+}
+```
+
+While running it with this object will succeed:
```
tbarn: I think this is missing a word or two between "object" and "will"

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+    then:
+      field: description
+      function: truthy
+```
+
+The example above will apply the recommended rules from the built in OpenAPI v2 ruleset AND apply the custom `my-rule-name` rule.
+
+### Enabling specific OpenAPI rules
+
+Sometimes you might want to apply specific rules from another ruleset. Use the `extends` property, and pass `off` as the second argument in order to load add the rules from another ruleset, but disable them all by default. This allows you to pick and choose which rules you would like to enable.
```
tbarn: in "load add" remove the word "add"

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+**spectral.yml**
+
+```yaml
+extends: [[spectral:oas2, off]]
+rules:
+  # This rule is defined in the spectral:oas2 ruleset. We're passing `true` to turn it on and inherit the severity defined in the spectral:oas2 ruleset.
+  operation-operationId-unique: true
+```
+
+The example above will run the two rules we enabled, since we pass `off` to disable all rules by default when extending the `spectral:oas2` ruleset.
```
tbarn: I only see one rule, not two, so I am confused.

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+### Disabling specific OpenAPI rules
+
+This example shows the opposite of the `Enabling specific OpenAPI rules` example. Sometimes you might want to enable all rules by default, and disable a few.
+
+**spectral.yml**
+
+```yaml
+extends: [[spectral:oas2, all]]
+
+# we could also just extend spectral:oas2 like this, which would only run the recommended rules (rather than every single rule)
```
tbarn: I'm confused on how I will know the recommended rules.

tbarn:
One more thing I just thought of while working on the blog post, do we have an example of using a URL with the CLI anywhere? Or at least mention it is possible? Not just a local file.

docs/cli.md
P0lip: It's removed in https://github.com/stoplightio/spectral/pull/345

docs/rulesets.md
```diff
@@ -1,41 +1,128 @@
...
+# extends: spectral:oas2
+
+rules:
+  operation-operationId-unique: false
 ```
 
-## Example Ruleset
+The example above will run all of the rules defined in the `spectral:oas2` ruleset (rather than the default behavior that runs just the recommended ones), with two exceptions - we turned `operation-operationId-unique` and `operation-2xx-response` off.
```
tbarn: only see one rule in the example, but two mentioned here.

docs/rulesets.md
marbemac: A great question - we should probably add another article or two that render the openapi v2 and v3 rulesets in a table or something. One of the table columns could indicate if the rule is "recommended" or not.

docs/rulesets.md
marbemac: Reads as intended to me - what do you suggest? Have to read it in the context of the whole section.

README.md
marbemac: It's just a longer section that's less relevant to most folks (majority I assume will use NPM or Docker). I don't have a strong opinion on this though, we can move it back up if you want.

marbemac:
> One more thing I just thought of while working on the blog post, do we have an example of using a URL with the CLI anywhere? Or at least mention it is possible? Not just a local file.

The `cli.md` article kind of mentions it in `-r, --ruleset=ruleset        path to a ruleset file (supports remote files)`, but yeah it's not very obvious. Feel free to add a note about it somewhere :).

Overall I think we're in a decent spot. We should probably document the built in rulesets at some point though... could be separate PR. @P0lip could you double check the `js-api.md` article to make sure it's all still accurate (feel free to clean it up if anything is irrelevant or you have other ideas for improvement while you're at it)?

docs/rulesets.md
marbemac: Kind of on purpose - I think we wait to encourage more advanced use of functions until we have time to really think them through. In either case, we can add more docs for functions in a separate PR.

docs/rulesets.md
tbarn: "While running it with this object, it will succeed:"

P0lip:
@marbemac we are good in terms of JS API documentation. I executed every example and it worked fine. 
We are missing `summary` in each result, but it's not a big deal.
I haven't changed JS API much, just replaced generic lodash merge with our merging function.

P0lip:
...unless we decide to drop addRules/addFunctions as a part of this release and replace it with `loadConfig`.
@marbemac whatcha think? I'd say, let's do it now, might be less painful, but on the other hand - Graphite would be affected, so the change would certainly come at some cost.

marbemac:
> unless we decide to drop addRules/addFunctions as a part of this release and replace it with loadConfig

Would be something like `loadRuleset`, not config ;). Regardless, I think we should wait. The theme of the v4 Spectral release is looking like rulesets, while v5 could focus on programmatic usage.

philsturgeon:
Let’s not drop anything until v5 it’s past feature freeze time!

--
Phil Sturgeon
@philsturgeon

README.md
```diff
@@ -5,117 +5,91 @@
...
-Now, `spectral` command will be accessible in your terminal.
-
-Head over to [releases](https://github.com/stoplightio/spectral/releases) for the latest binaries.
+Spectral is avaiable as a Docker image as well.
 
-
-### Docker
 ```bash
-docker run --rm -it stoplight/spectral lint "${URL}"`
+docker run --rm -it stoplight/spectral lint "${url or filepath to openapi file}"`
```
XVincentX: This is not correct, in case you want to use a file on your local computer you effectively need to mount a volume in the container. I'll change the command.

README.md
XVincentX: This makes me wonder a little bit why we're offering binaries

XVincentX:
> we are good in terms of JS API documentation. I executed every example and it worked fine.

Testing our README examples for real would be really cool, and there's a tool that does that! @P0lip 

https://byexamples.github.io/byexample/

P0lip:
Marc, yeah, `loadRulest` obviously. :sweat_smile: 

@XVincentX 
Yeah, good idea. Would be actually nice to have the output data automatically generated.
Should be quite easy to do, but I would need an hour.

docs/rulesets.md
tbarn: Even if it just means pointing to a file for now where they are, it is better than them being kind of a mystery to the user.

docs/rulesets.md
XVincentX: Do you mean on the title itself?
```suggestion
## [CLI Usage](./cli.md)
```

docs/rulesets.md
XVincentX: My understanding is that this is kind of a continuation of the previous sentence:

```
running spectral with this will fail…
…
…
while running it with this object will succeed
```

therefore, to me it looks correct. @tbarn ?

docs/rulesets.md
XVincentX: Done!

docs/rulesets.md
tbarn: Usually I avoid links in headers, I should have specified "here" as in that section somewhere.

docs/rulesets.md
tbarn: It doesn't read as a sentence. You should see every new line as a new sentence in docs since people's eyes skip around and will miss context.

docs/rulesets.md
XVincentX: Makes sense. So let's put a sentence such as 

> To get an overview of the CLI chek out the [documentation page](link)

?

docs/rulesets.md
XVincentX: All right!

XVincentX: 32ba8be

docs/rulesets.md
tbarn: 👍 

(minus "chek" ;) )

XVincentX:
Done!

PR #301, opened 2019-07-06
feat: include recomended rules by default

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

There are a couple of changes included in this PR.
The most notable ones are:
- unspecified severity in extends is considered recommended now
- stricter normalization of diagnostic, to make sure we work with DiagnosticSeverity severity level during the linting stage rather than DiagnosticSeverity and HumanReadableSeverity. The reason I didn't do it earlier is the fact we didn't have `-1` specified anywhere, therefore we returned `off` everywhere.
- extracted a few utils related to severity
- minor changes in interfaces
- better severity level handling
- moar tests


src/cli/commands/__tests__/lint.test.ts
```diff
@@ -218,7 +218,7 @@ describe('lint', () => {
           expect(ctx.stdout).toContain(
             '46:24  warning  operation-description   Operation `description` must be present and non-empty string',
           );
-          expect(ctx.stdout).toContain('22 problems (0 errors, 22 warnings, 0 infos)');
+          expect(ctx.stdout).toContain('28 problems (0 errors, 28 warnings, 0 infos)');
```
marbemac: Why did this go up?

src/rulesets/__tests__/merger.unit.ts
```diff
@@ -58,7 +59,7 @@ describe('Rulesets merger', () => {
       },
     });
 
-    expect(ruleset).toHaveProperty('rules.test.severity', 'off');
+    expect(ruleset).toHaveProperty('rules.test.severity', -1);
```
marbemac: Should we just add `-1` to `DiagnosticSeverity` for consistency?

src/rulesets/__tests__/reader.test.ts
```diff
@@ -47,13 +48,53 @@ describe('Rulesets reader', () => {
     });
   });
 
-  it('should inherit properties of extended rulesets', () => {
-    return expect(readRulesFromRulesets(extendsOas2Ruleset)).resolves.toEqual({
+  it('should inherit properties of extended rulesets', async () => {
+    const rules = await readRulesFromRulesets(extendsOas2Ruleset);
+
+    // we pick up recommended rules only from spectral:oas
```
marbemac: Could you elaborate a bit more re what this test is checking?

We might have missed a nuance in the original implementation description. The initial ruleset extends should probably determine the whole thing. For example:

If `base-rulset2.json` has `extends: ['base-ruleset1.json', 'all']`, and `my-ruleset.json` has `extends: ['base-ruleset2.json', 'off']`, running spectral with the `my-ruleset.json` should result all rules being off (from all the extended rulesets). Otherwise it would be pretty confusing to ask for recommended rules and not get them, or to ask for extended rules to start off as off, and not have that be the case, etc.

Thoughts?

src/runner.ts
```diff
@@ -13,12 +14,12 @@ export const runRules = (
   let results: IRuleResult[] = [];
 
   for (const name in rules) {
-    if (!rules.hasOwnProperty(name)) continue;
+    if (!Object.prototype.hasOwnProperty.call(rules, name)) continue;
```
marbemac: What does this fix?

src/runner.ts
```diff
@@ -13,12 +14,12 @@ export const runRules = (
...
 
     const rule = rules[name];
     if (!rule) continue;
 
-    if (rule.hasOwnProperty('recommended') && !rule.recommended) {
+    if (rule.severity !== undefined && getDiagnosticSeverity(rule.severity) === -1) {
```
marbemac: Does `if (getDiagnosticSeverity(rule.severity) === -1) {` not work?

src/rulesets/__tests__/reader.test.ts
P0lip: I think we might have missed it. The current implementation applies the ruleset severity only to direct dependencies (in other words it's one level deep), therefore it won't override the ruleset severity declared in extended rulesets.
I fully agree with you this is very likely to cause confusion. I'll adjust the implementation.

src/rulesets/__tests__/merger.unit.ts
P0lip: I considered doing that, but AFAIR our initial intention was to follow LSP and I believe they don't have such a level, would need to check.

src/runner.ts
P0lip: It's not a fix, it's just a slightly safer approach as `ownHasProperty` can be overridden, in theory. Obviously it's not a case here, so I'm eager to revert the change.

src/runner.ts
P0lip: Yeah, it does not expect undefined value to be provided. While it's unlikely rule.severity will be undefined at this very point as we normalize all rules beforehand, it's just safer in case rules haven't been processed by merger for some reason, i.e. they were added somehow manually using API. I suppose we could get rid of this once we drop addRules method as this is potentially the only way where a breakage could be introduced.

src/rulesets/__tests__/merger.unit.ts
```diff
@@ -142,7 +143,7 @@ describe('Rulesets merger', () => {
       },
     });
 
-    expect(ruleset).toHaveProperty('rules.test.severity', -1);
+    expect(ruleset).toHaveProperty('rules.test.severity', DiagnosticSeverity.Error);
```
P0lip: My apologies, but I had to make this change in order to make the following case work properly
```
extends: [[spectral:oas2, off]]
rules:
  operation-operationId-unique: true
```
Actually, if you dive into it, we have two causes of a severity change:
- the rule severity (the severity you put on rule)
- the ruleset severity (the severity you put on ruleset via extends)

This design a flaw:
spectral:oas2 *turns off* rules of spectral:oas, so severity of  `operation-operationId-unique` is set to -1
Then, we inherit severity value, therefore it's -1.

Now that I think of it, we could potentially bring back the `enabled` property to handle `off` just for the sake of normalization process, but this is still very tricky.

Let me know what your thoughts are @marbemac.

src/cli/commands/__tests__/lint.test.ts
P0lip: It's because of the fix included here https://github.com/stoplightio/spectral/pull/336.
Basically, we couldn't evaluate oas-schema rue properly in some cases, therefore we didn't 
report all errors/warnings
The change is unrelated to this PR, I simply included the code since I worked on that at the same time.
The code is already merged, so this code is no longer present on this branch.
Is that clear now?

P0lip: If you revert the change introduced in 336, you will get 22 problems again.

src/runner.ts
marbemac: Let's revert it :).

src/rulesets/__tests__/merger.unit.ts
marbemac: Interesting - I don't think we should block the release for this. Folks can put in a severity for now, rather than true.

P0lip: Well, I do support `true` and `false`, but to a certain extent.
When you provide `true` we take the severity from the original/first rule definition.
I believe this is a safe assumption, for now, to make, if one wants to have a particular severity, one can always set it.

PR #343, opened 2019-07-08
feat: offload docker image creation to dockerhub

XVincentX:


XVincentX:
It's now configurable from docker hub directly so the tagging strategy can be changed from there with no code changes.

PR #345, opened 2019-07-08
feat(cli): remove --max-results flag

P0lip:
BREAKING CHANGE: --max-results flag is gone.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

### DIFF
--- a/src/cli/commands/lint.ts
+++ b/src/cli/commands/lint.ts
@@ -49,9 +49,6 @@ linting ./openapi.yaml
       char: 'o',
       description: 'output to a file instead of stdout',
     }),
-    'max-results': flagHelpers.integer({
-      description: '[default: all] maximum results to show',
-    }),
     ruleset: flagHelpers.string({
       char: 'r',
       description: 'path to a ruleset file (supports remote files)',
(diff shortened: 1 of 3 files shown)


src/cli/commands/lint.ts
```diff
@@ -217,38 +214,40 @@ async function lint(name: string, flags: any, command: Lint, rules?: RuleCollect
...
   const skippedRules: string[] = [];
   const invalidRules: string[] = [];
 
-  for (const rule of flags.skipRule) {
-    if (rule in rules) {
-      delete rules[rule];
-      skippedRules.push(rule);
-    } else {
-      invalidRules.push(rule);
+  if (flags.skipRule !== undefined) {
```
XVincentX: Hmm, was this working before? Ah ok it was just passing because of the `any` as type, right? 🤦‍♂

P0lip: Yeah, we had `any` everywhere. I guess we normalize that data *somewhere*, but I wouldn't assume we do that for every scenario.

PR #346, opened 2019-07-08
test(lint): revert previous assertion

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

Yeah, I know previous looked a bit different, but it didn't make much sense.
Right now it's slightly better and what's more important consistent for both CRLF and LF line endings.

### DIFF
--- a/src/cli/commands/__tests__/lint.test.ts
+++ b/src/cli/commands/__tests__/lint.test.ts
@@ -202,8 +202,7 @@ describe('lint', () => {
         .command(['lint', invalidOas3SpecPath, '-r', standardOas3RulesetPath])
         .it('outputs warnings in default format', ctx => {
           expect(ctx.stdout).toContain(
-            'warning  api-servers       OpenAPI `servers` must be present and non-empty array',
-            // Restore to once Windows new line is sorted '1:5  warning  api-servers       OpenAPI `servers` must be present and non-empty array',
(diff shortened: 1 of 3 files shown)


src/cli/commands/__tests__/lint.test.ts
```diff
@@ -202,8 +202,7 @@ describe('lint', () => {
         .command(['lint', invalidOas3SpecPath, '-r', standardOas3RulesetPath])
         .it('outputs warnings in default format', ctx => {
           expect(ctx.stdout).toContain(
-            'warning  api-servers       OpenAPI `servers` must be present and non-empty array',
-            // Restore to once Windows new line is sorted '1:5  warning  api-servers       OpenAPI `servers` must be present and non-empty array',
+            '1:1  warning  api-servers       OpenAPI `servers` must be present and non-empty array',
```
XVincentX: 1:1? Wasn't it 1:5?

P0lip: image
:laughing:

P0lip: TFW you try to add a description and he doesn't read it anyway :trollface:

XVincentX: Yeah the template is so misleading that seems like nobody uses it and I just skip it.

PR #348, opened 2019-07-09
Merge latest master in develop

XVincentX:


philsturgeon:
We shouldn’t need to do this, do we know why we did? Things should only be going into master when a release happens, and we will update develop then. If stuff is going into master arbitrarily it will lead to conflicts with releases.

PR #353, opened 2019-07-09
build: inline pkg version

P0lip:
Fixes #[ISSUE_NUMBER].

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

:scream:

Let me explain...
In a browser environment without access to any kind of fs, Spectral will try to obtain rulesets from npm. This approach is fairly fine, but comes with a downside - you cannot change the location of rulesets if you don't want to break Spectral for users. This PR addresses this issue thanks to specifying the actual version of a given release.
I didn't want to rely on `package.json` on purpose - it would require `require` :laughing: or imports and so on. You know. The solution proposed here should, in theory, be more reliable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,7 @@ jobs:
       - cached-dependencies
       - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
       - run: npm version from-git
+      - run: yarn inline-version
       - run: yarn build
       - run: |
           if [[ "$CIRCLE_TAG" == *beta* ]] ; then
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "test.update": "yarn test --updateSnapshot",
     "test.watch": "yarn test --watch",
     "test.karma": "karma start",
+    "inline-version": "./scripts/inline-version.js",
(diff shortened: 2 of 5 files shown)


wmhilton:
I..... I know using AST is your style, but... wouldn't something [like this](https://github.com/isomorphic-git/isomorphic-git/blob/master/__tests__/__helpers__/fix-version-number.js) be more robust and easier to maintain?

```js

#! /usr/bin/env node
var pkg = require('../../package.json')
var replace = require('replace-in-file')
var options = {
  files: ['src/**/*.ts'],
  from: /INJECT_SPECTRAL_VERSION_HERE/g,
  to: pkg.version
}
;(async function () {
  await replace(options)
})()
```

P0lip:
@wmhilton 
As you know, I am not a fan of modifying a code via replace. 
You know, If we needed to inject the code in multiple places, I'd likely use such a replace or write a more robust solution relying on AST modifications, but hm. It just this particular place for now.
If we happen to find a need for supporting more places, we can easily throw some walker in and make a  modification on every matching string.

P0lip:
We are short on time, therefore I am going to merging the PR now, but we can always revert this in future or revisit the approach.
I'm up to the discussion, just leave a comment here.

PR #387, opened 2019-07-16
Release v4.0.1

philsturgeon:
Two reviews required as this is going to master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,24 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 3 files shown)


XVincentX:
@P0lip This bump was required because, since we're installing the dep from Git directly the build script isn't run, right?

philsturgeon:
@XVincentX that's right. Check the CHANGELOG I just pushed, it'll explain it.

XVincentX:
We could really use the GitHub registry package to quickly publish our forks :(

philsturgeon:
Make github issues for any cool ideas you have and/or start a slack thread :D

P0lip:
> @P0lip This bump was required because, since we're installing the dep from Git directly the build script isn't run, right?

No, it's the other way around actually. We do not want to run the build script, since it's supposed to fail on Windows. Moreover, we ship bundled code, so there is no valid reason to run the script given the built files are already accessible.

PR #392, opened 2019-07-17
Updated develop with v4.0.1 changes

philsturgeon:
This is an annoying part of git flow on github when protected branches are involved. We merge hotfixes into master, then we have to get things back into develop. I'm looking into automation for this, but for now... thumbs please!


philsturgeon:
@tbarn im into that, its something I've done before, but didn't have the time during this hotfix to do it. We can start on the next planned release, or if somebody wants to send a followup PR they can.

PR #406, opened 2019-07-26
new rule: no anyOf & oneOf in OpenAPI v2

### LINKED ISSUE #393: new rules: error for anyOf & oneOf in OpenAPI v2
**User story.**

As an average user who is not completely familiar with the intricacies of JSON Schema <≠> OpenAPI and the differences between OpenAPI v2 and v3, I often mix up JSON Schema, OAS2 and OAS3 keywords like anyOf and oneOf, so please let me know when I put the wrong keyword in the wrong place.

API description documents should conform to the description format in question, as putting invalid keywords in those documents can lead to inconsistent behavior across various tools that might mistakenly support those keywords. Other tooling which is actually compliant will then look like its failing.

The rule should error if an `anyOf`, `oneOf` or `not` appears in OpenAPI v2, but `anyOf` is in fact allowed.

philsturgeon:
Fixes #393.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**

Screen Shot 2019-07-26 at 15 31 33

**Additional context**

### DIFF
--- a/src/cli/commands/__tests__/lint.test.ts
+++ b/src/cli/commands/__tests__/lint.test.ts
@@ -337,47 +337,49 @@ describe('lint', () => {
       .stdout()
       .command(['lint', draftRefSpec, '-q', '-f=json'])
       .it('outputs errors occurring in referenced files', ctx => {
-        expect(JSON.parse(ctx.stdout)).toEqual([
-          expect.objectContaining({
-            code: 'info-description',
-            message: 'OpenAPI object info `description` must be present and non-empty string.',
-            path: ['info', 'description'], // todo: relative path or absolute path? there is no such path in linted ref, but there is such in spec when working on resolved file
-            range: {
-              end: {
-                character: 22,
-                line: 5,
+        expect(JSON.parse(ctx.stdout)).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({
+              code: 'info-description',
+              message: 'OpenAPI object info `description` must be present and non-empty string.',
+              path: ['info', 'description'], // todo: relative path or absolute path? there is no such path in linted ref, but there is such in spec when working on resolved file
+              range: {
+                end: {
+                  character: 22,
+                  line: 5,
+                },
+                start: {
+                  character: 21,
+                  line: 5,
+                },
               },
-              start: {
-                character: 21,
-                line: 5,
+              source: expect.stringContaining('src/cli/commands/__tests__/__fixtures__/refs/info.json'),
+            }),
+            expect.objectContaining({
+              code: 'api-schemes',
+              message: 'OpenAPI host `schemes` must be present and non-empty array.',
+              path: ['schemes'],
+              range: expect.any(Object),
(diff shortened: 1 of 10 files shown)


src/rulesets/oas2/index.json
```diff
@@ -103,8 +103,36 @@
...
+        "field": "anyOf",
+        "function": "undefined"
+      },
+      "tags": [
+        "schema"
+      ]
+    },
+    "oas2-oneOf": {
+      "description": "OpenAPI v3 keyword `oneOf` detected in OpenAPI v2 document.",
+      "message": "anyOf is not available in OpenAPI v2, it was added in OpenAPI v3",
```
P0lip:
```suggestion
      "message": "oneOf is not available in OpenAPI v2, it was added in OpenAPI v3",
```

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -337,47 +337,49 @@ describe('lint', () => {
...
-          }),
-          expect.objectContaining({
-            code: 'oas2-schema',
-            message: '/info Property foo is not expected to be here',
-            path: ['info'],
-            range: {
-              end: {
-                character: 5,
-                line: 9,
+        expect(JSON.parse(ctx.stdout)).toEqual(
```
P0lip: Let's rever this change.

src/functions/__tests__/falsy.test.ts
```diff
@@ -0,0 +1,38 @@
...
+describe('falsy', () => {
+  test('returns undefined if target value is falsy', () => {
+    expect(runFalsy(false)).toBeUndefined();
+  });
+
+  test('returns undefined if target value is null', () => {
+    expect(runFalsy(null)).toBeUndefined();
+  });
+
+  test('returns error message if target value is falsy', () => {
```
P0lip:
```suggestion
  test('returns error message if target value is not falsy', () => {
```

src/cli/commands/__tests__/lint.test.ts
philsturgeon: Why? Reasons for things should be included on review comments. We gotta know why we're doing stuff.

src/functions/__tests__/falsy.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(runFalsy(true)).toEqual([
+      {
+        message: 'property is not falsy',
+      },
+    ]);
+  });
+
+  test('returns undefined if target path is set', () => {
```
P0lip: Let's add another test case with a path.

src/cli/commands/__tests__/lint.test.ts
P0lip: You haven't touched any of existing fixtures, therefore this change should not be needed.
We actually did change the assertion here, as toEqual array might produce different results than  toEqual array containing (which IIRC matches just a subset of an array and items can be placed in any order)

PR #413, opened 2019-07-31
perf: improve paths lookup

P0lip:
- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

`paths` is a top-level property, so there is no reason to perform a deep scanning, which is utterly slow for larger specs.


philsturgeon:
We don’t use semantic release for spectral :)

Issue #325, opened 2019-07-06
Switch from oclif to yargs


brianmrock:
oclif has been causing all sorts of oddities and weirdness, including a recent issue where it starts moaning about globby for no reason. 

```
$ spectral --version
(node:30001) [MODULE_NOT_FOUND] Error Plugin: @stoplight/spectral: Cannot find module 'globby'
module: @oclif/config@1.13.0
task: not loading commands, globby not found
plugin: @stoplight/spectral
root: /Users/phil/.asdf/installs/nodejs/10.15.3/.npm/lib/node_modules/@stoplight/spectral
See more details with DEBUG=*
@stoplight/spectral/4.0.0-beta1 darwin-x64 node-v10.15.3
```
Switch it out for Yargs as Prism has done. 

**Acceptance Criteria**
- No breaking changes to the CLI (e.g: its ok if help page looks a bit different)
- Oclif is gone
- Lint command is still tested, coverage going up would be nice

PR #430, opened 2019-08-10
feat: custom functions

### LINKED ISSUE #311: Custom functions in Rulesets
**User Story**

As a ruleset maintainer, I would like to be able to define my own functions, to achieve things the built in functions might not allow. How that is done I do not much care, so long as it is easy, and the rulesets remain portable as simple YAML or JSON files.

**Details**

Both the CLI and the JS API are converging on a simple `loadRulesets()` method (#366), leaving behind the separate `loadRules()` and `loadFunctions()` methods that have been there forever. 

This will make things easier for Studio, and all JS API users, but we still need a way to support custom functions. Those functions should be defined in, or at the very least referenced by, the ruleset.

A complementary feature that needs to be delivered as a part of this story is the function options schema. In other words - we need to have to configure the function without falling back to JS code too much.

**Implementation**

```
{
  // recommended rules are on, using their default severity values
  // it is expected that an `index.json` file results at the target location if the path does not end in `.json`
  "extends": "spectral:oas2",

  // OPTIONAL
  // use to customize where functions are located, by default they are located in the `functions` dir relative to this ruleset file
  "functionsDir": "functions",

  // OPTIONAL - only needed if defining new functions for your ruleset
  "functions": {
    // convention is that there must be a corresponding `functions/op-id-unique.js` file relative to this config file's location
    "op-id-unique": {
      // OPTIONAL the json schema that describes the function options, if any
      "schema": {
        "type": "object",
        "properties": {
          "foo": {
            "type": "string"
          }
        }
      }
    }
  },

  "rules": {
    // define a new rule that uses a the custom function defined in this config (op-id-unique)
    "my-rule": {
      "summary": "Every operation must have a unique `operationId`.",
      "type": "validation",
      "severity": "error",
      "given": "$",
      "then": {
        "function": "op-id-unique",
        "functionOptions": {
          "foo": "hello"
        }
      }
    }
  }
}
```

When we switch to doing this, lets implement this approach for all of our core functions, and clean out all of the old TypeScript definitions for them, so there is no different between core functions and custom functions.

```
export interface ILengthRuleOptions {
  min?: number;
  max?: number;
}
export type LengthRule = IRule<RuleFunction.LENGTH, ILengthRuleOptions>;
```

We can also clean out the function collections in our rulesets index.ts: 

```
export const commonOasFunctions = (): FunctionCollection => {
  return {
    oasPathParam: require('./functions/oasPathParam').oasPathParam,
    oasOp2xxResponse: require('./functions/oasOp2xxResponse').oasOp2xxResponse,
    oasOpSecurityDefined: require('./functions/oasOpSecurityDefined').oasOpSecurityDefined, // used in oas2/oas3 differently see their rulesets for details
    oasOpIdUnique: require('./functions/oasOpIdUnique').oasOpIdUnique,
    oasOpFormDataConsumeCheck: require('./functions/oasOpFormDataConsumeCheck').oasOpFormDataConsumeCheck,
    oasOpParams: require('./functions/oasOpParams').oasOpParams,
  };
};
```

P0lip:
Closes #311

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

Here it is.
The PR is likely to introduce a breaking change, as I shuffled some methods around and perhaps removed a few.
There is still cleanup left to be done, mostly around the usage of deprecated methods in our code as well as tests.
I'm not really sure whether we are good with making breaking changes, as we are not planning to release 5.0 anytime soon, and the code is meant to be shipped as a part of 4.1, so I suspect I might need to revert a few changes and be more careful with the cleanup.

The code is ready to be reviewed, I covered all the cases that I could think of.
I'll write a word on the updates I had to make to tests as well as the rulesets functions.

A note on functions.
This is the final syntax I implemented:
```json
"functions": [
"foo",
["bar", { "type": "string" }]
]
```
There are a couple of fixtures and tests covering these, so you can take a look to get a grasp of that in action.
The syntax follows the other JS-based tooling such as Babel etc., therefore we should be aligned with others.
The object pattern felt very awkward in practice.

@philsturgeon 
I believe the custom functions deserve their own document, thoughts?

Moreover, I'm planning to start a developer guide or something similar, as everything gets more complicated and you need to be wary of certain patterns.

I'm planning to do that in the scope of this PR, but beforehand, if you could at least take a look at the code and the functions concept I eventually went with, that would be great.


src/meta/ruleset.schema.json
```diff
@@ -31,6 +31,22 @@
...
                 ]
             }
+        },
+        "functionsDir": {
+            "type": "string"
+        },
+        "functions": {
+            "type": "array",
+            "items": {
+                "anyOf": [
```
philsturgeon: oneOf as it can’t be both.

src/cli/commands/lint.ts
```diff
@@ -14,7 +14,7 @@ import { getDefaultRulesetFile } from '../../rulesets/loader';
 import { isOpenApiv2, isOpenApiv3 } from '../../rulesets/lookups';
 import { oas2Functions, rules as oas2Rules } from '../../rulesets/oas2';
```
philsturgeon: Hopefully we can clear out this old stuff too? I would expect that with loading rulesets, and rulesets being able to have functions, that we dont need all this oas2Functions, oas3Functions, oas2Rules, oas3Rules, etc? Especially seeing as oas2Functions and oas3Functions are identical. 

We should move our tests over to then new stuff to avoid divergence between what we're testing and what we're suggesting people use.

philsturgeon:
Part of the plan with Spectral v4.1 was to add loadRulesets and custom functions, but something I didn't emphasize enough was that this was meant to be used throughout Spectral itself. We have some rather confusing code for our internal rulesets, and loadRulesets, formats, etc is meant to make it a bit more simple.

People should be able to include `spectral:oas` and have it load _all_ the rules, without needing to specify oas2 or oas3 (thanks formats!) but the way our oas2 and oas3 are build don't allow for it.

We should be able to have the current oas2 ruleset look like this:

```
  "extends": ["spectral:oas2", "spectral:oas3"],
  "formats": ["oas2", "oas3"],
  "rules": {
    "operation-2xx-response": {
```

Then oas2 and oas3 specific stuff can stay where it is. This will involve a bunch of change in our tests. 

It would be ok to split this work into another issue and follow up PR, but it would still need to be done for the 4.1 launch, otherwise we are suggesting features we dont use, and carrying around this extra baggage instead of using our nice new functionality. :)

philsturgeon:
Other cleanup, lets get rid of this:

```
  if (parseInt(spec.data.swagger) === 2) {
    command.log('Adding OpenAPI 2.0 (Swagger) functions');
    spectral.addFunctions(oas2Functions());
  } else if (parseInt(spec.data.openapi) === 3) {
    command.log('Adding OpenAPI 3.x functions');
  }
```

They're the same thing anyway! We don't need to check the object to detect format, formats do that, and now functions are gonna be loaded through the ruleset right?

philsturgeon:
One last thing, we are getting deprecation warnings in the CLI because we use this old stuff:

```
▶ ./bin/run lint pets-mini.yaml --fail-severity=error
Adding OpenAPI 2.0 (Swagger) functions
Method `addRules` has been deprecated since version 4.1, use `loadRuleset` instead.
    at /Users/phil/src/spectral/src/cli/commands/lint.ts:213:12
OpenAPI 2.0 (Swagger) detected
```

:D

P0lip:
@philsturgeon I'm aware of deprecation warnings, but we don't have any alternative solutions just yet :thinking: We'd really need to get that `setRuleset` thingy in.

P0lip:
The code should be pretty much ready. I will just add a few extra test cases and write docs, so hold on with the reviews yet.

P0lip:
Docs added :sweat_smile:

P0lip:
@XVincentX https://dev.azure.com/vncz/vncz/_build/results?buildId=1117 any clue why azure jobs get stuck at `$ copyfiles -u 1 "dist/rulesets/oas*/functions/*.js" ./` all the time?

docs/getting-started/development.md
```diff
@@ -0,0 +1,38 @@
...
+
+It's nice to see you here. Every contribution is always more than welcome and very appreciated.
+The current development process for Spectral might seem to be complicated and make you feel overwhelmed.
+
+We prepared a common list of gotchas and things to keep in mind when writing any code.
+
+## Introduction
+
+Although, it might not be very obvious, Spectral is meant to be run in any ES2017 and above JS environment, therefore there are a few rules you need to be aware of.
+One of the main principle is to be very strict on using any Node.js exclusive modules, so if you can help it, avoid using it.
```
philsturgeon:
```suggestion
One of the main principle is to avoid using any Node.js exclusive modules, as these will not work when Spectral is run in the browser. See if there is another way to achieve the same goal without using Node.js-only code.
```

docs/getting-started/development.md
```diff
@@ -0,0 +1,38 @@
...
+The same applies to globals specific to particular environment, such as fetch in a browser, etc.
+In general, Spectral should run in a similar manner regardless of the environment it's executed in.
```
philsturgeon: I think you need new lines between paragraphs in order for markdown to know they are different paragraphs dont you?

docs/getting-started/development.md
```diff
@@ -0,0 +1,38 @@
...
+If you feel like there is really no other way than to use a Node module, i.e. a fs call is needed, you need to make sure to provide an alternative solution for other environments.
+Obviously, there are a few exceptions from that rule as for instance CLI usage of Spectral yet that portion of code is only executed in Node.js, 
+thus no particular restrictions apply.
+The rest of the code is likely to be executed in various environments, so please be be wary of that and be very mindful.
```
philsturgeon: I think the point was made above, this is redundant now.

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,96 @@ rules:
...
 Available severity levels are `error`, `warn`, `info`, `hint`, and `off`.
+
+
+### Adding a custom function
+
+If the built-in functions are not enough, Spectral gives you a possibility of providing your own ones.
+Previously, this used to be possible only in case of programmatic usage of Spectral, but there is no such limitation anymore,
+and functions are an integral part of Spectral rulesets.
+
+A custom function might be any JS function compliant with IFunction type.
```
philsturgeon: Where does IFunction come from?

src/rulesets/reader.ts
```diff
@@ -45,28 +50,64 @@ async function readRulesFromRuleset(
...
+
+  if (rulesetFunctions !== void 0) {
+    const rulesetFunctionsBaseDir = join(
+      rulesetUri,
+      ruleset.functionsDir !== void 0 ? join('..', ruleset.functionsDir) : '../functions',
+    );
+    const resolvedFunctions: FunctionCollection = {};
+
+    await Promise.all(
+      rulesetFunctions.map(async fn => {
```
philsturgeon: does this map with promises work? I seem to remember failing with this and seeing articles about it not working as expected, but im certainly not an expert on this.

philsturgeon:
So begin the conflicts!

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,96 @@ rules:
...
+Besides that, we encourage you to not transpile the code to ES5 if you can help it. Spectral does not support older environments than ES2017, so there is no need to bloat the bundle with useless transformations and polyfills.
+Ship untransformed async/await, do not include unneeded shims, it's all good.
+
+Another caveat is that ES Modules and other modules systems are not supported. Although, you are recommended to write ES2017 code, you should not be using require or imports.
+We don't support it, if you have any module system, you need to use some bundler, preferably Rollup.js as it generates efficient bundles.
+One note here - we are still evaluating that idea and perhaps we bring support for ES Modules at some point.
+
+#### How do I create a function?
+
+The process should be quite hassle-free.
```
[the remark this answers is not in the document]
P0lip: Is not not hassle-free? :trollface:

docs/getting-started/development.md
P0lip: Yeah, though I just added that whitespace to make writing easier :sweat_smile: 
Will make it easier to read now!

src/rulesets/reader.ts
P0lip: Yeah, in general you are right.
If any promise fails, the other might not be executed at all, and the whole `Promise.all` promise is rejected.

I am not really sure how we want to handle that case.
If one function cannot be loaded, some rules won't work correctly, yet the others should work just fine, so there is no need to throw entirely.
That said, we might want to wrap each async function here `rulesetFunctions.map(async fn => {`  with a try catch.

P0lip: https://github.com/stoplightio/spectral/pull/430/commits/7a99652ae4de81532212f7413597648421a3791b

src/rulesets/__tests__/validation.test.ts
```diff
@@ -153,4 +154,116 @@ describe('Ruleset Validation', () => {
...
+    ).not.toThrow();
+  });
+
+  it('recognizes invalid functions directory', () => {
+    expect(
+      assertValidRuleset.bind(null, {
+        functionsDir: 2,
+        rules: {},
+      }),
+    ).toThrow();
```
philsturgeon: Do we care what its throwing? Exception names can be considered part of the interface so if an exception type changes its actually a breaking change, and would mean we need to bump a major under SemVer. Should we check?

P0lip: These errors are schema errors reported by AJV, yet I can check for them so that we'll know that a new version of ajv changes something :+1:

P0lip: It'd look more or less like this
```ts
it('recognizes invalid schema functions', () => {
    expect(
      assertValidRuleset.bind(null, {
        functions: ['d', { typo: 'a' }],
        rules: {},
      }),
    ).toThrow(`/functions/1 should be string
/functions/1 should be array
/functions/1 should match exactly one schema in oneOf`);
  });
```

docs/getting-started/development.md
```diff
@@ -0,0 +1,39 @@
+# Development
```
marbemac: Does this file overlap with the traditional `contributing.md` file?

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
 The example above will run the recommended rules from the `spectral:oas2` ruleset, but report `operation-2xx-response` as a warning rather than as an error (as is the default behavior in the `spectral:oas2` ruleset).
 
 Available severity levels are `error`, `warn`, `info`, `hint`, and `off`.
+
+
+### Adding a custom function
+
+If the built-in functions are not enough, Spectral gives you a possibility of providing your own ones.
```
marbemac: > If the built-in functions are not enough, Spectral allows you to write and use your own custom functions.

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+Previously, this used to be possible only in case of programmatic usage of Spectral, but there is no such limitation anymore,
```
marbemac: Suggest removing this entire "previously..." sentence - it's not really relevant to anybody and will become stale basically immediately.

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+and functions are an integral part of Spectral rulesets.
+
+A custom function might be any JS function compliant with [IFunction](https://github.com/stoplightio/spectral/blob/90a0864863fa232bf367a26dace61fd9f93198db/src/types/function.ts#L3#L8) type.
```
marbemac: Suggest linking to where the file will be once this is merged - otherwise will have to open another PR to fix the link.

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+  targetValue: any,
+  options: O,
+  paths: IFunctionPaths,
+  otherValues: IFunctionValues,
+) => void | IFunctionResult[];
+```
+
+#### Introduction
+
+It takes exactly the same arguments as built-in functions do, so you are more than welcome to take a look at the existing implementation.
```
marbemac: What is "It"?

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+functions: [abc]
+rules:
+  my-rule:
+    message: "{{error}}"
+    given: "$.info"
+    then:
+      function: "abc"
+```
+
+Optionally, if you'd like to validate the data that it passed to abc function before the function gets executed, you can provide a JSON Schema.
```
marbemac: This is basically documenting the options that the function accepts, right? Not validating the incoming `targetVal`. Assuming this, suggest re-wording to make it more clear.

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+    then:
+      function: "abc"
+```
+
+Spectral would look for functions in a `my-functions` directory now.
+
+#### Inheritance
+
+Core functions can be overridden with custom rulesets, so if you'd like to make your own truthy go ahead. 
+Other than that, custom functions are only available in the ruleset which defines them, so loading a foo in one ruleset will not clobber a foo in another ruleset.
```
marbemac: 🤔 what does this mean? If somebody creates a ruleset that has some snazzy custom functions, I cannot extend that ruleset and add some rules of my own that use those functions?

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+It takes exactly the same arguments as built-in functions do, so you are more than welcome to take a look at the existing implementation.
+
+Before you write your first custom function, you must be wary of certain limitations that will apply.
+
+First of all, the custom function must be synchronous.
+
+Apart from the above, Spectral is meant to support a variety of environments, so ideally your function should have an equal or at least very similar functionality in Node.js, browsers etc.
+That said, that's one of the core restriction, namely: do not rely on globals or functions specific to one particular environment.
+If you need to access environment specific API, make sure you provide an alternative for other environments.
+A good example of such situation is `fetch` - a function available natively in a browser context, but missing in Node.js.
```
marbemac: We're already loading a number of really useful utilities for core spectral stuff - should we just expose some of these libraries to custom functions as a libs argument or something, rather than requiring them to import and bundle them with each function? Stuff like ajv, fetch, lodash, and whatever else. This would at least cover the obvious helpful utils - if they need more exotic stuff they'd still have to import and bundle it etc.

Would also greatly reduce the bundled ruleset file, since those common util libraries would no longer need to be bundled in.

docs/getting-started/rulesets.md
P0lip: This is unchanged, though. I didn't touch `IFunction`, it's already on master.

docs/getting-started/rulesets.md
P0lip: You can (see tests). Note, this is not recommended - if you extend a few rulesets that implement the same method and change the order of those rulesets, a different function will be picked up.

P0lip: https://github.com/stoplightio/spectral/blob/2defef04b140d3e93179345315002b458936b3d5/src/rulesets/mergers/__tests__/functions.jest.test.ts
https://github.com/stoplightio/spectral/blob/2defef04b140d3e93179345315002b458936b3d5/src/__tests__/linter.jest.test.ts#L27#L76 
https://github.com/stoplightio/spectral/blob/2defef04b140d3e93179345315002b458936b3d5/src/rulesets/__tests__/reader.jest.test.ts#L357#L412

docs/getting-started/rulesets.md
P0lip: > We're already loading a number of really useful utilities for core spectral stuff - should we just expose some of these libraries to custom functions as a libs argument or something, rather than requiring them to import and bundle them with each function? Stuff like ajv, fetch, lodash, and whatever else. This would at least cover the obvious helpful utils - if they need more exotic stuff they'd still have to import and bundle it etc.

I'm very skeptical about that. 
A couple of issues I see with this approach:
- code might break anytime we bump the dependency. While, we update the code on our own, we cannot assume the end-developer will do the same every time we happen to do it.
- we will lose flexibility and won't be able to get rid of lodash, ajv, fetch and so on without releasing a breaking release - Studio users will face lots of troubles if their custom functions use any of those deps.

I'm happy to let users create a library bundle that could be injected to each function, but I'm against injecting our own dependencies for others to use.

docs/getting-started/rulesets.md
P0lip: It validates target.

docs/getting-started/development.md
P0lip: Not really, but I should link that doc in `contributing.md`

docs/getting-started/rulesets.md
philsturgeon: @P0lip Marc is suggesting that a link to it on master would be better. I'd agree, so long as we don't change it a bunch or move it.

docs/getting-started/rulesets.md
philsturgeon: @P0lip If you extend two rulesets which declare an `apples` function, and you try to use the apples function in your ruleset, which one do you get?

@marbemac same scenario but what would you expect?

docs/getting-started/rulesets.md
philsturgeon: Yeah let's keep our internal utilities internal. 

I am thinking we should create a `resolve` command, but we can talk about that elsewhere.

docs/getting-started/rulesets.md
P0lip: > @P0lip If you extend two rulesets which declare an apples function, and you try to use the apples function in your ruleset, which one do you get?

@philsturgeon the most recent one one. These 2 extended rulesets would have to its own `apples` function though.

src/__tests__/spectral.test.ts
```diff
@@ -1,10 +1,61 @@
...
+            return oasRules;
+          },
+          {},
+        ),
+      );
+    });
+
+    test('should support loading multiple built-in rulesets', async () => {
+      const s = new Spectral();
+      await s.loadRuleset('spectral:oas2', 'spectral:oas3');
```
philsturgeon: Can we get another test which is loading `spectral:oas` only, and show that both oas2 and oas3 rules are loaded? 

This is the final missing piece that will allow people to load one ruleset and get both, depending on which format is loaded. We probably should have got it done as part of #432 but it slipped by.

P0lip: It's done in https://github.com/stoplightio/spectral/pull/461

philsturgeon:
>  I believe the custom functions deserve their own document, thoughts?

Yes I believe so. It can be linked to from the rulesets guide, and be its own thing.

src/rulesets/__tests__/validation.test.ts
philsturgeon: Oh I meant the _type_ of error, so `AjvException` or whatever, if that's a thing in TS.

docs/getting-started/rulesets.md
```diff
@@ -199,3 +198,141 @@ rules:
...
+```ts
+export type IFunction<O = any> = (
+  targetValue: any,
+  options: O,
+  paths: IFunctionPaths,
+  otherValues: IFunctionValues,
+) => void | IFunctionResult[];
+```
+
+#### Introduction
```
philsturgeon: We can't have an introduction that starts on line 220. Maybe this will be improved by moving custom functions to their own document, but an introduction should just be the start of writing about a topic and doesnt need a header.

src/__tests__/linter.test.ts
```diff
@@ -152,16 +150,11 @@ describe('linter', () => {
...
-      { rules: oas3Rules },
-      {
-        rules: {
-          'valid-example': 'off',
-          'model-description': -1,
-        },
-      },
-    ).rules as RuleCollection);
+    const { rules: oas3Rules } = await readRuleset('spectral:oas3');
+    spectral.addRules(mergeRules(oas3Rules, {
```
philsturgeon: I'm adding setRules in #325 if you just wanna grab that? might be a bit easier. I use it to update the rules after skipRules() in the CLI.

src/rulesets/oas2/index.ts
```diff
@@ -1,5 +1,5 @@
-import { readRulesFromRulesets } from '../reader';
+import { readRuleset } from '../reader';
 
 export { commonOasFunctions as oas2Functions } from '../oas';
```
philsturgeon: Do we need these files at all? If we keep them do we still need `commonOasFunctions`? I thought this PR would remove them.

src/rulesets/oas/index.ts
```diff
@@ -1,5 +1,5 @@
 import { FunctionCollection } from '../../types';
-import { readRulesFromRulesets } from '../reader';
+import { readRuleset } from '../reader';
 
 export const commonOasFunctions = (): FunctionCollection => {
```
philsturgeon: This definitely seems like a thing we should be able to delete?

src/rulesets/oas2/index.ts
P0lip: We can remove them, but if we do so, this will be breaking.

src/rulesets/oas/index.ts
P0lip: I am more than happy to remove it, but it will be breaking.
If there is a code relying on it (there certainly is - it must be used if you use Spectral programmatically), imports will break, hence the code will need to be updated.

src/__tests__/linter.test.ts
P0lip: Added.
I also have `setRuleset` in one of my other PRs, that we probably could recommend using.

docs/getting-started/development.md
marbemac: I feel like we should move the relevant bits of this over to contributing, and delete this file.

docs/guides/custom-functions.md
```diff
@@ -0,0 +1,131 @@
...
+functions: [abc]
+rules:
+  my-rule:
+    message: "{{error}}"
+    given: "$.info"
+    then:
+      function: "abc"
+```
+
+Optionally, if you'd like to validate the data that it passed to abc function before the function gets executed, you can provide a JSON Schema.
```
marbemac: Again, I'm not sure that this is what it should be doing?

We have a json schema rule that should be used to validate target data.

This JSON schema should describe the `functionOptions` available in this function. Eventually we'll use it to render documentation for the ruleset and custom functions, and I guess we could also use it to validate the `functionOptions` the end user passes in?

src/spectral.ts
```diff
@@ -143,74 +142,65 @@ export class Spectral {
...
-      const declaration = declarations[ruleName];
-
-      const rule = this.rules[ruleName];
-      if (rule) {
-        if (typeof declaration === 'boolean') {
-          this._rules[ruleName].recommended = declaration;
-        }
-      }
-    }
+  public async loadRuleset(...uris: string[]) {
```
marbemac: This is a very constrictive API - we'll never be able to change it or add any options without introducing a breaking change. I suggest something like this instead:

```ts
public async loadRulesets(uris: string[]) {
```

With this, we can easily add an optional `opts` or other argument after uris in the future, as needed. I also pluralized the method, since it supports multiple.

docs/guides/custom-functions.md
```diff
@@ -0,0 +1,131 @@
...
+    given: "$.info"
+    then:
+      function: "abc"
+```
+
+Spectral would look for functions in a `my-functions` directory now.
+
+## Inheritance
+
+Core functions can be overridden with custom rulesets, so if you'd like to make your own truthy go ahead. Custom functions are only available in the ruleset which defines them, so loading a foo in one ruleset will not clobber a foo in another ruleset.
```
marbemac: This is confusing. The `truthy` function is defined further up, but I can still use it in my ruleset. Can I not use any of the functions defined in the oas2 ruleset in my own ruleset file?

src/spectral.ts
P0lip: Yeah, you are right. I'll pull over the code I wrote herehttps://github.com/stoplightio/spectral/pull/460 it has a very similar APIs. I had to change it because of the reasons you mentioned.

docs/guides/custom-functions.md
P0lip: > This JSON schema should describe the functionOptions available in this function. Eventually we'll use it to render documentation for the ruleset and custom functions, and I guess we could also use it to validate the functionOptions the end user passes in?

Yeah, we validate the function options that are passed in.

docs/guides/custom-functions.md
P0lip: > Can I not use any of the functions defined in the oas2 ruleset in my own ruleset file?

You can use it.

P0lip: I'm not really sure how to explain it any better :thinking: 
As you could have all observed already, writing docs is not my biggest asset :laughing:

P0lip:
> I feel like we should move the relevant bits of this over to contributing, and delete this file.

It's up to you all.
If the information is not that useful we could even get rid of it entirely, so that we don't have to update it in future.
I wrote it because I thought it might be helpful for others, but if you believe it makes more harm than good, it's okay, I'm totally okay with removing it.

P0lip:
I feel like this is ready for another pass.
Regarding documentation - I suppose we can polish it in another PR.

docs/getting-started/rulesets.md
```diff
@@ -169,13 +168,12 @@ rules:
 
 The example above will run the single rule that we enabled, since we passed `off` to disable all rules by default when extending the `spectral:oas2` ruleset.
 
-### Disable Rules
+## Disabling rules
```
philsturgeon: Note that this breaks the readme

philsturgeon:
The docs you’ve written are excellent but need a little polish. I’ll take care of this in another PR on Monday, using the comments here to start and just giving it a general once over.

marbemac:
Ah, I was just polishing a few bits up - will open a separate PR :). 

Awesome to get this merged, great job @P0lip !

Issue #366, opened 2019-07-11
Load Rulesets via JS API 


philsturgeon:
**User story.**
As a developer working with the JS API, I can do reference a custom ruleset (path, url, required from a NPM module), so that I do not need to mess around with the addRules() and addFunctions() interface. 

_This ticket should deprecate those two methods, but is not responsible for adding custom functions to rulesets. That will be handled in_ https://github.com/stoplightio/spectral/issues/311

**Is your feature request related to a problem?**
V4.0 made it way easier to work with custom rulesets but working with the TS/JS interface still feels pretty clunky. 

**Describe the solution you'd like**

TBD on the exact interface. 
```
spectral.loadRuleset(‘file://something’)
```

**Additional context**

Relates to #365

marbemac:
Generally feels good - what about expecting end user to load/read the file themselves? So would look more like this:

```
spectral.setRuleset(ruleset: IRuleset)
```

Where IRuleset is https://github.com/stoplightio/spectral/blob/develop/src/types/ruleset.ts#L17.

I guess we could also have `loadRuleset` that takes a file or url, fetches it, and calls `setRuleset` with the result.

OR since ruleset extends supports files and urls, might be able to get away with something as simple as this for `loadRuleset`?

```ts
function loadRuleset(rulesetLocation: string) {
  spectral.setRuleset({
      extends: rulesetLocation,
  });
}
```

Kinda snazzy if that just works :).

P0lip:
@marbemac 
I believe we should support both - a plain ruleset path/name and a parsed ruleset.
This is kind of similar to what we do with spec - you can provide a parsed spec, but also a string if you'd like to.

The problem with the ruleset approach only is that API consumer would need to call `resolveRuleset` function on its own if shorthands or npm modules are meant to be resolved.
Not to mention potential issues in browser environment if `fs` module is used.

P0lip:
Another potential issue with `spectral.setRuleset(ruleset: IRuleset)`.
A consumer would also need to resolve extended rulesets on its own.
We could resolve them, but relative paths to rulesets would be *very* tricky to support... I'd say almost impossible since we don't know the base path of parent ruleset and would need to speculate a lot and make lots of assumptions that will never work under all circumstances.

marbemac:
> A consumer would also need to resolve extended rulesets on its own.

[...]

PR #431, opened 2019-08-11
feat: implement loadRuleset method

P0lip:
Closes https://github.com/stoplightio/spectral/issues/366

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

Although I marked `addRules` as deprecated, I'd keep it as well as `addFunctions` method as long as we do not implement `setRuleset` method.
If we remove them, and don't provide any reasonable alternatives, the API will fairly limited.
There will be no clean way to load inlined (not included in any file etc.) rules or functions anymore.

Note, besides the new method, I also fixed a minor issue with built-in rulesets loading that could fail under certain condition, more here https://github.com/stoplightio/spectral/commit/d03aeeac653f668b1b8464b6724b233f5fc605c5


docs/guides/javascript.md
```diff
@@ -55,14 +55,19 @@ Spectral comes with some rules and functions that are very specific to OpenAPI v
 ```js
 const { Spectral } = require('@stoplight/spectral');
 const { parseWithPointers } = require("@stoplight/yaml");
-const { oas3Functions, rules: oas3Rules } = require('@stoplight/spectral/dist/rulesets/oas3');
+const { oas3Functions } = require('@stoplight/spectral/dist/rulesets/oas3');
```
philsturgeon: We're going ahead with this feature first, then getting rid of functions in #430 afterwards? I thought I saw you mention doing it the other way around but it does not matter to me. Just making sure that is the plan.

src/__tests__/spectral.jest.test.ts
```diff
@@ -0,0 +1,123 @@
...
+              function: 'pattern',
+              functionOptions: {
+                match: 'Stoplight',
+              },
+            },
+          },
+        },
+      };
+
+      (fetch as FetchMockSandbox).mock('https://localhost:4000/custom-ruleset', {
```
philsturgeon: We already use [nock](https://github.com/nock/nock), can we use this instead of fetch-sandbox? 

```cat package.json | grep nock
    "nock": "~10.0.6",```

Or can we switch that nock usage to fetch sandbox?

Either way these seem to do the same sort of thing and we should not have both.

File: docs/guides/javascript.md [hunk 3]
```diff
@@ -55,14 +55,19 @@ Spectral comes with some rules and functions that are very specific to OpenAPI v
...
 spectral.addFunctions(oas3Functions());
-oas3Rules().then(rules => spectral.addRules(rules));
+spectral.loadRuleset('spectral:oas3')
+  .then(() => spectral.run(myOpenApiDocument))
+  .then(results => {
+    console.log('here are the results', results);
+  });
 ``` 
 
 _Spectral JS API does not yet support loading rulesets directly, but will be added in an upcoming version to avoid needing to load rules and functions separately._
```
philsturgeon:
```suggestion
```

philsturgeon: Can you check the rest of this file to make sure nothing else is inaccurate.

docs/guides/javascript.md
P0lip: Yeah, that's the plan.

src/__tests__/spectral.jest.test.ts
P0lip: Yeah, I was aware of `nock` in the project.
The main obstacle with usage of `nock` is that it's built on top of Node.js and hence we cannot mock requests in a browser using it.
That said, I can convert that Jest test to use `nock`, but will need to leave that mocking solution for Karma-based (browser) tests.
Does it sound good?

Side note,
I planned to share those tests initially, hence I used fetch-mock in both, but then realized I cannot do that in the way I want to and decide to keep them separate (jest and karma).

docs/guides/javascript.md
P0lip: We still have `addRules` in a document covering custom rules
I'm not really sure how we could replace that example, as there is no way to do that without having another file with a ruleset.
The only way to have an equivalent solution would be to create a custom ruleset with a custom rule and load that ruleset.

philsturgeon: That''s OK! This PR is about adding the loadRuleset ability to catch JS up on the CLI, and we are swapping the common examples to show loadRuleset because it is preferred. We cannot totally remove addRules until custom functions are done, so this is all fine. 

When we do #430 we should make a new section in rulesets, and keep the addRules docs in here, but mark it as deprecated in the code _and_ in the docs. Then in v5.0 we remove it from both.

Issue #369, opened 2019-07-15
Register rules against specific versions of API description formats


philsturgeon:
**User story.**
I'd like to write one ruleset for a specific API description format (e.g.: OpenAPI), knowing that some rules will only effect v2.0 or v3.0 or v3.1 specifically. 

**Is your feature request related to a problem?**

Before CLI users could chose a custom ruleset, Spectral would look at the contents of the file, and automatically apply oas2 or oas3, both of which extend the oas ruleset but add some version-specific rules. 

When we added --ruleset in v3, this autodetect behavior was kept, but only as a default. v3 and v4 have no knowledge of what version is being run, so whatever ruleset you apply will be used without any further thought. If you apply a ruleset with OpenAPI v3 rules to a OpenAPI v2 document, it will silently fail in some scenarios, or loudly fail in other scenarios. 

Folks can create their own version specific rulesets, then use their own detection code in order to decide which ruleset to run, but this should not be something our users need to do. 

**Describe the solution you'd like**

```
extends: spectral:oas
rules:
  my-rule-name:
    description: Tags must have a description.
    formats:
      - oas2
      - oas3
    given: $.tags[*]
    then:
      field: description
      function: truthy
```

Or something. Not tied to that, and open to recommendations. 

For backwards compatibility, we just empty the `spectral:oas2` and `spectral:oas3` rulesets, move everything to the main `spectral:oas` ruleset, and tie them to the appropriate formats.

philsturgeon:
@marbemac @P0lip what do you folks think of this one?

P0lip:
@philsturgeon 
I agree that having it automated would be a very nice addition.
Instead of merging all rulesets, we could provide a new property to a ruleset, such as `spec` or `appliesTo` or something similar.
Then, Spectral would recognize that property and act accordingly to the supplied spec.
It's not much of work and would play nicely with `loadRuleset` change we are planning to introduce.
Thoughts?

philsturgeon:
Yeah whatever we do here should be implemented in loadRuleset() so whichever order that happens is fine. It would be an optional keyword which would limit which api description formats it applies to, meaning that you would not need to add it on every single rule. I think that is what you were asking about? I agree it would be a bit rubbish typing it out every time.

P0lip:
> It would be an optional keyword which would limit which api description formats it applies to, meaning that you would not need to add it on every single rule. I think that is what you were asking about? I

Yeah, this is exactly what I meant.

PR #432, opened 2019-08-11
feat: register rules against specific formats

P0lip:
Closes https://github.com/stoplightio/spectral/issues/369

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

Heads up - I haven't written any documentation yet, but if we all agree on the proposed API, I'm more than happy to give it a stab.

A few words regarding the implementation I decided to go with.
The PR features support for both of the ideas described in the linked GH issue:
- rule-scoped formats as suggested by @philsturgeon 
- ruleset-scoped formats as suggested by @P0lip

There hasn't been anything said in terms of inheritance, so I decide to keep it simple and make formats ruleset-specific, with no inheritance at all.

Moreover, I hesitated from registering any default formats such as OAS2 or OAS3 globally and made them specific to CLI version only.
I exposed 2 simple functions that can be used for the spec detection, so if one wants to register oas2/oas3 formats, it should be pretty straightforward.

### DIFF
--- a/src/cli/commands/lint.ts
+++ b/src/cli/commands/lint.ts
@@ -11,6 +11,7 @@ import { json, stylish } from '../../formatters';
 import { readParsable } from '../../fs/reader';
 import { httpAndFileResolver } from '../../resolvers/http-and-file';
 import { getDefaultRulesetFile } from '../../rulesets/loader';
+import { isOAS2Spec, isOAS3Spec } from '../../rulesets/lookups';
 import { oas2Functions, rules as oas2Rules } from '../../rulesets/oas2';
 import { oas3Functions, rules as oas3Rules } from '../../rulesets/oas3';
 import { readRulesFromRulesets } from '../../rulesets/reader';
@@ -150,6 +151,24 @@ async function lint(name: string, flags: ILintConfig, command: Lint, rules?: Rul
     mergeKeys: true,
   });
   const spectral = new Spectral({ resolver: httpAndFileResolver });
+  spectral.registerFormat('oas2', document => {
+    if (isOAS2Spec(document)) {
+      command.log('OpenAPI 2.0 (Swagger) detected');
+      return true;
+    }
+
+    return false;
+  });
+
+  spectral.registerFormat('oas3', document => {
+    if (isOAS3Spec(document)) {
+      command.log('OpenAPI 3.x detected');
+      return true;
+    }
+
+    return false;
+  });
+
   if (parseInt(spec.data.swagger) === 2) {
     command.log('Adding OpenAPI 2.0 (Swagger) functions');
     spectral.addFunctions(oas2Functions());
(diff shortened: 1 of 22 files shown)


src/cli/commands/lint.ts
```diff
@@ -11,6 +11,7 @@ import { json, stylish } from '../../formatters';
 import { readParsable } from '../../fs/reader';
 import { httpAndFileResolver } from '../../resolvers/http-and-file';
 import { getDefaultRulesetFile } from '../../rulesets/loader';
+import { isOAS2Spec, isOAS3Spec } from '../../rulesets/lookups';
```
philsturgeon: Some terminology style guide stuff here, instead of saying "Spec" (JSON is a spec, YAML is a spec, HTTP is a spec, we dont say spec after all of them), and the S in OAS is already "spec" anyway.

API Descriptions are documents written in special formats, so there is a description Document - the file(s) - and description Formats (OpenAPI, RAML, etc). 

Can we try and make sure that at least our new code is consistent with this style guide, and maybe boy scout anything internal we bump into on the way?

tldr: `isOpenApiV2` & `isOpenApiV3`

philsturgeon: oh we should probably make them isOpenApiv2_0, isOpenApiv3_0, isOpenApiv3_1, etc or something less ugly. People will want to have special rules for 3.0 like "hey upgrade to 3.1" lol

P0lip: Good catch. Didn't pay much attention to that, but you are absolutely right about it.
Will do.

philsturgeon:
Btw the end goal here is to let people extend `spectral: oas` and it’ll all just magically work. I think we should merge all the rules into a single ruleset, and leave oas2 and oas3 there, empty, both just extending that main one. It’ll all work the same because yay extension.

src/cli/commands/lint.ts
```diff
@@ -150,6 +151,24 @@ async function lint(name: string, flags: ILintConfig, command: Lint, rules?: Rul
     mergeKeys: true,
   });
   const spectral = new Spectral({ resolver: httpAndFileResolver });
+  spectral.registerFormat('oas-2', document => {
```
philsturgeon:
```suggestion
  spectral.registerFormat('oas2', document => {
```

src/cli/commands/lint.ts
```diff
@@ -150,6 +151,24 @@ async function lint(name: string, flags: ILintConfig, command: Lint, rules?: Rul
...
+    if (isOpenApiv2_0(document)) {
+      command.log('OpenAPI 2.0 (Swagger) detected');
+      return true;
+    }
+
+    return false;
+  });
+
+  spectral.registerFormat('oas-3', document => {
```
philsturgeon:
```suggestion
  spectral.registerFormat('oas-3.0, document => {
```

src/cli/commands/lint.ts
```diff
@@ -150,6 +151,24 @@ async function lint(name: string, flags: ILintConfig, command: Lint, rules?: Rul
...
+    if (isOpenApiv3(document)) {
```
philsturgeon: It’d more important for v3 to have the _0 because it’s gonna get v3.1 at some point. Should we leave them as OpenApiv2 and OpenApiV3 then just add _1 when it comes out? If so we can probably ignore the above comments on format names.

P0lip: Having `oas2`, `oas3` and potentially `oas3.1` seems like a reasonable choice.
Are we sticking to this?

philsturgeon:
Let’s do that :)

--
Phil Sturgeon
@philsturgeon

P0lip:
@philsturgeon 
Thank you for taking care of docs!

PR #445, opened 2019-08-13
docs: ruleset formats

philsturgeon:
Added docs for #432


docs/getting-started/rulesets.md
```diff
@@ -40,6 +40,60 @@ While running it with this object, it will succeed:
...
 ```
 
+### Formats
+
+Formats are an optional way to specify which API description formats a rule, or ruleset, is applicable to. Currently Spectral supports these two formats:
+
+- `oas2` (this is OpenAPI v2.0 - formerly known as Swagger)
+- `oas3` (this is OpenAPI v3.0)
+
+Specifying the format is optional, so you can completely ignore this if all the rules you are writing apply to any document you lint, or if you have specific rulesets for different formats. If you'd like to use one ruleset for multiple formats, the format key is here to help.
```
P0lip:
```suggestion
Specifying the format is optional, so you can completely ignore this if all the rules you are writing apply to any document you lint, or if you have specific rulesets for different formats. If you'd like to use one ruleset for multiple formats, the formats key is here to help.
```

P0lip:
```suggestion
Specifying the format is optional, so you can completely ignore this if all the rules you are writing apply to any document you lint, or if you have specific rulesets for different formats. If you'd like to use one ruleset for multiple formats, the formats key is here to help.
```

docs/getting-started/rulesets.md
```diff
@@ -40,6 +40,60 @@ While running it with this object, it will succeed:
...
+          minItems: 1
+          type: array
+```
+
+Seeing as the `servers` array only appeared in OpenAPI v3.0, we don't want this rule appearing when Spectral lints OpenAPI v2.0 documents.
+
+Alternatively, formats can be specified at the ruleset level:
+
+```
+```yaml
```
P0lip: Shall we say a word about the fact that the local rule formats definition has a higher priority?
```yaml
formats: ["oas2", "oas3"]
rules:
  api-servers: # this rule would be applied only to oas2
    description: "OpenAPI `servers` must be present and non-empty array."
    recommended: true
    given: "$"
    formats: ["oas2"] 
    then:
      field: servers
      function: schema
      functionOptions: 
        schema:
          items: 
            type: object
          minItems: 1
          type: array
```

docs/guides/javascript.md
```diff
@@ -313,3 +313,33 @@ spectral.run({description: 'Swagger is pretty cool!',}).then(results => {
...
+
+Spectral supports two core formats: `oas2` and `oas3`. Using `registerFormat` you can add support for autodetecting other formats. You might want to do this for a ruleset which is run against multiple major versions of description format like RAML v0.8 and v1.0.
+
+```js
+spectral.registerFormat('foo-bar', obj => typeof obj === 'object' && obj !== null && 'foo-bar' in obj);
+
+spectral.addRules({
+  rule1: {
+    given: '$.x',
+    formats: ['foo-bar'],
```
P0lip: Ha, we didn't discuss such a case in the original issue or I might have misunderstood something, but... if no format can be found, we run all rules regardless of their formats.
In other words, if you change it to
```suggestion
    formats: ['foo-dsadaczxcz'],
```

and 

```
const result = await spectral.run({
  'foo-baraaaa': true,
```

the rule would be executed.

Now that I think of it, I don't believe it was wise.
Changing it would be very easy though, so can do it right away.

P0lip: https://github.com/stoplightio/spectral/pull/449

philsturgeon: This example was not actually talking about what you are talking about currently, I just copied and pasted it from a test. But, im glad it got you thinking, because #449 is a good change. :D

docs/getting-started/rulesets.md
philsturgeon: For now I think I'd like to avoid drowning them in options but if the question comes up we can add it and make a new Formats page which goes more in depth with all the things.

P0lip: Alrighty, sounds good to me.

Issue #380, opened 2019-07-15
Do not require JSON Schema to be inlined in rulesets


philsturgeon:
**User story.**
As a ruleset maintainer, I would like to create rulesets which do not have JSON Schema inline, so that I can have more rules in a more concise file. 

**Is your feature request related to a problem?**
Currently the oas2 and oas3 rulesets have the OpenAPI metaschema files, and this is bloating a 100 line ruleset to almost 2k lines.

https://github.com/stoplightio/spectral/blob/develop/src/rulesets/oas2/index.json#L106

**Describe the solution you'd like**

Let’s add support for $ref with json-ref-resolver. 

```
    "oas2-schema": {
      "description": "Validate structure of OpenAPIv2 specification.",
      "message": "{{error}}",
      "type": "validation",
      "severity": 0,
      "recommended": true,
      "given": "$",
      "then": {
        "function": "schema",
        "functionOptions": {
          "schema": {
            "$ref": "../oas2-metaschema.json"
          }
        }
      }
```

marbemac:
How about we take advantage of $refs, since we already have a very capable json-ref-resolver loaded up in Spectral? Could just dereference the ruleset file after loading it. This would be like an hour of work. Something like this:

```
"oas2-schema": {
      "description": "Validate structure of OpenAPIv2 specification.",
      "message": "{{error}}",
      "type": "validation",
      "severity": 0,
      "recommended": true,
      "given": "$",
      "then": {
        "function": "schema",
        "functionOptions": {
          "schema": {
            "$ref": "../oas2-metaschema.json"
          }
        }
      }
```

In fact, if we supported $refs in the ruleset file, they could be used in all sorts of ways to de-dupe things.

philsturgeon:
@marbemac I think this would 100% lock us in to rulesets always being defined in JSON/YAML, and definitely not being TypeScript, right? Is that something we want to do?

marbemac:
Probably missing something - could you share an example?

---

Defining in json versus ts shouldn't really matter, at least for the use cases I've considered - but again I might be missing something.

**.spectral.json**

```json
{
  "rules": {
    "oas2-schema": {
      "description": "Validate structure of OpenAPIv2 specification.",
      "message": "{{error}}",
      "type": "validation",
      "severity": 0,
      "recommended": true,
      "given": "$",
      "then": {
        "function": "schema",
        "functionOptions": {
          "schema": {
            "$ref": "../oas2-metaschema.json"
          }
        }
      }
[...]

PR #454, opened 2019-08-14
feat: support refs in rulesets

P0lip:
Closes https://github.com/stoplightio/spectral/issues/380

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

NOTE: I'm not planning to add any docs - I don't think we should advertise this feature.


P0lip:
@marbemac @philsturgeon 
I got it working yet the tests fail when executed in parallel.
This is caused by some json-ref-resolver quirky behavior.
Will verify why this happens.

That said, I'd rather revert that change and go with a simpler concept. It does increase complexity in terms of Studio and I don't' really see whether $refs in rulesets would be that useful given `extends` we have.

marbemac:
What’s the alternative that leverages extends to accomplish the goal?

Re this implementation, rather than making load ruleset async, what about leveraging the async run (I think it’s called run? Not sure, on phone, but something like this)  method we already have? Basically at the start of run, resolve the merged ruleset if it has not been previously resolved. Could hash the merged ruleset and use that as a cache key so that multiple invocations of run only need to resolve once, unless the underlying ruleset object has changed.

philsturgeon:
Just to remind everyone, the original use case was "we want to get the inline JSON Schema definitions out of the ruleset files and into their own locations" so if this larger more generic feature is causing trouble, maybe we can take a step back, and look at just solving the original use case? Sometimes I feel like making things more general is more trouble than its worth.

P0lip:
> Just to remind everyone, the original use case was "we want to get the inline JSON Schema definitions out of the ruleset files and into their own files" so if this larger more generic feature is causing trouble, maybe we can take a step back, and look at just solving the original use case? Sometimes I feel like making things more general is more trouble than its worth.

Yeah, exactly. I mean, we are bringing something very robust that is by far more flexible than we need it to be. It's certainly something we'll need to maintain even if we keep it undocumented, as others might just discover that and use it in their own rulesets.
Either way, I'm pretty much there - it works like a charm when tests are executed in full isolation (one by one), when they are run in parallel they fail for some reason.

P0lip:
> What’s the alternative that leverages extends to accomplish the goal?

`extends: [oas2-schema]`?
I know it might look awkward, but it'd its job and end developers wouldn't even notice since they extend `oas2` or `oas3` rulesets.

>  Not sure, on phone, but something like this) method we already have? Basically at the start of run, resolve the merged ruleset if it has not been previously resolved. 

This would work for `schema`, but what if one places `$ref` in a rule? In order to merge everything safely, we'd need to operate on resolved data.

P0lip:
@philsturgeon @marbemac 
Okay, I think I nailed the issue.
It seems like `expect` is having some serious troubles with our assertions.
Not sure if it's diffing or just the result printing, but basically if I adjust the assertion a bit, it all works fine.

P0lip:
https://github.com/stoplightio/spectral/pull/454/commits/c4d4365e8a8603aa36eb58393df0860784ae128e is unrelated to the PR, but it's a fix for an issue exposed by the PR - we didn't handle schema ids properly.

src/rulesets/oas2/__tests__/oas2-schema.ts
```diff
@@ -1,12 +1,20 @@
...
   const s = new Spectral();
+
   s.addRules({
-    'oas2-schema': Object.assign(ruleset.rules['oas2-schema'], {
+    'oas2-schema': Object.assign({}, ruleset.rules['oas2-schema'], {
       recommended: true,
       type: RuleType[ruleset.rules['oas2-schema'].type],
+      then: {
+        ...ruleset.rules['oas2-schema'].then,
+        functionOptions: {
```
philsturgeon: Is this meant to be here? Looks like we’re oassing in the contents not a ref. Is that just to make the unit test work and avoid it being an integration test? If so, should we do the same for oas3?

P0lip: Yeah, it's to avoid it being an integration test.
We don't have any tests for oas3-schema

philsturgeon:
YAAAAY

Issue #91, opened 2019-02-17
Spectral is generating paths that do not exist in the original document


XVincentX:
It seems like the included rules for OAS2/3 can generate paths that do not exist in the original document.

An easy way to reproduce this is to simply install the default rules and create an OAS document with a path that's missing the `description` property.

marbemac:
This is intended, otherwise how would it report missing properties?

Or, put another way, what is the expected result that still gives the spectral user enough info to pinpoint the issue?

XVincentX:
Hmm I guess it depends how we see Spectral.

If it's an "end product" (which is my view right now) then it should figure out something I can use without having to further elaborate the returned value. In this specific case, it should give me the closes parent (as you mentioned somewhere else).

If it's not an end product then maybe it could be fine as long we clearly document the result format on the README; albeit it looks suspicious to me a sentence that would go like this:

* `path:` The JSON path to the offending line _or a non existent path_  in the document indicating where the required field would have been placed.

Thoughts?

marbemac:
Hmm - what should the result object look like for this description error?

PR #459, opened 2019-08-18
fix: do not generate non-existing paths

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/91

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


src/rulesets/oas2/__tests__/valid-example.ts
```diff
@@ -156,7 +156,7 @@ describe('valid-example', () => {
       expect.objectContaining({
         code: 'valid-example',
         message: '"self" property type should be array',
-        path: ['definitions', 'halRoot', '_links', 'self'],
+        path: ['definitions', 'halRoot', 'example', '_links', 'self'],
```
P0lip: This is quite tricky. I still need to fix it.

P0lip: This is a bit of a different issue, fixed in https://github.com/stoplightio/spectral/pull/459/commits/0e3844f2025cd5fb07d41f217e1c33c5e3d85ef0
yet I believe the new error message we generate is actually worse than the former one, or am I wrong?

src/rulesets/oas3/__tests__/valid-example.ts
```diff
@@ -169,7 +169,7 @@ describe('valid-example', () => {
       expect(results).toEqual([
         expect.objectContaining({
           code: 'valid-example',
-          message: `"ip_address" property format should match format "${format}"`,
+          message: `"example" property format should match format "${format}"`, // hm, ip_address is likely to be more meaningful no?
```
P0lip: @philsturgeon thoughts?

philsturgeon: Depends? This feels like a limitation of trying to shove everything into a simple string message. We are talking about the example, but it sure could be useful to let them know its the ip_address example that is wrong.

P0lip: Yeah, exactly.
I'd expect to receive something like `"ip_address" example property format should match format "${format}"`. Will do it this way.

PR #460, opened 2019-08-19
feat: transferable rulesets

P0lip:
Closes https://github.com/stoplightio/spectral/issues/494

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

Everything can be found in https://github.com/stoplightio/spectral/issues/494

### DIFF
--- a/src/cli/commands/lint.ts
+++ b/src/cli/commands/lint.ts
@@ -121,7 +121,7 @@ async function lint(name: string, flags: ILintConfig, command: Lint, rulesets: O
   const spectral = new Spectral({ resolver: httpAndFileResolver });
 
   try {
-    await spectral.loadRuleset(rulesets ? rulesets : ['spectral:oas']);
+    await spectral.loadRuleset(rulesets || ['spectral:oas']);
   } catch (ex) {
     command.log(ex.message);
     process.exitCode = 2;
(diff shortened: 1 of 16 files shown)


src/cli/commands/lint.ts
```diff
@@ -169,39 +160,14 @@ async function lint(name: string, flags: ILintConfig, command: Lint, rules?: Rul
...
+  if (rulesets && flags.verbose) {
+    command.log(`Found ${Object.keys(rulesets).length} rulesets`);
   }
 
   if (flags.skipRule) {
-    rules = skipRules({ ...rules }, flags, command);
-  }
-  if (!rules) {
-    throw new Error('No rules provided, and document type does not have any default rules, so lint has nothing to do');
+    skipRules(spectral.rules, flags, command);
```
philsturgeon: Is this function mutating the public property? Is that the way we want to do this?

P0lip: You are correct, it mutates the property.
Should be harmless in this case, but actually I think it'd be nice if we provided a `removeRule` method or something to Spectral.

BTW, review https://github.com/stoplightio/spectral/pull/430 first - #460 PR is based on it.

P0lip:
I know you @marbemac had some concerns about that bundling approach, so don't hesitate to share them in this PR provided that you still have them :D

PR #461, opened 2019-08-19
 feat: handle circular rulesets

P0lip:
Closes https://github.com/stoplightio/spectral/issues/479

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

See ticket for more info :) It has implementation details included.


philsturgeon:
If circular rulesets is getting a bit tough, we can change things around a bit in a different way.

new ruleset: `spectral:openapi`

This has all of the rules in it, every single one, because now we have formats we don't need to split by version.

Then we just mark oas, oas2 and oas3 for deprecation in v4.1, and remove them in spectral v5.0. What do you think?

P0lip:
> If circular rulesets is getting a bit tough, we can change things around a bit in a different way.

Actually, it's all good, no worries!
I'll just need to update a bunch of tests, but I got it working.
Just waiting for custom functions to be merged...

P0lip:
@philsturgeon @marbemac 
I'd like to get the oas ruleset extends oas2 and oas3 rulesets thing done as a part of this PR.
There is on issue, though. A few rule definitions collide, i.e. `valid-example`.
Shall I just move such rules to the base oas ruleset?

philsturgeon:
Yeah anything with the same name can just have oas2 or oas3 prefixed. 

--
Phil Sturgeon
@philsturgeon

P0lip:
@philsturgeon @marbemac 
Added support for oas ruleset extends oas2 and oas3 rulesets in https://github.com/stoplightio/spectral/pull/461/commits/d47b53e7ce60abd728e80720361825a7b5f60b54

I know you will ask "Why so many changes? Shouldn't it work out of the box?", so in order to give you an answer in advance, let me explain it.

There was a subtle yet fairly significant issue/quirk with the current implementation of rulesets.
Some may say it's a feature, some may say it's a bug, but IMHO it's rather an issue that needs to be fixed (and it's fixed).
So the case is that a base ruleset treats its own rule as they were all recommended.
This might be observed quite well here
https://github.com/stoplightio/spectral/pull/461/commits/d47b53e7ce60abd728e80720361825a7b5f60b54#diff-1d22b24e26ececefd0c87be1d0569b68L543
`model-description` rule is *not* recommended, therefore we should not be running that rule when you load oas2 ruleset in the following way `spectral.loadRuleset('spectral:oas2')`.
It did work fine if you extended spectral:oas2 ruleset, i.e.
```
extends: [spectral:oas2]
rules:
```
There are also other tests showing this invalid behavior, for instance https://github.com/stoplightio/spectral/pull/461/commits/d47b53e7ce60abd728e80720361825a7b5f60b54#diff-62b31a98be4036e65207ec7971253e93R19
In general, all updated tests cover that issue.
It's not straightforward to spot. I wouldn't notice it either if we didn't bring support for `oas ruleset extends oas2 and oas3 rulesets` feature.

The diff is also large due to prettifying (lines shuffled etc). Moreover, I made assertions a bit cleaner in a few places.

I hope it's all good now, but I'd really encourage you to build the code and try it in a couple of rulesets. Also, pay attention to tests, validate whether the expected results are okay etc.
We had a few invalid assertions, well at least they 

If we get all of that right, we should be in a very good spot - we have lots of various tests now, so in general, I feel fairly confident about that particular feature now.

Anyway, as we all know, tests will detect issues, but will never detect the lack of issues :D

LMK if you have any other questions.

src/rulesets/oas/__tests__/valid-example.ts
```diff
@@ -2,6 +2,10 @@ import { DiagnosticSeverity } from '@stoplight/types';
 import { RuleType, Spectral } from '../../../spectral';
 import * as ruleset from '../index.json';
 
+// @oclif/test packages requires @types/mocha, therefore we have 2 packages coming up with similar typings
+// TS is confused and prefers the mocha ones, so we need to instrument it to pick up the Jest ones
+declare var test: jest.It;
```
philsturgeon: If we get #325 merged first then all of these can go away.

PR #470, opened 2019-08-20
Auto-Approve Dependabot

philsturgeon:
Automatically approve PRs from Dependabot so we don't have to.

This is a bit less pretty than it could be, due to an error with [auto-approve](https://github.com/hmarr/auto-approve-action/issues/8). I'm following the issue and will clean this up later. It's fine.


.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,17 @@
...
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: kjvalencik/github-actions/approve@master
+        if: github.event_name == 'pull_request' && (github.actor == 'dependabot-preview[bot]' || github.actor == "dependabot[bot]')
```
P0lip:
```suggestion
        if: github.event_name == 'pull_request' && (github.actor == 'dependabot-preview[bot]' || github.actor == 'dependabot[bot]')
```
Syntax error :sweat_smile: 
image

philsturgeon: Just pushed the fix!

install.sh
```diff
@@ -1,35 +0,0 @@
-#!/bin/sh
```
P0lip: Did you forget to commit the file? I  see it's removed, but not added anywhere

philsturgeon: good catch! I was rushing.

hmarr:
FYI, I rebuilt the [auto-approve action](https://github.com/hmarr/auto-approve-action) for the new version of GitHub Actions (released as v2.0.0), so that may fix the issues you were having. Sorry it was broken for a little while!

Issue #368, opened 2019-07-12
Make CLI not to return non-zero exist code when only warnings occur


aleung:
**User story.**

As a continuous integration user, I can provide a parameter to CLI, so that the CI pipeline won't fail when there is no error but only warnings.

As a continuous integration user, I can provide a parameter to the CLI, so that only results of a certain severity will be returned as output.

**Describe the solution you'd like**

_Updated 2019-07-16 by @philsturgeon_

Two new CLI options:

1. `--fail-severity=warn` - enum[`hint`, `info`, `warn`, `error`] - change the level of results which come out of Spectral.
2. `--display-only-fail-severity-results` - boolean - spectral will error if results of any level at all are reported. We should make that configurable.

We should probably not change the default behavior in this as its aiming for v4.1. Place a TODO note in the code to change the default to `warn` in v5.0.

philsturgeon:
I’m open to something along these lines, but not exactly what ESLint does.  I think number of warnings is arbitrary, if you’re asking for feedback on your specs and you get some then you should fix it, and if you don’t like what feedback you’re getting use skip, make a custom ruleset, or extend the default ruleset to change the severity of a specific rule. 

With that said, can you think of a switch? —error-only or something?

aleung:
@philsturgeon `--error-only` is fine for me. I just thought `--max-warnings` would give more flexibility to some other users.

soulcutter:
`--ignore-warnings` maybe?

I mildly favor a boolean-style flag vs tracking how many warnings trigger an error. It's lower-complexity, and also I am skeptical of the use case where 3 warnings is ok, but 4 is an error.

philsturgeon:
I prefer a Boolean flag too but —ignore-warnings could be a tad confusing. We’re not ignoring them, just not failing the run if we see them. Ignoring might be closer to #381. 

Should these two features perhaps be the same feature? 

We could have —severity=error and it will only show errors, and only fail if errors come back. 

—severity=warm would show warnings and errors, and fail if either came back. 

—severity=info would show info, warnings, errors, and fail if... info came back? Or warnings/errors? 

Not sure on the last one.

philsturgeon:
There could be two switches, —fail-severity=warn —severity=info

First is where it fails, second is what it shows. Both have same defaults they have now.

XVincentX:
[...]

Issue #304, opened 2019-07-06
Port Prism test harness


brianmrock:
**User story.**
Port the existing Prism test harness once it has settled. At time of writing it is still being reworked and refactored, but once done, Spectral should get the same treatment.

**AC*
-Spectral uses the same test-harness library Prism does
- Run test-harness for every PR in CI
- Write some e2e smoke tests (up to 5 should be enough)

**Scope of Work**
- [ ] Extract the Prism harness utilities in a separate package
- [ ] Reinstall the utilities in Prism and Spectral

PR #471, opened 2019-08-20
chore: switching from oclif to yargs

### LINKED ISSUE #325: Switch from oclif to yargs

philsturgeon:
Fixes #325

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


XVincentX:
@philsturgeon Maybe put me as a reviewer when you're done — I made a BIG mistake while switching to Yargs that I do not remember, but I want to make sure we do not make it again with Spectral

P0lip:
@philsturgeon @XVincentX 
I'm afraid we won't be able to implement the yargs testing approach that established in Prism.
There are a few significant differences between the case we deal with here and Spectral.
First of all,
- we don't need middleware, which seems to be needed to make `commandPromise` an actual promise,
- there is no way to grab output - we could mock `writeOutput`, yet this has a huge downside - it's not equal to actual output printed in the console, so we won't be able to test `--format=json` flag
- it does not wait until handler promise gets resolved, so there is no way to make an assertion when spectral has in fact finished linting

Prism does not face any of above.

That said, I suppose we'd need to resort to process spawning - I used this approach on a number of occasions and it's by far the most reliable.

Basically, we need something more or less like this
```js
import { promisify  } from 'util'

const exec = promisify(require('child_process').exec);

it('writes something', async () => {
  await { stdout, stderr} = exec('./bin/run lint foo.yaml');
  expect(stdout).toContain('blabala');
});
```

philsturgeon:
Before trying the commandPromise.then approach I was using code from the yargs advanced documentation showing how you capture output... so why not? We are providing the help() format is output, so how is that output differently to what our commands are doing? Should we log somewhere else other than console?

If that approach cannot be saved any way, we can move writeOutput to the service, mock it, and just unit test what is being passed in. We then have unit tests on the service showing those inputs give the expected functionality? The integration will be skipped and end to end testing will cover it properly when we port a test harness (#304).

P0lip:
> If that approach cannot be saved any way, we can move writeOutput to the service, mock it, and just unit test what is being passed 

I tried that, but it gets called too late. commandPromise promise does not wait until handler is resolved.
We'd need to the whole handler to the linter service and mock it.

As said, that approach won't let us catch `console.warn` calls and such, but if we mean to have unit tests for now, we should be okay.

philsturgeon:
I still don't quite understand. I don't care about commandPromise, that was just me trying something out.

https://github.com/yargs/yargs/blob/master/docs/advanced.md

```
it("returns help output", async () => {
  // Initialize parser using the command module
  const parser = yargs.command(require('./my-command-module')).help();

  // Run the command module with --help as argument
  const output = await new Promise((resolve) => {
    parser.parse("--help", (err, argv, output) => {
      resolve(output);
    })
  });

  // Verify the output is correct
  expect(output).toBe(expect.stringContaining("helpful message"));
});
```

1. Why can we test help like this (this works) but we cannot test anything other than help? 
2. If we know the difference, can we can make our code log more like help?

P0lip:
@philsturgeon 
Uhhh, now I get what you mean.
Let me see

philsturgeon:
@P0lip ok, thats a bit of a shame but also your code is looking amazing in this branch now. Do you think you a can take this over the finishing line with some unit tests on the services, and any sort of test which shows the CLI is doing something productive, and I'll get to work on #368 based off of this branch?

P0lip:
@philsturgeon 
The CI might be green at this point, yet I'd like to enable a few more tests in `src/cli/services/__tests__/linter.test.ts`

P0lip:
Although, we all want to get rid of oclif so badly I'd suggest holding on with this one.
I don't feel confident enough to to tell it does not introduce any breaking changes.
We don't verify stdout too much now, therefore in particular json format might (or might not) be broken.
If someone can test a few cases manually to verify the actual output printed to stdout, this should be safe to go. 

Other than that, should be good to review! @philsturgeon, @marbemac, @XVincentX it's all yours now.

philsturgeon:
Ok, seems like we need the test harness before this one can be done. I'll get started on #304 now.

P0lip:
`--format=json` is broken since we log `OpenAPI 3.x detected` now.
I'll take care of it.

philsturgeon:
@XVincentX ready! :D

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
-        .stdout()
-        .command(['lint', validOas3SpecPath, '-r', validRulesetPath])
-        .it('outputs warnings in default format', ctx => {
-          expect(ctx.stdout).toContain('5:10  warning  info-matches-stoplight  Info must contain Stoplight');
-          expect(ctx.stdout).not.toContain('Info object should contain `contact` object');
-          expect(ctx.stdout).toContain('OpenAPI 3.x detected');
-        });
+  it('shows help when invalid arguments are passed', async () => {
+    const output = await run('lint --foo');
+    expect(output).toContain('lint a JSON/YAML document from a file or URL');
```
lag-of-death: hm.. assuming this indeed is help for invalid arguments. If something relating to `foo` is printed, can we have the expectation with `foo` in it? If not, then it OK I guess :)

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
-        .it('given valid remote ruleset file, outputs no issues', ctx => {
-          expect(ctx.stdout).toContain('No errors or warnings found!');
-        });
-    });
+  it('calls lint with document and default options', async () => {
+    const doc = './__fixtures__/empty-oas2-document.json';
+    await run(`lint ${doc}`);
+    expect(lint).toBeCalledWith(
+      doc,
+      expect.objectContaining({
```
lag-of-death: `calls lint with document and default options`, so there's `doc`, so I know this is the document. But not sure what default options are? If default options is the object containing `.encoding` and `.format`, then what is `undefined`? @philsturgeon, do you think that we can use vars (like you did for `doc`) to name these things?

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
-            '3:6  warning  info-description  OpenAPI object info `description` must be present and non-empty string',
-          );
-          expect(ctx.stdout).toContain('OpenAPI 3.x detected');
-        });
+  it('calls lint with document and custom encoding', async () => {
+    const doc = './__fixtures__/empty-oas2-document.json';
+    await run(`lint --encoding utf16 ${doc}`);
+    expect(lint).toBeCalledWith(
+      doc,
+      expect.objectContaining({
```
lag-of-death: ditto

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
-          expect(ctx.stdout).toContain('22 problems (0 errors, 22 warnings, 0 infos)');
-          expect(ctx.stdout).toContain('OpenAPI 2.0 (Swagger) detected');
-        });
-    });
+  it('calls lint with document and custom encoding and format', async () => {
+    const doc = './__fixtures__/empty-oas2-document.json';
+    await run(`lint -f json --encoding utf16 ${doc}`);
+    expect(lint).toBeCalledWith(
+      doc,
+      expect.objectContaining({
```
lag-of-death: ditto

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
-      .stdout()
-      .command(['lint', invalidOas3SpecPath, '--quiet'])
-      .it('does not log any additional feedback', ctx => {
-        expect(ctx.stdout).not.toContain('OpenAPI 3.x detected');
-      });
+  it('calls lint with document and custom rulesets', async () => {
+    const doc = './__fixtures__/empty-oas2-document.json';
+    const ruleset = 'custom-ruleset.json';
+    await run(`lint -r ${ruleset} ${doc}`);
+    expect(lint).toBeCalledWith(doc, expect.any(Object), ruleset);
```
lag-of-death: ok, so now I understand that `undefined` was just a `not defined ruleset`. But maybe it would read better if we did:
```
const ruleset = undefined;
```
in `calls lint with document and custom encoding`?

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
   });
 
-  describe('when --quiet flag is provided', () => {
-    test
-      .stdout()
-      .command(['lint', invalidOas3SpecPath, '--quiet'])
-      .it('does not log any additional feedback', ctx => {
-        expect(ctx.stdout).not.toContain('OpenAPI 3.x detected');
-      });
+  it('calls lint with document and custom rulesets', async () => {
```
lag-of-death:
```suggestion
  it('calls lint with document and custom ruleset', async () => {
```
later we have `calls lint with document and multiple custom rulesets`

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
-        api.get('/openapi').replyWithFile(200, invalidOas3SpecPath, {
-          'Content-Type': 'application/yaml',
-        }),
-      )
-      .stdout()
-      .command(['lint', 'http://foo.local/openapi'])
-      .it('outputs warnings in default format', ctx => {
-        expect(ctx.stdout).toContain('Info object should contain `contact` object');
-      });
+  it('writes formatted output to a file', async () => {
```
lag-of-death: hm.. is this a duplicate?

src/cli/commands/lint.ts
```diff
@@ -1,261 +1,92 @@
...
+const lintCommand: CommandModule = {
+  describe: 'lint a JSON/YAML document from a file or URL',
+  command: 'lint <document>',
+  builder: yargs =>
+    yargs
+      .positional('document', {
+        description: 'Location of a JSON/YAML document. Can be either a file or a fetchable resource on the web.',
+        type: 'string',
+      })
+      .fail(() => {
```
lag-of-death:
```suggestion
      .fail(showHelp)
```

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+              line: 5,
+            },
+          },
+          source: expect.stringContaining('/__tests__/__fixtures__/gh-474/common.yaml'),
+        }),
+      ]),
+    );
+  });
+
+  it('should handle relative path to a document #2', async () => {
```
lag-of-death: do we need 2 of `should handle relative path to a document`?

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+        expect(output).toEqual(
+          expect.arrayContaining([
+            expect.objectContaining({ code: 'api-servers' }),
+            expect.objectContaining({ code: 'info-contact' }),
+          ]),
+        );
+      });
+
+      describe('and --skip-rule=info-contact is set', () => {
+        it('output other warnings but not info-contact', async () => {
```
lag-of-death:
```suggestion
        it('outputs other warnings but not info-contact', async () => {
```

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+          expect(output).toEqual(expect.not.arrayContaining([expect.objectContaining({ code: 'info-contact' })]));
+          expect(output).toEqual(expect.not.arrayContaining([expect.objectContaining({ code: 'api-servers' })]));
+        });
+      });
+    });
+  });
+
+  describe('--ruleset', () => {
+    describe('extends feature', () => {
+      it('should extend a valid relative ruleset', () => {
```
lag-of-death: hm.. I don't know. Some `it`s are in the form of `does` and others and in the form of `should do`. Maybe it would be good to unify these?

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+        return expect(run(`lint ${validCustomOas3SpecPath} -r ${validNestedRulesetPath}`)).resolves.toEqual([]);
+      });
+
+      it('should fail trying to extend an invalid relative ruleset', () => {
+        return expect(run(`lint ${validCustomOas3SpecPath} -r ${invalidNestedRulesetPath}`)).rejects.toThrowError(
+          ValidationError,
+        );
+      });
+
+      it('given remote nested ruleset should resolve', () => {
```
lag-of-death: `given` in `it`? Not sure here, up to you :)

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+          });
+
+        return expect(run(`lint ${validCustomOas3SpecPath} -r http://foo.local/ruleset-master.yaml`)).resolves.toEqual(
+          [],
+        );
+      });
+    });
+
+    describe('when multiple ruleset options provided', () => {
+      it('given one is valid other is not, outputs "invalid ruleset" error', () => {
```
lag-of-death: ditto

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+        .persist()
+        .get('/openapi')
+        .replyWithFile(200, validOas3SpecPath, {
+          'Content-Type': 'application/yaml',
+        });
+
+      return expect(run('lint http://foo.local/openapi')).resolves.toEqual([]);
+    });
+
+    it('throws if cannot load URI', () => {
```
lag-of-death: condition in `it`, not sure here

src/cli/services/linter.ts
```diff
@@ -0,0 +1,128 @@
...
+export async function loadRulesets(cwd: string, rulesetFiles: string | string[] | null): Promise<IRuleset> {
+  if (rulesetFiles === null || rulesetFiles.length === 0) {
+    return {
+      functions: {},
+      rules: {},
+    };
+  }
+
+  return readRuleset(
+    (Array.isArray(rulesetFiles) ? rulesetFiles : [rulesetFiles]).map(
```
lag-of-death: nice, me likey :+1:

src/cli/services/linter.ts
```diff
@@ -0,0 +1,128 @@
...
+    ),
+  );
+}
+
+export async function lint(name: string, flags: ILintConfig, rulesetFile: Optional<string | string[]>) {
+  if (flags.verbose) {
+    console.info(`Linting ${name}`);
+  }
+
+  let targetUri = name;
```
lag-of-death:
```suggestion
  const targetUri = (!/^https?:\/\//.test(name) && !isAbsolute(name)) ? resolve(process.cwd(), name) : name;
```

src/cli/services/linter.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  if (invalidRules.length !== 0) {
+    if (!flags.quiet) {
+      console.warn(`ignoring invalid ${invalidRules.length > 1 ? 'rules' : 'rule'} "${invalidRules.join(', ')}"`);
+    }
+  }
+
+  if (skippedRules.length !== 0 && flags.verbose) {
+    if (flags.verbose) {
```
lag-of-death: seems like this was already checked for in the line above

src/cli/commands/__tests__/lint.test.ts
philsturgeon: Help shows up whenever there is invalid arguments, so here are two issues:

1. the lack of the required `{document}` argument
2. the --foo option is nonsense

Either one of these two things would cause the help menu to display, but it will not complain about --foo.

philsturgeon: Actually it probably should. This is copied from Prism, and I thought prism was only doing it when the argument was missing, not invalid options too.

- `spectral lint` should show help
- `spectral lint --foo` should complain about `--foo`

src/cli/commands/__tests__/lint.test.ts
philsturgeon: I dunno @P0lip wrote these tests :D

src/types/config.ts
```diff
@@ -7,8 +7,8 @@ export interface ILintConfig {
   encoding: string;
   format: OutputFormat;
   output?: string;
-  ruleset?: string[];
-  skipRule?: string[];
+  ruleset?: string | string[];
+  skipRule?: string | string[];
```
XVincentX: Would it be worth to make this always an array so you do not have to deal with this?

src/cli/commands/__tests__/lint.test.ts
P0lip: These tests have been ported :D. Eight way, the defaults in this particular case means no options specified. Its up to the command definition what those defaults are.

src/cli/services/linter.ts
XVincentX: Interesting stuff here — in theory this could be handled by a logger that will decide whether output something or not based on its configuration; so you can always output stuff and not having to use ifs everywhere. Could be a good next step

src/cli/services/linter.ts
```diff
@@ -0,0 +1,128 @@
...
+import { httpAndFileResolver } from '../../resolvers/http-and-file';
+import { getDefaultRulesetFile } from '../../rulesets/loader';
+import { isOpenApiv2, isOpenApiv3 } from '../../rulesets/lookups';
+import { readRuleset } from '../../rulesets/reader';
+import { Spectral } from '../../spectral';
+import { IParsedResult, RuleCollection } from '../../types';
+import { ILintConfig } from '../../types/config';
+import { IRuleset } from '../../types/ruleset';
+
+export async function loadRulesets(cwd: string, rulesetFiles: string | string[] | null): Promise<IRuleset> {
```
XVincentX: Looking at this code it's clear that if `rulesetFiles` is empty or null — nothing will effectively happen. 

You can make this function type-safe (and avoid the if and returning empty stuff) by declaring that the array must not be empty

```suggestion
export async function loadRulesets(cwd: string, rulesetFiles: string[] & { 0: string }): Promise<IRuleset> {
```

src/cli/commands/__tests__/lint.test.ts
P0lip: In general, all options are passed. I'll explicitly pass them here rather than just a subset of them.

src/cli/commands/__tests__/lint.test.ts
P0lip: In this case, we don't really care about other options - we just want to test whether this one is changed or not.

src/cli/services/linter.ts
P0lip: > y a logger that will decide whether output something or not based on its configuration; 

Yeah, I have a basic implementation of Logger already. Going to ship it in 4.2

src/cli/services/__tests__/linter.test.ts
```diff
@@ -0,0 +1,386 @@
...
+const invalidNestedRulesetPath = resolve(__dirname, '__fixtures__/ruleset-extends-invalid.yaml');
+const standardOas3RulesetPath = resolve(__dirname, '../../../rulesets/oas3/index.json');
+const standardOas2RulesetPath = resolve(__dirname, '../../../rulesets/oas2/index.json');
+const draftRefSpec = resolve(__dirname, './__fixtures__/draft-ref.oas2.json');
+const draftNestedRefSpec = resolve(__dirname, './__fixtures__/draft-nested-ref.oas2.json');
+const validOas3SpecPath = resolve(__dirname, './__fixtures__/openapi-3.0-valid.yaml');
+const invalidOas3SpecPath = resolve(__dirname, '__fixtures__/openapi-3.0-no-contact.yaml');
+
+function run(command: string) {
+  const parser = yargs.command(lintCommand).help();
```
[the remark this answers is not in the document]
P0lip: Let's see.

src/cli/commands/__tests__/lint.test.ts
P0lip: It's a dupe, good catch.

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -1,448 +1,172 @@
...
 
-  describe('when using default ruleset file', () => {
-    let spy: SpyInstance;
-    beforeAll(() => {
-      spy = jest.spyOn(process, 'cwd').mockReturnValue(resolve(__dirname, '__fixtures__'));
+  it('errors upon exception', async () => {
+    const error = new Error('Failure');
+    (lint as jest.Mock).mockReset();
+    (lint as jest.Mock).mockReturnValueOnce({
+      // could be mockRejectedValueOnce, but Node 8 does not like it (different ticking?), so here is the workaround
```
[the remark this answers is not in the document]
P0lip: Returning a promise does not help. Basically the fn is never called.

package.json
```diff
@@ -51,21 +41,19 @@
...
+    "prebuild.binary": "yarn build",
     "pretest.karma": "node ./scripts/generate-karma-fixtures.js",
     "schema.update": "yarn typescript-json-schema --id \"http://stoplight.io/schemas/rule.schema.json\" --required tsconfig.json IRule --out ./src/meta/rule.schema.json",
     "test.harness": "jest -c ./jest.harness.config.js",
     "test.karma": "karma start",
     "test.prod": "yarn build && yarn lint && yarn test.karma && yarn test --coverage --maxWorkers=2",
     "test.update": "yarn test --updateSnapshot",
     "test.watch": "yarn test --watch",
-    "test": "jest"
+    "test": "jest --silent"
```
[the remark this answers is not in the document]
P0lip: We have a few console.* and they cause unnecessary noise. Should be improved once Logger is in place.

src/cli/services/__tests__/linter.test.ts
P0lip: Yeah, there are 2 types of paths there. Can you think of any other naming here? :D

src/cli/commands/lint.ts
P0lip: Can't do unfortunately. Fail method take 3 arguments.

src/cli/services/linter.ts
P0lip: It's okay if it's empty, isn't it? I mean, if one wants to do so, why not.

src/cli/commands/__tests__/lint.test.ts
P0lip: Yeah, seems fair.

src/cli/services/__tests__/linter.test.ts
P0lip: @philsturgeon thoughts? I don't really mind

src/cli/services/linter.ts
lag-of-death: if `rulesetFiles: string[]`, then you only have to worry about one data structure though

P0lip: Yeah, it's mostly for other developers' convenience, since that function is accessible.
I can move that to `loadRuleset` method, don't have any preference, up to you.

P0lip:
@philsturgeon @XVincentX @lag-of-death 
Could you have another pass folks?

src/cli/services/linter.ts
lag-of-death: heh, don't know what moving "that to loadRuleset" solves. If it is about limiting number of data structures, then yeah, let's go for it :)

src/cli/commands/lint.ts
```diff
@@ -1,261 +1,92 @@
...
+        showHelp();
```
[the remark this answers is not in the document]
P0lip: No, we just don't make use of signale, so didn't move that part.

Issue #437, opened 2019-08-12
Improve Spectral performance from O(n^2) to O(n)


chris-miaskowski:
This is what we have now:
- foreach rule
  - call jsonpath.find(rootObj, targetPath)
    - json path will recurse through relevant parts of object to find matches and return them

We want to have:
- foreach property in rootObject (recursively)
  - for each rule
    - call jsonpath.match(currentPath, targetPath)

Issue #272, opened 2019-06-26
Parameters in links objects shouldn't be linted for not having a description property.


garethj-msft:
### **I'm submitting a...**
  - [ x ] bug report

### What is the current behavior?

Lint an OAS 3 with links section, where a link object has a parameters collection containing the usual name to path dictionary.
Get warnings like this:

Rule:
Parameter objects should have a description.

Details:
paths./allSchools/createUploadSession.post.responses.201.links.getUploadSessionById.parameters.sessionId.description is not truthy

But the parameters collection in a links object is not a parameters object.

### What is the expected behavior?
No warning.

### What is the motivation / use case for changing the behavior?
Correctness.

### Please tell us about your environment:
Web version.

philsturgeon:
Ok this sounds like the rule for parameters is just a little greedy. The trouble with making them too specific is that then we have trouble linting on partial files with $ref, as if we just see the word parameters we dont know if its links parameters or operation parameters. 🤔

marbemac:
Hmm, right now we look for parameters objects that don't have a $ref key. While still not perfect, a better indicator might be check for presence of an `in` key. This would cover $ref and also fix the issue described here re link parameters property false identification.

Relevant rule: https://github.com/stoplightio/spectral/blob/develop/src/rulesets/oas/index.json#L344

PR #472, opened 2019-08-21
Handle example keyword validation

### LINKED ISSUE #223: Parameter with required and example fails valid-example rule

lag-of-death:
Fixes #223 .

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

# TL;DR

This PR adds all that is needed to validate `example` against its `schema` in both OAS2 and OAS3. Validation of `examples` is not handled. Also validation of `example` against encoding (OAS3) is also not handled.

# Findings:

I have not found a need for separate rules for example(s) in different places (like in a header, in body etc). The added rule and the adjusted one handle `example` validation against its schema. But, OAS3 talks about validating `example(s)` against its `encoding`. This is vary vague. Encoding, it seems, is defined in `Encoding Object` and `Encoding Object` is only to be found in `Media Type Object` (so in Response/Request Object(s)). This might need further investigation and maybe should be handled later (not in this PR).

If we are to validate `examples`, then there is a possible need for additional rules. Couldn't work out anything to validate `examples` with the current rules and `json-path-plus` magic.

### `example` keyword

The only place that allows `example` keyword is inside of `Schema Object`. Schema Object also allows `required` keyword. But this `required` keyword is an array, so we have no problem with https://github.com/stoplightio/spectral/blob/4307a0cfe6e7d03647090cdec5082313c8e5401e/src/functions/schema.ts#L36
because the required attribute will never be a boolean. It will always be an array (just because it is Schema Object, which is a set of JSON Schema.

Required attribute is also found in Parameter Object, but Parameter Object does not allow an example keyword. So there is no conflict there.

So, it seems to me, that the greed of https://github.com/stoplightio/spectral/blob/027664fcf7f58f542a323d338b9017ba18adc9a3/src/rulesets/oas2/index.json#L98
does not pose a problem and it's OK to leave it as is.

Links:

parameter object: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-7
schema object: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema-object

### `examples` keyword

To be found only in Response Object.

> The name of the property MUST be one of the Operation produces values (either implicit or inherited).
> The value SHOULD be an example of what such a response would look like.

from: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#example-object

I couldn't find a way to validate `examples` in any way with the existing rules.

### A spec

I'm attaching an OAS2 example with some comments to further clarify, that there is no issue with the existing OAS2 rule:

```yaml
swagger: "2.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: Swagger Petstore
  license:
    name: MIT
  contact:
    name: Test
paths:
  /pets/{petId}:
    get:
      summary: Info for a specific pet
      operationId: showPetById
      description: Get a pet
      tags:
        - pets
      parameters: # .example in any kind of parameter (any kind of in:) is not allowed
        - name: petId
          in: path
          type: integer
          required: true  # required is is allowed (and it is a boolean), but there never is an example alongside it
        - name: petSomething
          in: body
          required: true
          schema: # every schema is based on JSON Schema Specification Draft 4, which allows required (but as an array, so OK)
            type: object
            properties:
              abc:
                required: # required is allowed here, but only as an array
                  - age
                type: object
                example: 123
                properties:
                  age:
                    type: number
            example:
              abc: 888
      responses:
        '200':
          description: Expected response to a valid request
          examples:
            application/json: { "id": 38, "title": "T-shirt" }
              text/csv: >
                id,title
                38,T-shirt
definitions:
  Abc: # Scheme Object, no boolean required field allowed
    type: object
    properties:
      id:
        type: integer
        format: int64
        example:
          23233
      name:
        type: string
    required:
      - name
    example:
      name: Puma
      id: 1
```

### `example` key to be found in:

1. Parameter Object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-10

> The example SHOULD match the specified schema and encoding properties if present.

Not sure about "encoding properties" ?

2. Schema Object: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schema-object

> A free-form property to include an example of an instance for this schema. 

3. Media Type Object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-12

> Each media type includes a schema, defining the data type of the message body, and, optionally, one or several examples. 

from: https://swagger.io/docs/specification/media-types/

4. Header Object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#header-object

> The Header Object follows the structure of the Parameter Object with the following changes (...)

### `required` key to be found in:

1. Request Body Object: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#request-body-object
2. schema Object: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schema-object
3. Parameter Object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-10
4. Header Object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#header-object

### the conflict

Parameter object (and header object), where `required` field is a Boolean and `example` is allowed. This PR solves this.

```yaml
parameters:
    skipParam:
      name: skip
      in: query
      description: number of items to skip
      required: true  # boolean required and example both inside Parameter Object
      example: 34 # boolean required and example both inside Parameter Object
      schema:
        type: integer
        example: 34
        required: []
        format: int32
```

### `examples` keyword

All `examples` are built with `Example Object` https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#example-object

> "In all cases, the example value is expected to be compatible with the type schema of its associated value. "

#### To be found in:

1. parameter object ( https://swagger.io/docs/specification/adding-examples/  and  https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object )

2. Media Type object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#media-type-object in:
  a) requestBody ( https://swagger.io/docs/specification/adding-examples/ )
  b) response body: ( https://swagger.io/docs/specification/adding-examples/ )

> The examples in response and request bodies are free-form, but are expected to be compatible with the body schema.

3. components object https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#components-object

The same as with OAS2, I couldn't find a way to validate examples in any way with the existing rules (for OAS3).


lag-of-death:
@philsturgeon, could you take a look if you find the time? Thanks!

philsturgeon:
@lag-of-death sadly this does not solve the problem. You have added an extra conditional which does help hide the symptom of the bug, but it does not solve the feature. 

We talked about different types of rules because in OAS2 and OAS3 there are various different places the `example` keyword can show up, and it may or may not be inside a `schema`. The issue you are fixing showed this valid partial spec:

```
      parameters:
        - in: query
          name: petId
          # required: true
          schema:
            type: integer
          example: "3"
```

Your code will simply ignore this, and not do anything about it. 

I can find a bunch of other examples too. In OAS2 only some parameters have the `schema` keyword and some don't. 

`example` is not a valid keyword in OAS2 but that doesn't stop a bunch of people using it anyway. In OAS2 we can either ignore it or support it, but if we support it we should support it properly, and in OAS3 we need to support `example` in either of its valid locations above.

lag-of-death:
@philsturgeon, given this spec (that has your example included):
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: Swagger Petstore
  license:
    name: MIT
  contact:
    name: Test
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      summary: Info for a specific pet
      description: pets
      operationId: showPetById
      tags: [pets]
      parameters:
        - in: query
          name: petId
          required: true
          schema:
            type: integer
          example: "3"
      responses:
        '200':
          description: Expected response to a valid request
```
I get:
```
Adding OpenAPI 3.x functions
Method `addRules` has been deprecated since version 4.1, use `loadRuleset` instead.
    at /snapshot/spectral/dist/cli/commands/lint.js:182:18
OpenAPI 3.x detected

ok.yaml
 26:20  warning  valid-example  "0.example" property type should be integer

✖ 1 problem (0 errors, 1 warning, 0 infos)
```
Am I missing something?
I kinda still believe that the implementation is OK.

chris-miaskowski:
> @philsturgeon, given this spec (that has your example included):
> 
> ```yaml
> openapi: "3.0.0"
> info:
>   version: 1.0.0
>   title: Swagger Petstore
>   description: Swagger Petstore
>   license:
>     name: MIT
>   contact:
>     name: Test
> servers:
>   - url: http://petstore.swagger.io/v1
> paths:
>   /pets:
>     get:
>       summary: Info for a specific pet
>       description: pets
>       operationId: showPetById
>       tags: [pets]
>       parameters:
>         - in: query
>           name: petId
>           required: true
>           description: The id of the pet to retrieve
>           schema:
>             type: integer
>           example: "3"
>       responses:
>         '200':
>           description: Expected response to a valid request
> ```
> 
> I get:
> 
> ```
> Adding OpenAPI 3.x functions
> Method `addRules` has been deprecated since version 4.1, use `loadRuleset` instead.
>     at /snapshot/spectral/dist/cli/commands/lint.js:182:18
> OpenAPI 3.x detected
> 
> ok.yaml
>  26:20  warning  valid-example  "0.example" property type should be integer
> 
> ✖ 1 problem (0 errors, 1 warning, 0 infos)
> ```
> 
> But.. hm, yeah, I think I'll try to narrow the validation to places where `example` can exist. I think I was wrong in assuming that `example` will be put (by users) only in places where it should be put.

One comment to add to this is as far as I remember the original request asked for `error` rather than `warning`, right?

lag-of-death:
@chris-miaskowski, thanks, yes, it should be reported as errors

src/rulesets/oas3/index.json
```diff
@@ -83,7 +83,20 @@
       "message": "{{error}}",
       "recommended": true,
       "type": "validation",
-      "given": "$..[?(@.example)]",
+      "given": "$..[?(@.example && !@.type)]",
```
chris-miaskowski: I was about to suggest that you could try narrowing it down by `schema` but then I realized it won't work out of the box. There would still be a corner case where someone has an example but didn't provide a schema.

On a second thought I will suggest it :)

```suggestion
      "given": "$..[?(@.example && !@.schema)]",
```

If I understand this query correctly it will match all 'nodes' that have `example` but doesn't have `schema` next to it. This could mean two things:
- either the `example` you found is **inside** the `schema`
- or the `example` should be next to `schema` but that `schema` was not provided!

If it was not provided you will get `undefined` but that's fine because you should be able to check that in the actual rule implementation.

src/rulesets/oas3/index.json
```diff
@@ -83,7 +83,20 @@
...
+          "field": "example",
+          "schemaPath": "$.schema"
+        }
+      }
+    },
+    "valid-type-example": {
+      "summary": "Examples must be valid against their type.",
+      "message": "{{error}}",
+      "type": "validation",
+      "given": "$..[?(@.example && @.type)]",
```
chris-miaskowski: ditto

The problem with `type` is that I'm not sure there isn't a nasty corner case where that would fail.

lag-of-death:
thanks @chris-miaskowski, I will add the tests :+1:
about:
> Furthermore, if referencing a schema that contains an example, the example value SHALL override the example provided by the schema.

I have stumbled upon this, but my understanding was rather that `example` value should take precedence over that of `scheme.example` when generating a response for example. But still, even if `scheme.example` is overwritten, it ought to be validated. Not sure though, but basically was following this logic.

src/rulesets/oas3/index.json
P0lip: Once we are at it, is there any way to make the json path expression more specific?
I understand `$..[?(@.example && !@.type)]` feels very handy, yet it's utterly slow to the point where we have that rule turned off in Studio.
The other issue I see with that expression is the use of function filter expression that accesses the properties (`@.example && @.type`). This will pretty much impossible to support when https://github.com/stoplightio/spectral/issues/437 becomes a thing.

If there is no way to make it less general, I'm okay with leaving it as is.

src/rulesets/oas3/index.json
```diff
@@ -83,7 +83,20 @@
...
+      "given": "$..[?(@.example && !@.type)]",
+      "then": {
+        "function": "schemaPath",
+        "functionOptions": {
+          "field": "example",
+          "schemaPath": "$.schema"
+        }
+      }
+    },
+    "valid-type-example": {
```
P0lip: Should it be a recommended rule, or no?
I'd rather disabled by default.

philsturgeon:
Bang on I need to double check the contents of the message @chris-miaskowski wrote I’m pretty sure it’s not entirely correct.

We are definitely not talking about media type objects so don’t worry about that.

src/rulesets/oas3/index.json
philsturgeon: Yeah I was trying to get more specific and less greedy rules like intentionally pretending with “parameters” so we know we’re looking st parameters and can make the rules more specific seeing as different types of parameters may or may not have a schema keyword, etc.

src/rulesets/oas3/index.json
philsturgeon: We can make it not recommended until we are more confident with it, and bump it up in v5?

src/rulesets/oas/__tests__/valid-example.ts
```diff
@@ -1,61 +1,101 @@
...
-import * as ruleset from '../index.json';
+import * as ruleset from '../../oas2/index.json';
 
 // @oclif/test packages requires @types/mocha, therefore we have 2 packages coming up with similar typings
 // TS is confused and prefers the mocha ones, so we need to instrument it to pick up the Jest ones
 declare var test: jest.It;
 
 describe('valid-example', () => {
   const s = new Spectral();
-  s.setRules({
```
philsturgeon: setRules is a handy new way of doing things that simplifies the API a little, and addRules is deprecated now.

Also the convention here has been for `__tests__/{foo}.ts` to match the rule name it is testing, but this test file is now testing two different rules.

@P0lip can explain a bit more about the recent changes in the spectral develop branch if you are not aware of what has happened in PRs like #430.

lag-of-death: thanks, I don't know how to load a single rule with `loadRuleset`, is it possible @P0lip ?

philsturgeon: I said setRules not loadRuleset. 😅

lag-of-death: ah, yes, sorry. My bad

chris-miaskowski:
Hey @philsturgeon , @lag-of-death I double check one of the examples I provided
```
parameters:
  - in: query
    name: petId
    required: true
    example: "3"
```

I said it "should be OK" but according to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject `type` is a required field.

Which means... not sure what. We should probably say it's **NOT** ok, but at the same time it looks like this situation in undefined.

WDYT?

lag-of-death:
> Hey @philsturgeon , @lag-of-death I double check one of the examples I provided
> 
> ```
> parameters:
>   - in: query
>     name: petId
>     required: true
>     example: "3"
> ```
> 
> I said it "should be OK" but according to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject `type` is a required field.
> 
> Which means... not sure what. We should probably say it's **NOT** ok, but at the same time it looks like this situation in undefined.
> 
> WDYT?

@chris-miaskowski, from what I know, `Parameter Object` does not have `.example` keyword in `OAS2`, it's possible in `OAS3` though. And `types` is only required if `in !== 'body'`

chris-miaskowski:
> > Hey @philsturgeon , @lag-of-death I double check one of the examples I provided
> > ```
> > parameters:
> >   - in: query
> >     name: petId
> >     required: true
> >     example: "3"
> > ```
> > 
> > 
> > I said it "should be OK" but according to https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject `type` is a required field.
> > Which means... not sure what. We should probably say it's **NOT** ok, but at the same time it looks like this situation in undefined.
> > WDYT?
> 
> @chris-miaskowski, from what I know, `Parameter Object` does not have `.example` keyword in `OAS2`, it's possible in `OAS3` though. And `types` is only required if `in !== 'body'`

Bummers! Yes, you are right I looked at the incorrect doc! 

OAS3 says
>  The example SHOULD match the specified schema and encoding properties if present
It doesn't say it MUST, though. So I _guess_ my example is valid (???) @philsturgeon @lag-of-death

lag-of-death:
@chris-miaskowski, from the spec:
>The encoding object SHALL only apply to requestBody objects when the media type is multipart or application/x-www-form-urlencoded

from this PR's description:
> But, OAS3 talks about validating example(s) against its encoding. This is vary vague. Encoding, it seems, is defined in Encoding Object and Encoding Object is only to be found in Media Type Object (so in Response/Request Object(s)). This might need further investigation and maybe should be handled later (not in this PR).
bottom line: didn't care about `encoding` as it would make this PR stuck in progress/review for ages

Also, I think that `SHOULD match the specified schema and encoding properties **if present**`. This answers it (although it is a bit vague). But, since you have no `.schema`, then you're example is valid

philsturgeon:
The Encoding Object lives inside the Media Type Object which does have example and examples but remember that this PR is not about Media Type Object example/examples. That example is the `Example Object` found [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#exampleObject).

This is about the `example` keyword that lives in properties and inside `schema` and has nothing to do with encoding.

src/rulesets/oas3/index.json
```diff
@@ -78,6 +78,111 @@
         }
       }
     },
+    "valid-oas-example-in-parameters": {
```
P0lip: Whoah, whoah, whoah. Why so many rules.
We could create a custom function and utilize `functionOptions`.

src/rulesets/oas3/index.json
```diff
@@ -78,6 +78,111 @@
...
+      "description": "Examples must be valid against their defined schema.",
+      "message": "{{error}}",
+      "recommended": true,
+      "severity": 0,
+      "type": "validation",
+      "given": "$..parameters..[?(@.example && @.schema)]",
```
P0lip: Paths are too loose.

src/rulesets/oas3/index.json
lag-of-death: @philsturgeon, I only added rules, like you said - no new functions. But maybe again I'm mistaken? cc @P0lip

src/rulesets/oas3/index.json
lag-of-death: @P0lip, what path for OAS3 `schema` next to `example` in `parameters` you'd rather see? I can't think of anything better atm

src/rulesets/oas3/index.json
P0lip: We could have a `valid-oas-example` rule and simply instrument it using options, like so
```yaml
# your ruleset
valid-oas-example: [error, {
  in: [parameters, headers, content] # etc
}]
```

src/rulesets/oas3/index.json
P0lip: I don't really know, but this is going to be utterly slow.
@philsturgeon can you think of anything better?

philsturgeon: I was wondering about this. Would it be better performance wise to:

1. run on all parameters then check for example and schema keys
2. run on all parameters that have an example then check for schema

Or is it the `$..parameters` bit that's the problem? There is an issue where #272 points out there are different types of parameters in the [Link Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#linkObject) so we don't want to double up on that mistake.

src/rulesets/oas3/index.json
lag-of-death: sorry @P0lip, I don't quite follow, where ^ would go?

philsturgeon: @P0lip there are a lot of different ways in which examples can show up, so we were trying to get them done sharing the same generic schemaPath function, as the only difference between most of the types of differences are example and schema are siblings, or example is nested inside schema, or both, and that's just parameters. 

This PR ignores Media Type Object `example` or `examples` as explained elsewhere.

@P0lip this PR has been going on for a long time and people have wandered in and started asking lots of questions, then we've explained it, then we have started the process over again with another person. You should definitely get caught up but doing it all over text would take a fairly long time. If you two can have a call (or I can jump in too tomorrow) then it might be easiest. 

```
      parameters:
        - in: query
          name: petId
          required: true
          schema:
            type: integer
            example: "3"
          example: "5"
```

Check the tests to see all sorts of examples of parameters in different places. 

If you think it'll be drastically better for performance we can switch approaches another time and make `osOas3ParameterExamples` look for `example` in parameter objects  and `osOas3SchemaExamples` that will go through all schema's that have examples nested anywhere in them or whatever. 

There's a million ways to do this we just gotta pick one that works and that isn't slow. The performance issues were discovered after this approach was decided upon.

src/rulesets/oas3/index.json
P0lip: > Or is it the $..parameters bit that's the problem? 

Yeah, it's `$..parameters` that is the problem, if we could narrow it down a bit, that would be great.
I don't think we are interested in parameters under info object, are we?
If I am not mistaken, parameters are bound to an endpoint/path, so we could at least have paths there, i.e. `$.paths..parameters`.
I truly doubt we aim to validate every property with `parameters` key, so let's limit the scope as much as we can.

src/rulesets/oas3/index.json
P0lip: >  The performance issues were discovered after this approach was decided upon.

I have known for at least a month or so. I might have not communicated that though, so it's on me.

Yeah, I understand the background behind this PR and I and don't really want to prevent this PR from being merged. I am aware lots of folks are waiting for that. On the other hand, we are introducing plenty of new rules that will be disabled in Studio anyway until the performance thingy is completed, which, by the way, might require these rules to be rewritten anyway.

Let's sync tomorrow in the morning or so.
I believe a custom function based approach shouldn't take longer than half of a day / day, so we should be fine after all. Either way, as I said, I won't block this PR.
If you all really want to ship it, I'm fine with it. I'll just create a tech debt ticket to be addressed in 4.2 or 5.0 in such a case.

lag-of-death:
hi @P0lip 

**For OAS2:**

Parameters can be found in:
```
a) Path Item object -> LIKE $.paths.pets.parameters 
b) Operation Object -> LIKE $.paths.pets.get.parameters 
c) Parameters Definitions Object -> $.parameters
```
**For OAS3:**

Parameters can be found in:
```
a) Components Object -> $.components.parameters
b) Path Item Object  -> LIKE $.paths.pets.parameters
c) Operation Object  -> LIKE $.paths.pets.get.parameters
```
So, it seems like 3 rules (or two, one with `path`, and the other for the definitions) instead of one for both OAS2/OAS3. Or one custom rule you mentioned.

What I could do for this PR is to have these two rules instead of one. But I think it would be best to create that tech debt card and handle performance issues there, not here.

Maybe we can take a look at https://github.com/stoplightio/spectral/blob/129a23d582558074266c44e3e7f4a2e0fdcf1c71/src/rulesets/oas/index.json#L345 too in that tech debt?

LMK :)
cc @philsturgeon

P0lip:
`parameter-description` is turned off by default, so it's quite okay, but yeah, we should improve it as well.
Frankly speaking, the parameters case just feels like a perfect candidate for a custom oas2 and/or oas3 function.
Either way, I'm quite okay with leaving it as is for now.
Could perhaps dedupe some of the tests, i.e. the ones checking format and such - it's enough to have them in one place.

lag-of-death:
@P0lip, tests:
1. `reports invalid usage of %s format`
2. `does not report valid usage of %s format`
3. `does not report example mismatches for unknown AJV formats`
are now to only be found in:
1. `src/rulesets/oas3/__tests__/templates/_oas-example.ts`
2. `src/rulesets/oas3/__tests__/templates/_schema-example.ts`
I've left them there because of the difference in `.example` location. LMK if this is enough :)

philsturgeon:
@P0lip let's guide this one in for a landing and improve in v5.0 with cleverly made performant OAS2 and OAS3-specific custom functions. Can you make a ticket for that?

P0lip:
@philsturgeon @lag-of-death 

> Can you make a ticket for that?

Coolio. Will do today.
Looking good for me then, let's merge it.

lag-of-death:
@P0lip, @philsturgeon, @chris-miaskowski, can anyone approve?

src/rulesets/oas2/index.json
```diff
@@ -90,6 +90,36 @@
         "operation"
       ]
     },
+    "valid-example-in-parameters": {
+      "description": "Examples must be valid against their defined schema.",
+      "message": "{{error}}",
+      "recommended": true,
```
P0lip: Does the rule have to be recommended? @philsturgeon 
It's expensive, so if we can have it disabled by default, that'd be great.
If you find it particularly useful, I'm fine with leaving it as is.

src/rulesets/oas2/index.json
```diff
@@ -90,6 +90,36 @@
...
+        "functionOptions": {
+          "field": "example",
+          "schemaPath": "$"
+        }
+      }
+    },
+    "valid-example-in-definitions": {
+      "description": "Examples must be valid against their defined schema.",
+      "message": "{{error}}",
+      "recommended": true,
```
P0lip: Ditto

src/rulesets/oas3/index.json
```diff
@@ -78,6 +78,111 @@
         }
       }
     },
+    "valid-oas-example-in-parameters": {
+      "description": "Examples must be valid against their defined schema.",
+      "message": "{{error}}",
+      "recommended": true,
```
P0lip: Ditto

src/rulesets/oas3/index.json
```diff
@@ -78,6 +78,111 @@
...
+        "functionOptions": {
+          "field": "example",
+          "schemaPath": "$.schema"
+        }
+      }
+    },
+    "valid-oas-example-in-headers": {
+      "description": "Examples must be valid against their defined schema.",
+      "message": "{{error}}",
+      "recommended": true,
```
P0lip: Ditto

src/rulesets/oas3/index.json
```diff
@@ -78,6 +78,111 @@
...
+        "functionOptions": {
+          "field": "example",
+          "schemaPath": "$.schema"
+        }
+      }
+    },
+    "valid-oas-example-in-content": {
+      "description": "Examples must be valid against their defined schema.",
+      "message": "{{error}}",
+      "recommended": true,
```
P0lip: Same for all other rules.

src/rulesets/oas2/index.json
philsturgeon: This is a pretty important feature, people have reported spectral as broken when its not run, so leaving it to only people who need to know about it and turn it on seems... unideal. People can disable slow rules if they want but honestly Studio is probably one of the few scenarios where speed matters?  (That comment may not age well, but for right now...)

P0lip: > People can disable slow rules if they want but honestly Studio is probably one of the few scenarios where speed matters? 

Yeah, it's because of the UI. In case of CI it usually doesn't really matter whether linting takes 5 seconds or 20 seconds... unless you run out of memory which is what the issue is all about.
So in the end, if one lints a huge document, such as Sendgrid or Stripe, it **will** matter.

philsturgeon: Ok, lets get this feature in and it can be a bit slow, as we know we are working on #437 anyway.

P0lip: Either way, let's just merge it.
I'll create a follow up ticket for this feature today and will try to address it ASAP

philsturgeon: If anyone has their computer catch on fire im not paying the bill, but hopefully they can give us some beefy description docs to play with and we can use them for benchmarking.

P0lip: > If anyone has their computer catch on fire im not paying the bill, but hopefully they can give us some beefy description docs to play with and we can use them for benchmarking.

:laughing: we should be fine.
4.2 or 5.0 will certainly come with optimized performance, and in the meantime one can always disable the rules.

lag-of-death:
@chris-miaskowski, can you approve / can I dismiss the request for changes?

philsturgeon:
@lag-of-death yeah you can dismiss old reviews if you're confident you've done the thing.

Issue #394, opened 2019-07-19
oas core rule: $ref cannot have siblings


philsturgeon:
**User story.**
As a OpenAPI v2 or v3 user, I am completely unaware of the very subtle mentions that $ref cannot have siblings, and I would like Spectral to let me know about this so I don't write invalid specs.

**Is your feature request related to a problem?**

The OpenAPI v3.0 spec says that any properties sat next to a $ref SHALL be ignored, but some tooling does not ignore them. 

<img width="912" alt="Screen Shot 2019-07-19 at 16 10 55" src="https://user-images.githubusercontent.com/67381/61541476-029d4180-aa40-11e9-98e6-24c1ea476aaa.png">

This can lead to a spooky weird situation, where any tools ignoring those properties seem broken, when really they are abiding the specification. They any tools which do support the sibling properties look like they're working, but are actually going against the spec. 

Relying on accidental undefined behavior is not gonna be a good time.

**Describe the solution you'd like**
The argument could be made for error or warning. 

**error:** technically what you've written goes against the spec, and that is generally what we've been using to classify something as an error, buuuut...

**warn:** those extra properties are being ignored, which _should_ be fine with spec abiding tooling.

Dear Users, please let us know if it should be an error or a warning. 

If no clear winner happens in the comments, this can be implementers choice.

**Additional context**

Some folks think $ref siblings are allowed in OpenAPI v2.0 because the spec does not forbid it, but really I think they just forgot to mention it. OpenAPI v2.0 does say its a JSON Reference, and links to https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03, which also says siblings SHALL be ignored:

<img width="677" alt="Screen Shot 2019-07-19 at 16 19 14" src="https://user-images.githubusercontent.com/67381/61542054-18f7cd00-aa41-11e9-9e67-db05b1ee27b1.png">

Both OpenAPI v2.0 and v3.0 should be treated the same. But, interestingly, if OpenAPI v3.1 gets bumped up to JSON Schema Draft 8, then $ref siblings _will_ be allowed. That's probably another story for another time, but for now lets make both 2.0 and 3.0 do the same thing.

PR #473, opened 2019-08-21
feat: $ref cannot have siblings

P0lip:
Closes https://github.com/stoplightio/spectral/issues/394

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

It's not a rule, as rules have no access to original (unresolved) data.
I don't know if it's a big deal or not, but if we do need a rule, I can make it a rule.
The code is placed in a wrong place and an extra test wouldn't hurt, so it's still a WIP.


P0lip:
@philsturgeon 
> It's not a rule, as rules have no access to original (unresolved) data.
> I don't know if it's a big deal or not, but if we do need a rule, I can make it a rule.

How do you feel about not making it a rule and failing all the time same as invalid-ref does?

P0lip:
@philsturgeon got it, will make it a rule then.

P0lip:
@philsturgeon 
I'm planning to add an extra property to a rule named `unresolved` or something, which would make each rule operate on initial data

P0lip:
Hold on with the review, forgot to add one test.

src/types/rule.ts
```diff
@@ -26,6 +26,9 @@ export interface IRule<T = string, O = any> {
   // Filter the target down to a subset[] with a JSON path
   given: string;
 
+  // If true, rule will operate on original (unresolved) data
+  unresolved?: boolean;
```
philsturgeon: Would it be better to swap? `unresolved: false` sounds a bit awkward and double-negative. Would something like `isResolved: true` be easier for others to read and reason about?

rollup.config.js
```diff
@@ -7,22 +7,21 @@ import { terser } from 'rollup-plugin-terser';
...
+  'refSiblings'
 ].map(fn => ({
   input: path.resolve(BASE_PATH, 'dist/rulesets/oas/functions', `${fn}.js`),
   plugins: [
     typescript({
       tsconfig: path.join(BASE_PATH, './tsconfig.rollup.json'),
       include: ['dist/**/*.{ts,tsx}'],
     }),
-    resolve({
-      only: ['json-schema-merge-allof', /lodash\/?.*/],
```
philsturgeon: I'm not sure what `json-schema-merge-allof` but is it a dependency we can remove now?

P0lip: @philsturgeon actually it's a leftover :laughing: 
I copied that rollup config from json-schema-viewer repo and forgot to remove it.
We don't have that package included in our package.json, so we should be okay.

src/rulesets/oas/index.json
```diff
@@ -428,6 +429,18 @@
           "schemaPath": "$"
         }
       }
+    },
+    "$ref-siblings": {
+      "description": "$ref cannot be placed among any other property",
```
philsturgeon:
```suggestion
      "description": "$ref cannot be placed next to any other properties",
```

src/rulesets/oas/index.json
```diff
@@ -428,6 +429,18 @@
           "schemaPath": "$"
         }
       }
+    },
+    "$ref-siblings": {
```
philsturgeon:
```suggestion
    "no-$ref-siblings": {
```

src/types/rule.ts
```diff
@@ -26,6 +26,10 @@ export interface IRule<T = string, O = any> {
   // Filter the target down to a subset[] with a JSON path
   given: string;
 
+  // If false, rule will operate on original (unresolved) data
```
philsturgeon: It seems like this will update our docs a bit? Could you add mention of it to the custom rules section?

P0lip: @philsturgeon added a short note, lmk what you think.

docs/getting-started/rulesets.md
```diff
@@ -40,6 +40,24 @@ While running it with this object, it will succeed:
...
+By default, Spectral processes each rule on resolved document with all $refs resolved.
+If you would like to have an original input supplied to your rule, you can place `isResolved` property as follows:
+
+```yaml
+rules:
+  my-rule-name:
+    description: Tags must have a description.
+    given: $.tags[*]
+    severity: error
+    isResolved: false # note - if not specified or true, a resolved document will be given
```
philsturgeon: I'm sorry this is a switch and not a result flag. I missed it before but now it makes sense. Can we make it `resolved: true/false`?  😅

PR #477, opened 2019-08-22
fix(cli): resolve relative document paths

### LINKED ISSUE #474: Fail to load external reference file because a '/' is prepend to the path
**Describe the bug**
Spectral is unable to load external reference file because a `/` is prepend to the path.

**To Reproduce**

1. Given this OpenAPI document and referenced document in same folder:

index.yaml

```yaml
openapi: 3.0.0
info:
  title: Test
  version: 1.0.0
paths:
  '/':
    get:
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: 'common.yaml#/components/schemas/ResponseType'
                # Below doesn't work as well
                # $ref: './common.yaml#/components/schemas/ResponseType'
```

common.yaml

```yaml
openapi: 3.0.0
info:
  title: Common schema definition
  version: 1.0.0
components:
  schema:
    ResponseType:
      type: string
```

2. Run this CLI command 
```
spectral lint index.yaml
```

3. See error

```
Adding OpenAPI 3.x functions
OpenAPI 3.x detected

index.yaml
  1:1   warning  api-servers            OpenAPI `servers` must be present and non-empty array.
  2:6   warning  info-contact           Info object should contain `contact` object.
  2:6   warning  info-description       OpenAPI object info `description` must be present and non-empty string.
  7:9   warning  operation-description  Operation `description` must be present and non-empty string.
  7:9   warning  operation-operationId  Operation should have an `operationId`.
  7:9   warning  operation-tags         Operation should have non-empty `tags` array.
 14:23    error  invalid-ref            ENOENT: no such file or directory, open '/common.yaml'

✖ 7 problems (1 error, 6 warnings, 0 infos)

```

**Expected behavior**
There should not be an `invalid-ref` error. The reference file path should be `./common.yaml` instead of  `/common.yaml`

@stoplight/spectral/4.0.2 darwin-x64 node-v12.8.0

P0lip:
Fixes #474

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


src/__tests__/__fixtures__/gh-474/spec.yaml
```diff
@@ -0,0 +1,14 @@
...
+paths:
+  '/':
+    get:
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: 'common.yaml#/components/schemas/ResponseType'
```
philsturgeon: Can we get another file identical to this with ./common.yaml too? We want to show it’s definitely working both ways.

P0lip: Yup, will do in a moment.

src/cli/commands/__tests__/lint.test.ts
```diff
@@ -96,6 +96,20 @@ describe('lint', () => {
         });
       });
     });
+
+    test
+      .stdout()
+      .command(['lint', 'src/__tests__/__fixtures__/gh-474/spec.yaml'])
+      .it('should handle relative path to a document', ctx => {
```
philsturgeon: Sorry I didn’t spit before but can we assert a positive as proof? This sort of double negative doesn’t catch against situations where instead of invalid-ref we change the code to say broken-ref and nobody knows that refs aren’t working properly anymore. We can keep the negative assertion but we should add positive assertions too.

P0lip: > We can keep the negative assertion but we should add positive assertions too.

Yeah, I was thinking of it as well.
I'll introduce some errors in a referenced document.

P0lip: Ha, adding a test made me discover a subtle and unrelated to the original issue error.

P0lip: Alright, the above is related to caching... which made me discover yet another bug :laughing: this time in json-ref-resolver

PR #496, opened 2019-08-28
fix: generate proper paths and ranges for documents with $refs

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/498 

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

Discovered when working in Studio. ~It's not completed yet.~
There is still one weird issue - duplicate validation results. ~I believe my PR could have introduced it, so will fix it here.~
Nope, my code does not seem to be the cause of the issue. ~I'll file a bug report shortly.~ https://github.com/stoplightio/spectral/issues/500 it is 

~Blocked by https://github.com/stoplightio/json/pull/31~

### DIFF
--- a/src/__tests__/spectral.jest.test.ts
+++ b/src/__tests__/spectral.jest.test.ts
@@ -141,12 +141,32 @@ describe('Spectral', () => {
       expect.objectContaining({
         code: 'path-params',
         path: ['paths', '/todos/{todoId}'],
+        range: {
+          end: {
+            character: 5,
+            line: 45,
+          },
+          start: {
+            character: 23,
+            line: 19,
+          },
+        },
         source: undefined,
       }),
       expect.objectContaining({
         code: 'operation-description',
         path: ['paths', '/todos/{todoId}', 'get'],
         source: undefined,
+        range: {
+          end: {
+            character: 7,
+            line: 44,
+          },
+          start: {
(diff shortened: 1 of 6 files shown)


src/__tests__/spectral.jest.test.ts
```diff
@@ -72,4 +76,78 @@ describe('Spectral', () => {
       });
     });
   });
+
+  test('reports issues for correct files', async () => {
```
P0lip: Note, this test could land in CLI tests as we have 2-3 test cases covering `$refs` there.
I avoided doing it for now, as it would collide with "oclif to yargs" PR.
We can move it later if needed, but I find `spectral.jest.test.ts` a decent spot either.

philsturgeon: Throw a comment in there saying `// TODO could move to spectral.test.ts after #471` or something

P0lip: It's already in `spectral.jest.test.ts` :D
What I'm saying is that we have a few tests covering `$refs` placed in tests for cli lint command, therefore we could potentially have all of them in one place rather than spread across 2 files.

P0lip:
Once this one gets merged, we can release another beta.

PR #501, opened 2019-08-29
fix(encoding): use utf8 instead of utf-8

P0lip:
**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

The fs library we use in Studio expects `utf8` to be provided and throws otherwise.
Node seems to be fine with both.


wmhilton:
whaaaaaat 🤦‍♂

wmhilton:
for the record, the _documented_ one is `utf8` with no hyphen.
https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings

image

PR #503, opened 2019-08-29
fix: supress unknown format AJV error

chris-miaskowski:
Fixes https://github.com/stoplightio/spectral/issues/396

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Removes AJV warnings starting with 'unknown format'

### DIFF
--- a/src/functions/schema.ts
+++ b/src/functions/schema.ts
@@ -1,21 +1,34 @@
 import * as AJV from 'ajv';
-import * as jsonSpecv4 from 'ajv/lib/refs/json-schema-draft-04.json';
-const oasFormatValidator = require('ajv-oai/lib/format-validator');
 import { ValidateFunction } from 'ajv';
+import * as jsonSpecv4 from 'ajv/lib/refs/json-schema-draft-04.json';
 import { IOutputError } from 'better-ajv-errors';
 import { JSONSchema4, JSONSchema6 } from 'json-schema';
 import { IFunction, IFunctionResult, ISchemaOptions } from '../types';
+const oasFormatValidator = require('ajv-oai/lib/format-validator');
 const betterAjvErrors = require('better-ajv-errors/lib/modern');
 
 interface IAJVOutputError extends IOutputError {
   path?: string;
 }
 
+const logger = {
+  warn(...args: any[]) {
+    const firstArg = args[0];
+    if (typeof firstArg === 'string') {
+      if (firstArg.startsWith('unknown format')) return;
+      console.warn(...args);
+    }
+  },
+  log: console.log,
+  error: console.error,
+};
+
 const ajv = new AJV({
   meta: false,
   schemaId: 'auto',
   jsonPointers: true,
   unknownFormats: 'ignore',
+  logger,
 });
 ajv.addMetaSchema(jsonSpecv4);
 // @ts-ignore
--- a/src/functions/__tests__/schema.test.ts
+++ b/src/functions/__tests__/schema.test.ts
@@ -6,6 +6,27 @@ function runSchema(target: any, schemaObj: object) {
 }
 
 describe('schema', () => {
+  describe('when schema defines unknown format', () => {
+    const testSchema = {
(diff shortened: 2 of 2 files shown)


chris-miaskowski:
hey folks, what's that btw https://circleci.com/gh/stoplightio/spectral/5994?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ?? It isn't my fault :(

src/functions/schema.ts
```diff
@@ -1,21 +1,34 @@
...
 const betterAjvErrors = require('better-ajv-errors/lib/modern');
 
 interface IAJVOutputError extends IOutputError {
   path?: string;
 }
 
+const logger = {
+  warn(...args: any[]) {
+    const firstArg = args[0];
+    if (typeof firstArg === 'string') {
```
P0lip: I believe we want to hide those messages only for oas2/oas3 documents, don't we?

chris-miaskowski:
> See my comment

What comment? I can't see any.

src/functions/schema.ts
chris-miaskowski: @P0lip I don't think so. I think we want to hide for everyone trying to validate JSON Schema. According to the spec unknown values are OK.

P0lip: https://github.com/epoberezkin/ajv/issues/466#issuecomment-296290767 this was posted by Phil in the issue.
I agree with that. In general, we do want to log that. If we don't, then the fix is just a one-liner -> `unknownFormats: 'ignore' :P

chris-miaskowski: @P0lip `unknownFormats: ignore` does not suppress logging it replaces errors with `warn` logs.

chris-miaskowski: See that we already had this option turned on

```
const ajv = new AJV({
  meta: false,
  schemaId: 'auto',
  jsonPointers: true,
  unknownFormats: 'ignore',
  logger,
});
```
https://github.com/stoplightio/spectral/blob/94b91c32a5b854d6a896d9aee396347447ad450b/src/functions/schema.ts#L30

P0lip: I know we had it set up, I just thought there is a `false` option lol, but there isn't any :D

src/functions/schema.ts
```diff
@@ -1,21 +1,34 @@
...
 import { IFunction, IFunctionResult, ISchemaOptions } from '../types';
+const oasFormatValidator = require('ajv-oai/lib/format-validator');
 const betterAjvErrors = require('better-ajv-errors/lib/modern');
 
 interface IAJVOutputError extends IOutputError {
   path?: string;
 }
 
+const logger = {
+  warn(...args: any[]) {
```
P0lip:
```suggestion
  ...console,
  warn(...args: any[]) {
```

src/functions/schema.ts
```diff
@@ -1,21 +1,34 @@
...
+    const firstArg = args[0];
+    if (typeof firstArg === 'string') {
+      if (firstArg.startsWith('unknown format')) return;
+      console.warn(...args);
+    }
+  },
+  log: console.log,
```
P0lip:
```suggestion
```

src/functions/schema.ts
```diff
@@ -1,21 +1,34 @@
...
+  error: console.error,
```
P0lip:
```suggestion
```

P0lip:
Feel free to apply these 3 extra suggestions and then let's merge it.

src/functions/schema.ts
chris-miaskowski: Hey, not sure that's necessary. `logger` expects only three functions. it doesn't need everything console has. I think that'd be over-engineering it and I will leave it the way it is if you don't mind.

P0lip: A gotcha, np.

PR #514, opened 2019-08-30
chore(resolver): remove redundant parseResolveResult

P0lip:
The code is never really executed, since we pass our own `parseResolveResult` in `Spectral#run`.


lag-of-death:
@P0lip, can you point to a place where "we pass our own parseResolveResult in Spectral#run" ? would be happy to approve this :) always nice to see code go ;)

marbemac:
@lag-of-death https://github.com/stoplightio/spectral/blob/develop/src/spectral.ts#L88

PR #515, opened 2019-08-30
Do not validate description in parameters for links

### LINKED ISSUE #272: Parameters in links objects shouldn't be linted for not having a description property.

lag-of-death:
Fixes #272

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/src/rulesets/oas/index.json
+++ b/src/rulesets/oas/index.json
@@ -340,12 +340,12 @@
     },
     "parameter-description": {
       "description": "Parameter objects should have a `description`.",
-      "recommended": false,
+      "recommended": true,
       "type": "style",
-      "given": "$..parameters[*]",
+      "given": "$..[?(@property !== 'links' && @.parameters)]['parameters'][*]",
       "when": {
-        "field": "@key",
-        "pattern": "^(?!.*(\\$ref))"
+        "field": "in",
+        "pattern": "^(?!.*(in))"
       },
       "then": {
         "field": "description",
--- a/src/rulesets/oas/index.json
+++ b/src/rulesets/oas/index.json
@@ -340,9 +340,9 @@
     },
     "parameter-description": {
       "description": "Parameter objects should have a `description`.",
-      "recommended": true,
+      "recommended": false,
       "type": "style",
-      "given": "$..[?(@property !== 'links' && @.parameters)]['parameters']..[?(@.in)]",
+      "given": "$..[?(@path === \"$['parameters']\" || @path === \"$['components']['parameters']\" || @path === \"$['paths']\")]..[?(@.in)]",
       "then": {
         "field": "description",
         "function": "truthy"
--- a/src/rulesets/oas/index.json
+++ b/src/rulesets/oas/index.json
@@ -342,7 +342,7 @@
       "description": "Parameter objects should have a `description`.",
       "recommended": false,
       "type": "style",
-      "given": "$..[?(@path === \"$['parameters']\" || @path === \"$['components']['parameters']\" || @path === \"$['paths']\")]..[?(@.in)]",
+      "given": "$..[?(@path === \"$['parameters']\" || (@.parameters && @parentProperty !== 'links'))]..[?(@.in)]",
       "then": {
         "field": "description",
         "function": "truthy"
--- a/src/rulesets/oas/index.json
+++ b/src/rulesets/oas/index.json
@@ -338,19 +338,6 @@
         "operation"
       ]
     },
-    "parameter-description": {
-      "description": "Parameter objects should have a `description`.",
-      "recommended": false,
-      "type": "style",
-      "given": "$..[?(@path === \"$['parameters']\" || (@.parameters && @parentProperty !== 'links'))]..[?(@.in)]",
-      "then": {
-        "field": "description",
-        "function": "truthy"
-      },
-      "tags": [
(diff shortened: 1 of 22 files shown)


src/rulesets/oas/index.json
```diff
@@ -342,11 +342,7 @@
       "description": "Parameter objects should have a `description`.",
       "recommended": false,
       "type": "style",
-      "given": "$..parameters[*]",
-      "when": {
-        "field": "@key",
-        "pattern": "^(?!.*(\\$ref))"
-      },
+      "given": "$..[?(@path === \"$['parameters']\" || (@.parameters && @parentProperty !== 'links'))]..[?(@.in)]",
```
lag-of-death: @P0lip, I'm kinda pulling my hair here. I can't find a rule that would satisfy OAS2 and OAS3. I was trying to create a rule that would eliminate `parameters` under `links` from the equation. I also wanted to make sure that even if we have:
```yaml
links:
  abc: 
    parameters:
      param:
         in: body
```
then we don't require `.description`. I don't know, but I feel like we might need more rules or a custom function.
WDYT?

P0lip: Ideally, Spectral would support an array of paths, so that you can supply a few json path expressions.
We could implement it as a part of 4.2 or 5.0.
For now, we could simply have 2 separate rules, one for oas2, and the other for oas3.
Once an array of paths can be provided, we can merge the rules.
Thoughts?

philsturgeon: We can support "string or array" and keep it as non breaking. The next release is likely to be 5.0 but supporting both is just a good idea either way for the sake of existing users. No BC-breaks-for-the-sake-of-it here. :)

philsturgeon: @lag-of-death OAS2 doesn't have links :p

lag-of-death: thanks @P0lip, I'll try with more than one rule for this now. @philsturgeon, "OAS2 doesn't have links", that's correct. As I was saying in: https://github.com/stoplightio/spectral/pull/515/files/1a6bac8bc7ee4a3063b1e4d61f8c2b163993bb6a#r320012847 "I can't find a rule that would satisfy OAS2 and OAS3". So this was an attempt to have just one rule. Didn't work out

src/rulesets/oas3/index.json
```diff
@@ -36,7 +36,7 @@
     },
     "oas3-operation-security-defined": {
       "description": "Operation `security` values must match a scheme defined in the `components.securitySchemes` object.",
-      "recommended": true,
+      "recommended": false,
```
chris-miaskowski:
```suggestion
      "recommended": true,
```

src/rulesets/oas2/index.json
```diff
@@ -73,6 +73,19 @@
         "function": "truthy"
       }
     },
+    "oas2-parameter-description": {
+      "description": "Parameter objects should have a `description`.",
+      "recommended": false,
+      "type": "style",
+      "given": "$..[?(@parentProperty === 'parameters' && @.in)]",
```
chris-miaskowski: Over a call we tried to simplify it and use some sort of "union" but weren't able to find the right syntax.

src/rulesets/oas3/index.json
```diff
@@ -52,6 +52,19 @@
         "operation"
       ]
     },
+    "oas3-parameter-description": {
+      "description": "Parameter objects should have a `description`.",
+      "recommended": true,
+      "type": "style",
+      "given": "$..[?(@parentProperty !== 'links' && @.parameters)]['parameters']..[?(@.in)]",
```
chris-miaskowski:
```suggestion
      "given": "$..[?(@parentProperty !== 'links' && @.parameters)]['parameters'].[?(@.in)]",
```

Maybe check if single `dot` is enough - I think it should be.

test-harness/scenarios/parameter-description-parameters.oas2.scenario
```diff
@@ -0,0 +1,34 @@
+====test====
```
chris-miaskowski: Consider adding a positive test checking whether we don't return warnings if descriptions are provided.

src/rulesets/oas2/index.json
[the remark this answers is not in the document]
lag-of-death: @P0lip, what about `$.parameters`?

src/rulesets/oas3/index.json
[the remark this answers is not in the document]
lag-of-death: what about `$.components.parameters` ?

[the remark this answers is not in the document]
P0lip: Urgh, forgot about that. Yeah, you are right, let's leave it as is.

lag-of-death:
@P0lip, can you approve it?

Issue #565, opened 2019-09-18
`info` and `hint` severity level report shows as `white` instead of the appropriate name in pretty-print CLI


jerzyn:
**Describe the bug**
Rules with `info` or `hint` severity level shows as `white` instead of their name in the pretty print CLI display

**To Reproduce**

1. Apply a ruleset with a rule of severity `info` or `hint`
2. Run the CLI against a JSON/YAML file which would match those rules
3. See the screen

**Expected behavior**
Instead of displaying `white` display `info` or `hint`, depending on the severity level.

**Screenshots**
![image](https://user-images.githubusercontent.com/1447151/65153742-8ff21780-da2a-11e9-8cf2-002e8e91cd08.png)

**Environment (remove any that are not applicable):**
 - Library version: 4.1.1
 - OS: Ubuntu 18.04.3 LTS
 - Browser: command line

PR #569, opened 2019-09-18
fix(stylish): support hint and print proper severity

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/565

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**

After:
image
image
image
image


src/formatters/__tests__/__fixtures__/oas3-schema-errors.json
```diff
@@ -1,7 +1,6 @@
 [
   {
     "code": "oas3-schema",
-    "summary": "Validate structure of OpenAPIv3 specification.",
```
marbemac: Why did this go away? Was summary deprecated, and this PR is removing it?

test-harness/scenarios/results-skip-rule.oas3.scenario
```diff
@@ -15,4 +15,4 @@ OpenAPI 3.x detected
  1:1  warning  api-servers       OpenAPI `servers` must be present and non-empty array.
  2:6  warning  info-description  OpenAPI object info `description` must be present and non-empty string.
 
-✖ 2 problems (0 errors, 2 warnings, 0 infos)
\ No newline at end of file
+✖ 2 problems (0 errors, 2 warnings, 0 infos, 0 hints)
```
marbemac: Should we just tackle proper pluralization for the values in this message in this PR? Suggest either:

- https://github.com/prantlf/fast-plural-rules
- https://github.com/blakeembrey/pluralize

src/formatters/__tests__/__fixtures__/oas3-schema-errors.json
P0lip: >Why did this go away? Was summary deprecated, and this PR is removing it?

We removed `summary` some time ago. The fixture was just not updated at the time we did it.

test-harness/scenarios/results-skip-rule.oas3.scenario
P0lip: We already have a simple util for that.
```ts
function pluralize(word: string, count: number): string {
  return count === 1 ? word : `${word}s`;
}
```
I guess this should be quite enough for the simple case we deal with?

marbemac: Yeah that sounds good to me!

Issue #571, opened 2019-09-18
Create rule format for JSON schema files


rossmcdonald:
I believe there is only support today for OpenAPI formats. We should also allow users to run rules against JSON schema formats (or just generic objects, whichever is easiest).

The context here is that we need to allow Studio users to run rules on JSON schema files. I was thinking we could allow users to set a `format` attribute (`json-schema`, for example) on the rule to have it target JSON schema models.

philsturgeon:
You can lint any files not just OpenAPI, it could be asyncapi, agithub actuon yaml Cole, anything at all. Where did this request come from?

rossmcdonald:
@philsturgeon This is in the context of Spectral running within Studio. Isn't there a "format" designation set on rules based on which type of file they should be run against (ie, oas2, oas3)? Are there other formats that I'm just not aware of?

rossmcdonald:
BTW I've also opened https://github.com/stoplightio/studio/issues/122 since I didn't know exactly where the change would be needed. If this is just in Studio then we can close this one. I wasn't sure if we needed to add support for a new file format type or not.

philsturgeon:
AHA! Ok, so stuff will run against all formats by default. We added formats for oas2 and oas3 only so people could run certain rules on a specific version, but if you run a ruleset against a JSON Schema file it’ll work. I’m not sure what studio does with this.

PR #575, opened 2019-09-19
feat: introduce json-schema format

P0lip:
Closes https://github.com/stoplightio/spectral/issues/571

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

<will add some>

### DIFF
--- a/src/cli/services/linter.ts
+++ b/src/cli/services/linter.ts
@@ -4,7 +4,7 @@ import { getLocationForJsonPath, parseWithPointers } from '@stoplight/yaml';
 
 import {
   isJSONSchema,
-  isJSONSchema2019_09,
+  isJSONSchemaDraft2019_09,
   isJSONSchemaDraft4,
   isJSONSchemaDraft6,
   isJSONSchemaDraft7,
@@ -40,7 +40,7 @@ const KNOWN_FORMATS: Array<[string, FormatLookup, string]> = [
   ['json-schema-draft4', isJSONSchemaDraft4, 'JSON Schema Draft 4 detected'],
   ['json-schema-draft6', isJSONSchemaDraft6, 'JSON Schema Draft 6 detected'],
   ['json-schema-draft7', isJSONSchemaDraft7, 'JSON Schema Draft 7 detected'],
-  ['json-schema-2019-09', isJSONSchema2019_09, 'JSON Schema 2019-09 detected'],
+  ['json-schema-2019-09', isJSONSchemaDraft2019_09, 'JSON Schema Draft 2019-09 detected'],
 ];
 
 export async function lint(name: string, flags: ILintConfig, rulesetFile: Optional<string[]>) {
--- a/src/cli/services/linter.ts
+++ b/src/cli/services/linter.ts
@@ -2,13 +2,21 @@ import { isAbsolute, resolve } from '@stoplight/path';
 import { IParserResult, Optional } from '@stoplight/types';
 import { getLocationForJsonPath, parseWithPointers } from '@stoplight/yaml';
 
+import {
+  isJSONSchema,
+  isJSONSchemaDraft4,
+  isJSONSchemaDraft6,
+  isJSONSchemaDraft7,
+  isJSONSchemaLoose,
+  isOpenApiv2,
+  isOpenApiv3,
+} from '../../formats';
 import { readParsable } from '../../fs/reader';
 import { httpAndFileResolver } from '../../resolvers/http-and-file';
 import { getDefaultRulesetFile } from '../../rulesets/loader';
-import { isOpenApiv2, isOpenApiv3 } from '../../rulesets/lookups';
 import { readRuleset } from '../../rulesets/reader';
 import { Spectral } from '../../spectral';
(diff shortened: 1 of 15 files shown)


src/formats/__tests__/json-schema.spec.ts
```diff
@@ -0,0 +1,197 @@
+import {
+  isJSONSchema,
+  isJSONSchemaDraft4,
+  isJSONSchemaDraft6,
+  isJSONSchemaDraft7,
+  isJSONSchemaLoose,
```
philsturgeon: Can we add support for 2019-09? It came out yesterday.

P0lip: Sure, what's the URL?
` 'http://json-schema.org/draft-2019-09/schema#'`?

philsturgeon: http://json-schema.org/

marbemac:
Looks great, nice work. These lookup / format checkers are quite helpful from a library user perspective (and we could probably use them in studio) - should we export a portion of that stuff from the root package?

P0lip:
@marbemac 
Yeah, that's a valid point. Done.
Right now, I simply use full paths, like `@stoplight/spectral/dist/lookups/openapi`, but this will beak anytime we decide to move the file.

src/formats/json-schema.ts
```diff
@@ -0,0 +1,41 @@
+import { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema';
+import { isObject } from '../utils/isObject';
+
+type JSONSchema = JSONSchema4 | JSONSchema6 | JSONSchema7;
+
```
philsturgeon: Is any in the spec, or some shortcut/trick we use, or a mistake?

P0lip: @philsturgeon 
Yeah, it's specified in the following doc https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.1
The newer doc https://tools.ietf.org/html/draft-zyp-json-schema-04#section-3.5 does not have it specified though.

philsturgeon: @P0lip i think you can safely ignore anything in draft 3, that's 9 years old and draft 4 is the oldest thing we want to pay attention to as its the basis of OpenAPI 2, or draft 5 because OpenAPI 3. Folks are either using OpenAPI, or they're using modern JSON Schema (6 or 7). If they are on 3 then... erf? 

It doesn't hurt to have this in here, but lets not consider draft 3 for any other code we work on.

src/cli/services/linter.ts
```diff
@@ -23,6 +32,17 @@ export async function loadRulesets(cwd: string, rulesetFiles: string[]): Promise
...
 
+const KNOWN_FORMATS: Array<[string, FormatLookup, string]> = [
+  ['oas2', isOpenApiv2, 'OpenAPI 2.0 (Swagger) detected'],
+  ['oas3', isOpenApiv3, 'OpenAPI 3.x detected'],
+  ['json-schema', isJSONSchema, 'JSON Schema detected'],
+  ['json-schema-loose', isJSONSchemaLoose, 'JSON Schema (loose) detected'],
+  ['json-schema-draft4', isJSONSchemaDraft4, 'JSON Schema Draft 4 detected'],
+  ['json-schema-draft6', isJSONSchemaDraft6, 'JSON Schema Draft 6 detected'],
+  ['json-schema-draft7', isJSONSchemaDraft7, 'JSON Schema Draft 7 detected'],
+  ['json-schema-2019-09', isJSONSchema2019_09, 'JSON Schema 2019-09 detected'],
```
philsturgeon:
```suggestion
  ['json-schema-2019-09', isJSONSchema2019_09, 'JSON Schema Draft 2019-09 detected'],
```

P0lip: Is it a draft actually? I had `Draft` stated, but noticed they don't use 'Draft' word anymore.

src/formats/__tests__/json-schema.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+        'http://json-schema.org/draft-04/hyper-schema#',
+        'http://json-schema.org/draft-06/schema#',
+        'http://json-schema.org/draft-06/hyper-schema#',
+      ])('recognizes %s schema correctly', $schema => {
+        expect(isJSONSchemaLoose({ $schema })).toBe(true);
+      });
+    });
+
+    describe('by type', () => {
+      it.each(['any', 'array', 'boolean', 'integer', 'null', 'number', 'object', 'string'])(
```
philsturgeon:
```suggestion
      it.each(['array', 'boolean', 'integer', 'null', 'number', 'object', 'string'])(
```

actually I changed my mind let's not talk about any its gonna confuse things down the line

P0lip: Sure, makes sense.

src/cli/services/linter.ts
philsturgeon: It is, only the versioning scheme changed, it is still a draft.

Issue #573, opened 2019-09-19
add support for JSON Schema Draft-06, Draft-07 and 2019-09 in schema function (rulesets)


jerzyn:
**User story.**
As a user I want to reference JSON Schema Draft-06, Draft-07 or 2019-09 in schema function in my rulesets.

**Is your feature request related to a problem?**
I want to have a possibility to use newest JSON Schema drafts for schema validation in my custom rulesets. My data models are defined using draft-07 currently and they are not being loaded.

**Describe the solution you'd like**
I would like to add external referenced JSON Schema files following the latest drafts

**Additional context**
Custom Rulesets

philsturgeon:
@jerzyn hey, im not sure what you mean by "not being loaded"? Do you have some example files we can play with and some expected behavior vs actual behavior? 

AFAIK Spectral doesn't care about the JSON Schema draft version at all, it only cares that some sort of JSON/YAML file is being loaded and it will follow _any_ `$ref` it finds along the way (including some places maybe it should not).

If you are using JSON Schema in the context of OpenAPI then that's a different thing. OpenAPI v3.0 does not support drafts of JSON Schema newer than its own "extended subset" of draft 5 and there is nothing we can do to change that other than hope my PR (https://github.com/OAI/OpenAPI-Specification/pull/1977) goes through. Chatting to the OpenAPI TSC about that in 2 hours.

jerzyn:
i was chatting about this with @P0lip - maybe check with him. As for files - using this:  https://supermodel.io/adidas/api/HAL (not as a reference to this file but this exact code together with `$schema`) was not loading. Deleting `$schema` or changing it to draft-04 was fixing the issue. The ruleset was:
```  response-success-hal:
    description: All success responses MUST be of media type `application/hal+json`
    severity: error
    given: $.paths..responses[?( @property >= 200 && @property < 300 )]
    recommended: true
    severity: error
    type: "style"
    message: "Response documents MUST follow application/hal+json: {{error}}"
    then:
      field: schema
      function: schema
      functionOptions:
        schema:
          $ref: ./hal.yaml```

PR #576, opened 2019-09-20
fix(schema): support draft 6 and 7

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/573

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Contains 2 unrelated fixes (see commit messages).

Regarding ajv-better-errors and the exception that's thrown - I'll try to provide a fix for the package.

This is where it fails:
https://github.com/atlassian/better-ajv-errors/blob/126ee2e45ee829e69e3b2f3f6d56eb49d5c42135/src/validation-errors/enum.js#L52
https://github.com/janl/node-jsonpointer/blob/master/jsonpointer.js#L60
Basically, jsonpointer does not expect non-object value.

### DIFF
--- a/test-harness/scenarios/external-schemas-ruleset.scenario
+++ b/test-harness/scenarios/external-schemas-ruleset.scenario
@@ -1,0 +1,17 @@
+====test====
+Schemas referenced via $refs are resolved and used
+====document====
+openapi: 3.0.2
+info:
+  title: foo
+  description: baz
+====command====
+lint {document} --ruleset ./test-harness/scenarios/rulesets/external-schemas-ruleset.yaml
+====stdout====
+OpenAPI 3.x detected
+
+{document}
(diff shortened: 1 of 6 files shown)


test-harness/scenarios/external-schemas-ruleset.scenario
```diff
@@ -0,0 +1,17 @@
...
+info:
+  title: foo
+  description: baz
+====command====
+lint {document} --ruleset ./test-harness/scenarios/rulesets/external-schemas-ruleset.yaml
+====stdout====
+OpenAPI 3.x detected
+
+{document}
+ 3:10  error  info-title        should be equal to one of the allowed values
```
philsturgeon: Could you make a feature request for adding enum values to the message? This message looks funny.

PR #580, opened 2019-09-20
fix(request): proxy-agent breaks Studio

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

This is almost impossible to cover with tests, as Studio has quite a specific setup. We just gotta be careful and avoid running Node specific code as much as possible.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,16 +1,7 @@
 import fetch, { RequestInit } from 'node-fetch';
 
+export const DEFAULT_REQUEST_OPTIONS: RequestInit = {};
+
 export default async (uri: string, opts: RequestInit = {}) => {
-  let options = opts;
-
-  if (typeof process === 'object' && process.env.PROXY) {
-    try {
-      const ProxyAgent = await import('proxy-agent');
-      options = Object.assign(opts, { agent: new ProxyAgent(process.env.PROXY) });
-    } catch (ex) {
-      console.error(ex.message);
-    }
-  }
-
-  return fetch(uri, options);
+  return fetch(uri, { ...opts, ...DEFAULT_REQUEST_OPTIONS });
 };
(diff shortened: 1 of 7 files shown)


src/request.ts
```diff
@@ -1,11 +1,16 @@
...
 
-const fetch = require('node-fetch');
+export default async (uri: string, opts: RequestInit = {}) => {
+  let options = opts;
 
-export default (uri: string, opts: RequestInit = {}) => {
-  const options =
-    typeof process === 'object' && process.env.PROXY ? { ...opts, agent: new ProxyAgent(process.env.PROXY) } : opts;
+  if (typeof process === 'object' && process.env.PROXY) {
+    try {
```
philsturgeon: Is this going to import every time a request is made? Is that gonna be ok for performance/memory usage?

P0lip: Good point!
Luckily, Node.js [caches](https://nodejs.org/api/modules.html#modules_caching) modules (can be checked under [require.cache](https://nodejs.org/api/modules.html#modules_require_cache)), therefore they are loaded and executed only once unless one clears the cache.
That said, we should be okay.

Note, even though `import` is used here, it's replaced with a `require` call when we compile the code from TS to JS.

I don't take bundlers into account, since we don't aim to run this code in a browser, but they are no different.

test-harness/scenarios/proxy-agent.scenario
```diff
@@ -0,0 +1,15 @@
...
+foo:
+  $ref: http://localhost:3002/foo.json#/ref
+====env====
+PROXY=http://localhost:3001
+====command====
+lint {document}
+====stdout====
+
+{document}
+ 2:9  error  invalid-ref  FetchError: request to http://localhost:3002/foo.json failed, reason: connect ECONNREFUSED 127.0.0.1:3001
```
P0lip: This is a tad dumb. We could spin up a server.
Should be fine for now.

PR #604, opened 2019-09-27
Adds OS check to import of os-dependent binary in test-harness

AndrewBell:
**Checklist**

- [N/A] Tests added / updated
- [N/A] Docs added / updated

- [ ] Yes
- [X] No

**Additional context**

Testing some new scenarios locally, I ran across this bug where the linux binary is hard coded in the test-harness. This adds an OS check via `process.os()` [docs](https://nodejs.org/api/os.html#os_os_platform), which should resolve as "darwin" on OSX, and fallback to linux otherwise.

I didn't see a Windows binary being created, so I believe this should cover both scenarios. Linting passed, but unsure of if/how to test. Please let me know if there is anything else I can do to help here, and thank you for taking a look!

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "node": ">=8.3.0"
   },
   "scripts": {
-    "build.binary": "pkg . --targets linux,macos --out-path ./binaries",
+    "build.binary": "pkg . --output ./binaries/spectral",
     "build.oas-functions": "rollup -c",
     "build": "tsc -p ./tsconfig.build.json",
     "cli": "node -r ts-node/register -r tsconfig-paths/register src/cli/index.ts",
--- a/test-harness/index.ts
+++ b/test-harness/index.ts
@@ -10,7 +10,7 @@ import { parseScenarioFile } from './helpers';
 
 const glob = globFs({ gitignore: true });
 
-const spectralBin = path.join(__dirname, `../binaries/spectral-${os.platform() === 'darwin' ? 'macos' : 'linux'}`);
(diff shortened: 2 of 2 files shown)


AndrewBell:
I'm not sure the CI failures are related to my changes, but let me know if that's not the case.

P0lip:
@AndrewBell 
> I'm not sure the CI failures are related to my changes, but let me know if that's not the case.

They are not - everything is okay on the CI side in terms of tests.
We actually have a slightly invalid setup - any PRs coming from engineers outside Stoplight organization fail on the very last step, that is `cc-after`.

AndrewBell:
Thanks for the info on the CI. Here's the error I get trying to run the linux binary: `/bin/sh: /Users/user/workspace/spectral/binaries/spectral-linux: cannot execute binary file`, but wasn't sure how to investigate from there.

AndrewBell:
Let me take a crack at this later today. I'll let you know if I get stuck.

AndrewBell:
That definitely looks better, provided no one wants to build a binary for a system they're not currently using. I see that `.circleci/config.yml` calls this as well, but I assume it should be fine.

`pkg`, when not supplied a `--targets` parameter, [defaults to the hosts' OS.](https://www.npmjs.com/package/pkg#targets). For example, here's a snippet from my local build logs from `yarn build.binary` 

```
$ pkg . --output ./binaries/spectral
> pkg@4.4.0
  node10-macos-x64
✨  Done in 28.33s.
```

Thanks for the nudge in the right direction, and let me know if there's anything else I can improve.

PR #605, opened 2019-09-27
Add in an explicit example using functionOptions in custom functions

kylesykes:
Adds in another example of a `.js` file that implements optional arguments in a custom function.  The example in the docs also didn't really jive with me so I felt writing one that generalized the original example (checking equality against a string) was a good natural progression.

I ran into this issue, and someone on the community forum did as well, so I figured it was worth enhancing the docs a little bit to try and help future devs, especially given custom functions are so new.
https://community.stoplight.io/t/custom-functions-for-spectral-running-locally/877

I am not a JS developer, so please let me know if something looks off or broken.  

Thoughts?

**Checklist**

- [ ] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


philsturgeon:
This is great, thank you Kyle!

PR #606, opened 2019-09-27
In test-harness scenarios that expect a stdout, always check against actual stdout

AndrewBell:
**Checklist**

- [N/A] Tests added / updated
- [N/A] Docs added / updated

- [ ] Yes
- [X] No

**Additional context**

As documented and implemented, it is possible to receive false negatives in test scenarios where results of the `command` return no response (particularly when using grep). Consider the following scenario from the [example](https://github.com/stoplightio/spectral/tree/develop/test-harness#a-real-example)

```
====test====
Hello
====document====
openapi: 3.0.2
paths:
  /todos:
    get:
      responses:
        200:
          description: Get Todo Items
          content:
            'application/json':
              example: hello
====command====
lint {document} | grep "something"
====stdout====
 2:6  warning  some-rule  This rule is complaining about something.
```

One would expect this scenario as-is to fail, as there is no `some-rule` defined. Since we are `grep`ing for a particular error that is never generated, we receive a falsy `stdout`, which means we never resolve the `expect(stdout).toEqual(expectedStdout);` This change adds `expectedStdout` to the conditional, so that if a scenario defines an expected stdout, the harness will _always_ run the expect.

Alternatively, the project could advise scenario implementers to use the full expected error list [example](https://github.com/stoplightio/spectral/blob/develop/test-harness/scenarios/results-default-output.oas3.scenario) and remove the advice to use `grep`. My gut tells me the `grep` approach would prevent false positives in scenarios where additional rules are added that now error on un-related test scenarios. e.g. we add our own rule that the `results-output.oas3.scenario` does not conform to in it's definition, nor capture in it's stdout expect, and now is failing due to unrelated changes.

Linting passed, but did not see any tests to implement. Please let me know if there is a better approach or anything else I can do to help!

### DIFF
--- a/test-harness/README.md
+++ b/test-harness/README.md
@@ -40,18 +40,32 @@ lint --foo {document} --bar
 expected output
 ====stderr====
 optional/alternative to stdout for expected errors
+====status====
+(optional) expected exit code (`0` or `1`)
 ```
 
 ### A real example?
 
 ```
 ====test====
-Hello
+scenario-example
 ====document====
 openapi: 3.0.2
+servers:
+  - url: "localhost"
+info:
+  title: "my api"
+  description: "An example"
+  contact:
+    name: "Stoplight"
+  version: "1.0"
 paths:
   /todos:
     get:
+      tags:
+        - "example"
+      description: "Example endpoint"
+      operationId: getTodos
       responses:
         200:
           description: Get Todo Items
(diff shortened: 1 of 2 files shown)


AndrewBell:
I don't believe the CI failures are related to my changes, but let me know if that's not the case!

XVincentX:
The failures are not your fault, we have a CodeClimate step that we're not using whose secrets are not being injected into your build because it's a fork. We'll figure it out.

philsturgeon:
@XVincentX grep is great, let's chat about that elsewhere as this is a legit fix.

philsturgeon:
Oh ok @AndrewBell had questions. 

Yes, it is intentional. grep is not always recommended, it is just a supported option. I am trying to make it clear that people can do whatever shell commands they want here. 

Grep is intended and something I am a little sad Prism does not have in its test harness. The Prism test harness assumes that the command is always curl so the output is always a HTTP message, but the CLI acceptance testing can be far more generic than that. We want to let folks know they can pipe things, cat things, whatever. This is a feature not a bug. 

But yes, people can be too specific with this, much like you can be too specific with expect.objectContains({ foo }) and miss that bar was mentioned.

XVincentX:
I'm fine with it — just pointing out my thinking. Feel free to merge!

AndrewBell:
I'd be happy to help clarify the docs for Spectral. 

* I can't think of  a use-case that uses 'curl' for the Spectral test-harness, so we can remove references to curl in the `test-harness/README.md`
* Either remove `grep` from the example or remove the warning about it. Or, document as both the full error and `grep`ed with a note similar to @XVincentX 
 2nd bullet point https://github.com/stoplightio/spectral/pull/606#pullrequestreview-296378758 

Thoughts?

philsturgeon:
Curl is a layover from when we copied this from prism so yes please! I’m mobile but sounds like have good ideas, I trust your intuition on what to do about grep.

--
Phil Sturgeon
@philsturgeon

AndrewBell:
Made some updates to the README for clarity 👍

Issue #585, opened 2019-09-23
Rule: remove when


P0lip:
**Chore summary**
Remove `when` member from `IRule` interface.
```ts
  when?: {
    // the `path.to.prop` to field, or special `@key` value to target keys for matched `given` object
    // EXAMPLE: if the target object is an oas object and given = `$..responses[*]`, then `@key` would be the response code (200, 400, etc)
    field: string;

    // a regex pattern
    pattern?: string;
  };
```

It's used only by `parameter-description` rule and does not really seem to be needed, as a jsonpath expression could be used instead.
Moreover, we don't really have that property documented anywhere, therefore the impact is likely to be low.

P0lip:
@philsturgeon @marbemac do you have anything against?
Obviously, we don't need to remove it right now, but this would be a good candidate for 5.0.

philsturgeon:
Sure!

Amachua:
Hello, I would like to get your opinion/inputs on a rule I want to create (and on which I thought the `when` field could have been used) (even if I'm not sure how to do it properly for the moment).

When I've a request with a path parameter such as those in the attach file, then I would like to have an error that tells me that I've forget the 404 response for the requests:
- GET   /other_resources/{other_resource_id} on line 33
- PATCH   /other_resources/{other_resource_id} on line 39
- DELETE /other_resources/{other_resource_id} on line 48
- PUT /other_resources/{other_resource_id}/sub_resources on line 49

For the moment, I don't know how this can be performed and I imagine that the `when` field could have been used for such purpose. (if you do know how to do it properly, I'll be happy to have your opinion on this).

[404_responses.txt](https://github.com/stoplightio/spectral/files/3679907/404_responses.txt)

Amachua:
I've continued my research on this point and I think that I can find a work around using JSONPath expression. Only thing is... I don't think the tool you're using to parse a JSON can do what I have in mind. 

Before going to the point, I'd like to share with you the behavior I'm working in case this could also help in the building of the Spectral project.

### Ensure that the path parameters follow a pattern

The first rule I've created was to ensure that all the path parameters do follow a specific naming convention (often used with resource_id). This latter was written this way:

```
  path-parameter-names-snake-case:
    type: style
    severity: error
    recommended: true
    description: A parameter in the path should be written in snake_case and ends with _id
    message: "The path parameter must be snake_cased and finish with _id: {{error}}"
    given: $..parameters[?(@.in === 'path')]
    then:
      field: name
[...]

PR #614, opened 2019-09-30
feat(rule): remove 'when'

P0lip:
BREAKING CHANGE: 'when' is gone. You can write a jsonpath extension instead

Closes https://github.com/stoplightio/spectral/issues/585

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

~Don't know :thinking: 
@philsturgeon @marbemac do we treat it as breaking or not?
I think we made a change to rules (like removed a rule) in the past, and didn't mark it as breaking.~


philsturgeon:
@P0lip we have never documented this functionality so we could potentially release this as not a breaking change. If we want to play it cautious (Hyrum's Law is a thing) we can bundle this with another breaking change: clearing out the old TS cruf we've had floating around.

https://github.com/stoplightio/spectral/issues?q=is%3Aissue+is%3Aopen+label%3Abreaking

My recommendation would be to hold off on merging this, get v4.2.0 out in the next few days if we consider it ready, then consider develop (and therefore the next release) v5.0 so we can make these breaking changes and clear out a bunch of junk, along with this undocumented feature.

philsturgeon:
Seeing as we have a bunch of breaks coming in lets merge this, and #561 and all the other breaks and make a 5.0.

Issue #607, opened 2019-09-29
Junit formatter is not outputting according to XSD Schema


CasperJ:
**Describe the bug**
When using the JUnit formatter (`--format=junit --output=test.xml`) then the output XML is not aligned with the XSD Schema (https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd).

JUnit only supports "error" or "failure", not "warning", "hint" etc.

**To Reproduce**

1. Given an OpenAPI document that doesn't have an `operationId` should create a `warning`
2. Run this CLI command `lint myopapi.yaml --format=junit --output=test.xml`
3. See test.xml
```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
    <testsuite package="org.spectral" time="0" tests="2" errors="2" name="openapi.yaml">
        <testcase time="0" name="org.spectral.operation-description" classname="openapi">
            <warning message="Operation `description` must be present and non-empty string."><![CDATA[line 13, col 9, Warning - Operation `description` must be present and non-empty string. (operation-description)]]></warning>
        </testcase>
        <testcase time="0" name="org.spectral.operation-operationId" classname="openapi">
            <warning message="Operation should have an `operationId`."><![CDATA[line 13, col 9, Warning - Operation should have an `operationId`. (operation-operationId)]]></warning>
        </testcase>
    </testsuite>
</testsuites>
``` 

**Expected behavior**
 - testsuites/testsuite/testcase/warning is not a valid element. must be either "error" or "failure"
 - testsuites/testsuite@tests should not be equal "2" - ti should be the total number of tests run
 - if all tests succeeds then a file should still be outputted (which should include the number of tests passed)

**Additional context**
I would love to work on fixing this bug, but for now its here for reference ;)

P0lip:
Hey @CasperJ.
Thanks for raising the issue!
As you can see in the [PR](https://github.com/stoplightio/spectral/pull/519?cf_lbyyhhwhyjj5l3rs65cb3w=t77k9m2uwbiovv34a44xh#discussion_r322863637) implementing JUnit formatter, we initially generated `error` tag for each messages regardless of their severity.
Unfortunately, doing so didn't allow us to indicate the actual severity, so decided to make that change that lead to error/warning/information/hint tags being output.

I'm more than happy to adjust it to be more spec-compliant, but would still like to preserve the severity - which is pretty important in case of Spectral.
Is there any attribute we could make use of to retain the severity in the final output?

[...]

Issue #389, opened 2019-07-16
HTML Format


philsturgeon:
**User story.**
As a CLI user, I want to run the CLI, output information about failures, and put that HTML on S3 so folks can see a report of their errors and warnings. Maybe this will help them fix the various issues.

**Describe the solution you'd like**

Rubocop have a good HTML format: https://www.rubocop.org/en/stable/formatters/

Check out ESLint too.

PR #621, opened 2019-10-02
fix(junit): make formatter more spec-compliant

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/607

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Let's double-check whether everything is alright this time around.

### DIFF
--- a/src/formatters/__tests__/junit.test.ts
+++ b/src/formatters/__tests__/junit.test.ts
@@ -20,7 +20,8 @@ describe('JUnit formatter', () => {
         testsuite: [
           {
             $: {
-              errors: '3',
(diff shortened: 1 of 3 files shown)


philsturgeon:
Ok this is good but we lost the path, which will undo the done-ness of that previous issue. Seeing as JUnit does not care about Path, what can we recommend as a feature to follow in the mean time? #389 ?

P0lip:
@philsturgeon we didn't lose path. Path is now inlined and stands next to the location and message.

PR #642, opened 2019-10-07
Check markdown links

XVincentX:


XVincentX:
Can you do the same for Prism?

PR #651, opened 2019-10-08
build: create macOS binary

XVincentX:
Mac binary aren't getting generated because of the automatic command taking in consideration only the current machine; we need them for MacOs as well and so we'll use a custom one.


philsturgeon:
Test harness is all `    /bin/sh: 1: /home/circleci/project/binaries/spectral: not found` 

Please be careful I made this change before and missed out something in the release process which meant we had to do a follow up release. Can't remember what.

PR #657, opened 2019-10-09
feat: merge openapi-tags and openapi-tags-alphabetical

philsturgeon:
Two very similar rules, both were not recommended but people were not aware they needed to enable tags for Stoplight Docs. Having tags is just generally a great idea and something I've always recommended in general, so this seems like a reasonable addition. Now that studio and custom rulesets are more in the forefront, I figure folks can always turn this off if they hate it. It is also now recommended, because not recommended things are basically completely hidden from most users.

**Checklist**

- [ ] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

The develop branch is probably going to be v5.0 now, but lets be sure before merging. If unsure, we can make a `5.0` branch where we compile breaking changes until we are ready, then merge that to develop.


docs/reference/openapi-rules.md
```diff
@@ -246,11 +246,22 @@ info:
   title: 'some title with <script>alert("You are Hacked");</script>',
 ```
 
-### openapi-tags-alphabetical
+### openapi-tags
 
-OpenAPI object should have alphabetical `tags`. This will be sorted by the `name` property.
+OpenAPI object should have non-empty `tags` array, and the tags names should be ordered alphabetically.
```
P0lip: I'm absolutely fine with making them recommended by default, but what bothers me is that enforcing alphabetical order might not always be intended, at least looking from Studio perspective, since you might want to have X tag be displayed first in the UI.
Either way, if you all are fine with that, I won't oppose.
I just want to make sure we all aware of possible side effects of this change.
@philsturgeon, @marbemac, @lottamus, what do you think folks?

marbemac: We should not validate tags are alphabetically ordered by default IMHO. In the past, we've used the ordering of tags to control how things are ordered in docs.

philsturgeon: Interesting! Ok, I opened this PR as a conversation starter and was wondering if we had reasons for doing things this way. Intentionally out of order tags is not something I’d considered a feature but thanks for sharing!

P0lip: @philsturgeon I think we could still make `openapi-tags` rule recommended, right?
Sounds like a good default.

marbemac: Yeah I think there is still value in the `openapi-tags` rule being a recommended rule. The issue is that this PR changes it to do more than just check for presence of tags - it's also enforcing alphabetical order.

What about, as part of Spectral v5, we make `openapi-tags` recommended and add a new `openapi-tags-alphabetical` rule (that is not in the recommended set) that does what was added here: https://github.com/stoplightio/spectral/pull/657/files#diff-ce8b5719463f4eb8546e80d8f0a8e6d0R222-R228?

philsturgeon: @marbemac `openapi-tags-alphabetical` already exists and is not recommended already, so we can just switch `openapi-tags` to be recommended in develop and job done! Will do now.

PR #665, opened 2019-10-10
chore: drop support for Node 8

P0lip:
Node 8 is about to reach its end of life - https://nodejs.org/en/about/releases/
Users should not be really affected, as our code is likely to still be fully functional under Node 8, but we just won't support it officially anymore.


philsturgeon:
Normally I wouldn’t support removing tests on a version “for the sake of it” but if we are going to make develop v5 now then this is a good time to remove, otherwise we’d be forced to drop it as a v6 the first time a PR was struggling on v8.

PR #671, opened 2019-10-11
feat: static assets

P0lip:
Change driven by Studio.
~Will describe it in a great detail soon, once it's ready to review.~
Long story short - Studio will provide its own assets in the following manner
```ts
Spectral.registerStaticAssets(require('@stoplight/spectral/rulesets/assets/assets.json'));
```


scripts/generate-assets.js
```diff
@@ -0,0 +1,51 @@
...
+const baseDir = path.join(__dirname, '../rulesets/assets/');
+
+if (!fs.existsSync(baseDir)) {
+  fs.mkdirSync(baseDir);
+}
+
+const target = path.join(baseDir, `assets.json`);
+const assets = {};
+
+(async () => {
```
philsturgeon: Is this change needed if we are doing #561 ?

P0lip: Yeah, it's still going to be needed. It's mostly for Studio to use and we don't expose it anywhere in the docs, etc.
I'll leave a note somewhere in the code on what it actually does and how it works, as right now it's not obvious, and we will forget about why we need it.
I see Windows build is failing, so need to take a look what's broken.

collinbachi: We started a 2 week free trial Wednesday afternoon

philsturgeon: @P0lip ok, fair enough on the still needed. If you cannot get windows build working with Azure then maybe try out the CircleCI windows build, however that works.

P0lip: @philsturgeon I got Windows build working, but would love to try out Windows on CircleCI. Azure pipelines is tad slow.

philsturgeon: Another time! :)

PR #676, opened 2019-10-12
chore: minor JSONSchema typings tweak

P0lip:


P0lip:
I noticed `Check Markdown Links / markdown-link-check (pull_request) ` task fails pretty often.

philsturgeon:
That means there are broken links that we should fix!

XVincentX:
@P0lip Me too, and the error it is returning is kind of weird; I'll give it another week otherwise I'll probably remove it from Prism since it's slowing me down unfortunately :(

philsturgeon:
Great change 💯

Issue #695, opened 2019-10-18
Docs should be updated to remove deprecated functions


rossmcdonald:
> Method `addFunctions` has been deprecated since version 4.1

https://github.com/stoplightio/spectral/blob/develop/docs/guides/javascript.md#creating-a-custom-function

PR #711, opened 2019-10-28
chore: remove addRules and addFunctions

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/695

Addresses parts of https://github.com/stoplightio/spectral/issues/561

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No


docs/getting-started/rulesets.md
```diff
@@ -18,6 +18,7 @@ rules:
       function: truthy
 ```
 
+Spectral has a built-in set of functions which you can reference in your rules. This example uses the `truthy` to create a rule that checks that the value is true.
 The example above adds a single rule that checks that tags objects have a description property defined.
```
philsturgeon: "The example above" sounds a little funny now as we already say "This example". Can you merge the two sentences to make it less redundant?

P0lip: Yeah, you are right. It does sounds awkward. Will merge

Issue #447, opened 2019-08-13
Make recommended: true by default and fix inheritence


miller79:
_*Updated 2019-10-29:* Replaced with content from @marbemac's comment, which will be used to drive this issue._

recommended should be the default across the board - cli and ruleset extension. 

E.g. this:

```yaml
extends: spectral:oas3
```

Is the same as:

```yaml
extends: [[spectral:oas3, recommended]]
```

As far as extending w all, recommended, or none, it should bubble up the hierarchy (I thought that this was happening now, but maybe not?):

E.g. this:

```yaml
extends: [[foo.json, all]]
```

Is the same as:

```yaml
extends: [[bar.json, all], [baz.json, all]]
```

### Todo

- [x] change recommended to be true by default
- [ ] make sure all rules maintain their recommended status (no accidental changes)
- [ ] audit documentation to make sure there is no statements which disagree (for JS, CLI, rulesets in general, custom functions, etc)

P0lip:
@miller79 
Hello!
The reason `model-description` rule is not enabled is caused by the fact it's not a recommended rule, therefore, when you extend the built-in ruleset it's not enabled by default.

In order to have that rule enabled you can do the following:
```yaml
extends: spectral:oas3
rules:
  operation-operationId: error
  model-description: true # you can also set a custom severity error if you'd like to
```

or, alternatively, if you'd like to enable all rules
```
```yaml
extends: [[spectral:oas3, all]]
```
Hope this helps, sorry for any confusion caused.

We are in the process of making CLI default usage of rulesets more consistent.

I added a test here to make sure we never regress https://github.com/stoplightio/spectral/pull/448

miller79:
Thanks I have read this document but did not see the line talking about recommended rules.  My two cents would be that I think it would be easier to read if that recommended rule is stated in the extending rules section and not the disable rules section.  After some trial, I believe I am able to continue and this can be closed unless you would like to capture my comments.

P0lip:
@philsturgeon thoughts on above?
I think that Anthony is in fact right about that and there is a place for improvement in terms of built-in ruleset extension.

philsturgeon:
@P0lip yeah I think our stuff around recommended is incomplete. 

1. Does the CLI by default run all or only recommended 
2. Does ruleset extension by default include all or only recommended 

If those two things are different, then we’ll need to be veeeery clear in the docs about what can be expected. I think this is the current case.

miller79:
[...]

PR #719, opened 2019-10-30
feat: rules are recommended by default

P0lip:
Closes https://github.com/stoplightio/spectral/issues/447

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

### DIFF
--- a/docs/getting-started/rulesets.md
+++ b/docs/getting-started/rulesets.md
@@ -20,6 +20,7 @@ rules:
 
 Spectral has a built-in set of functions such as `truthy` or `pattern`, which you can reference in your rules.
 The example above adds a single rule that checks that tags objects have a description property defined.
+Each rule is recommended by default unless the value of recommended is explicitly to to false. 
 
 Running `spectral lint` on the following object with the ruleset above will result in an error being reported, since the tag does not have a description:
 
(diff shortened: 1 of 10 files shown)


docs/getting-started/rulesets.md
```diff
@@ -20,6 +20,7 @@ rules:
 
 Spectral has a built-in set of functions such as `truthy` or `pattern`, which you can reference in your rules.
 The example above adds a single rule that checks that tags objects have a description property defined.
+Each rule is recommended by default unless the value of recommended is explicitly to to false.
```
```suggestion
Since v5.0, each rule is recommended by default. Prior to that the default was to be not recommended. It is best to be explicit and set `recommended: true` or `recommended: false` if your ruleset is likely to be used across multiple versions.
```

src/rulesets/mergers/__tests__/rules.jest.test.ts
```diff
@@ -211,12 +214,49 @@ describe('Ruleset rules merging', () => {
...
       },
       'recommended',
     );
 
     expect(rules).toHaveProperty('test.severity', DiagnosticSeverity.Warning);
     expect(rules).toHaveProperty('test2.severity', -1);
+    expect(rules).toHaveProperty('test3.severity', DiagnosticSeverity.Warning);
+  });
+
+  it('rules with no severity and no recommended set are treated as warnings', () => {
```
philsturgeon: This message makes it sound like there is a crossover between severity and recommended, and im not sure there is?

P0lip: There is some. Basically, if there is no severity set, we either set a default value or disable it. Whether we set a default value or set it to off depends on the value of recommended.

In code, it covers this line https://github.com/stoplightio/spectral/pull/719/files#diff-7ba880b07a41aa1bf9a31d0a3bca8445R116

PR #723, opened 2019-10-31
fix: inheritance of rule with implicit severity and a few extra tests covering inheritance

P0lip:
Related to https://github.com/stoplightio/spectral/issues/447

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**
- [ ] help me with naming :sweat_smile:


philsturgeon:
Is there any way we can test a bit more directly? Instead of testing the results are trimmed in a likely way, can we test to see if it is passing a "off" or a "recommended" to some function? Basically make this a bit more of a unit test than an integration test? It's tough to look at this and see exactly what is going on.

P0lip:
> Is there any way we can test a bit more directly? Instead of testing the results are trimmed in a likely way, can we test to see if it is passing a "off" or a "recommended" to some function?

Yeah, you are right. I find them quite difficult to follow as well when I update them in future, so I'm eager to add a few unit tests.
That said, I'd still like to keep these integration tests, as they give me more confidence.
Are you okay with that?

philsturgeon:
Sure, both is good.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

P0lip:
@philsturgeon was writing a few unit tests, and spotted a case we don't support well
- contents of `foo.json` (rule without any explicit severity)
```json
{
  "rules": {
    "optional-rule": {
      "message": "foo",
      "given": "",
      "then": {
        "function": ""
      }
    }
}
```
- ruleset extending foo.json
```json
{
  "extends": [["./foo.json", "off"]],
  "rules": {
    "optional-rule": true
  }
}
```
Gonna fix this minor issue in this PR.
Initially I thought my unit test is just broken, but turned out I missed this one.

src/rulesets/mergers/__tests__/rules.jest.test.ts
```diff
@@ -146,18 +118,6 @@ describe('Ruleset rules merging', () => {
     expect(rules).toHaveProperty('test.severity', -1);
   });
 
-  it('supports array-ish syntax', () => {
```
P0lip: This was one was a dupe. See https://github.com/stoplightio/spectral/blob/6661202220798ce5139c16bcf887fbdf4721730f/src/rulesets/mergers/__tests__/rules.jest.test.ts#L109

P0lip:
@marbemac @philsturgeon alright fellas. I won't come with any more tests, I believe.
I added some unit tests for rules mergers we didn't have, but besides these, we already have quite a few tests, therefore I cannot really think of anything else I'd like to cover.
If you have any ideas, I'm all ears.

PR #758, opened 2019-11-13
feat(cli): accept stdin

### LINKED ISSUE #757: Accept input via stdin
Accept stdin as an alternative way to get a document, so that folks can use spectral in the middle of complex workflows without needing to write temporary files. 

```bash
cat openapi.yml | spectral -r some-ruleset.yml
```

This was mention by @radicarl in https://github.com/stoplightio/spectral/issues/656#issuecomment-550960358 but has been mentioned a few times in person and elsewhere. Just kept forgetting to write it down.

P0lip:
Closes #757

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

~Added isTTY env variable.
https://github.com/stoplightio/spectral/tree/chore/test-harness work has been pushed here. Still need to handle tty correctly, but it's getting closed.~

### DIFF
--- a/src/cli/commands/lint.ts
+++ b/src/cli/commands/lint.ts
@@ -1,7 +1,8 @@
 import { Dictionary } from '@stoplight/types';
+import { pick } from 'lodash';
+import { ReadStream } from 'tty';
 import { CommandModule, showHelp } from 'yargs';
 
-import { pick } from 'lodash';
 import { getDiagnosticSeverity } from '../../rulesets/severity';
 import { IRuleResult } from '../../types';
 import { FailSeverity, ILintConfig, OutputFormat } from '../../types/config';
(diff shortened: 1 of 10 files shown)


src/cli/services/linter/utils/listFiles.ts
```diff
@@ -1,10 +1,12 @@
 import { normalize } from '@stoplight/path';
 import * as fg from 'fast-glob';
 
-export async function listFiles(pattens: string[]): Promise<string[]> {
-  const { files, urls } = pattens.reduce<{ files: string[]; urls: string[] }>(
+export async function listFiles(pattens: Array<number | string>): Promise<Array<number | string>> {
+  const { files, fds, urls } = pattens.reduce<{ files: string[]; urls: string[]; fds: number[] }>(
```
philsturgeon: Could we rename fds to fileDescriptors or something? I'm not sure what a file descriptor is but that's more readable than fds.

src/cli/commands/lint.ts
```diff
@@ -14,13 +15,25 @@ const formatOptions = Object.values(OutputFormat);
...
       .positional('documents', {
         description:
           'Location of JSON/YAML documents. Can be either a file, a glob or fetchable resource(s) on the web.',
         type: 'string',
+        coerce(values) {
+          if (values.length > 0) {
+            return values;
+          }
+
+          if (process.env.isTTY || process.stdin.isTTY) {
```
philsturgeon: Im not sure I understand what is happening here? What is TTY and why does it need to be true?

test-harness/scenarios/help-no-document.scenario
```diff
@@ -2,13 +2,15 @@
 Errors when no document is provided
 ====command====
 lint
+====env====
+isTTY=true
```
philsturgeon: Could we get a test harness test showing `cat somefile.yaml | spectral --something` to show that its working as expected?

src/cli/commands/lint.ts
P0lip: I added comments explaining that. LMK if it helps.

PR #766, opened 2019-11-14
chore: improve test-harness

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

>

**Notable changes**
- assets are supported
- cwd is set to test-harness/scenarios
- spectral is no longer prepended to command, `{spectral}` needs to provided as of now
- replacements are hopefully slightly easier to specify now


test-harness/index.ts
```diff
@@ -1,105 +1,72 @@
...
-        replacements.push({
-          from: /\{documentWithoutExt\}/g,
-          to: tmpFileHandle.name.replace(/\.yml$/, ''),
-        });
-
-        fs.writeFileSync(tmpFileHandle.name, scenario.document, { encoding: 'utf8' });
-      }
+    const replacements: Dictionary<string> = {
+      __dirname,
+      spectral: spectralBin,
```
philsturgeon: Keeping in mind the end goal is to open source this, should we maybe instead of "spectral" call it "bin" or something?

XVincentX:
I’ve tried to look at this PR briefly but I can’t understand what are the pain points and what solution have been employed to fix these.

I also see the new `assets` feature, but I do not see any documentation update.

In any case, I’ve removed myself as a reviewer since I do not think I am going to be able to jump on that any time soon — feel free to move forward without me.

philsturgeon:
@XVincentX this is so we can test stdin for spectral. `cat foo.yaml | spectral -r ruleset.yaml` will not work if the command is forced to be `spectral ${cmd}`.

test-harness/index.ts
P0lip: Yeah, makes sense, I'll do it.

P0lip:
> I’ve tried to look at this PR briefly but I can’t understand what are the pain points and what solution have been employed to fix these.

Initially I was planning to drop child_process#spawn entirely and use pseudo terminals (i.e. forkpty) using exists Node.js bindings such as [node-pty](https://github.com/microsoft/node-pty).
It turned out to be quite complex, so I punted this, as the potential benefits certainly did not outweigh the extra complexity in test-harness. The revert was performed in 283553e368e4178575c9e2405ceb1d82cfd77f7.
This would solve some pains (it'd be closer to real terminal, so you could actually type the command as the user does), but as stated, not really worth the effort.

Besides that we were missing assets, so whenever I wanted to test a ruleset, I had to create a physical files.
No other major changes, just some polish and improvements here and there

> I also see the new assets feature, but I do not see any documentation update.

True, I'll add some!

PR #794, opened 2019-11-20
build: binary install was broken for linux

P0lip:
Note - we won't this need change for develop, as develop already generates correct binaries.
https://github.com/stoplightio/spectral/blob/develop/.circleci/config.yml#L88

Theoretically I could upload spectral-linux and let it be, but I'd rather not touch it at all.


P0lip:
```
  node13-linux-x64
> Error! No available node version satisfies 'node13'
```
eh

PR #798, opened 2019-11-21
feat: report unrecognized formats

P0lip:
Closes https://github.com/stoplightio/spectral/issues/678

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


philsturgeon:
Slight concern here. If I am working on a Spectral ruleset in studio it will not have a format, so im going to see complaints about a lack of format all the time I work on it.

Any other .json file I work on in studio will also complain about a lack of format. 

Is this a good feature?

P0lip:
> Slight concern here. If I am working on a Spectral ruleset in studio it will not have a format, so im going to see complaints about a lack of format all the time I work on it.

Should not be a problem, since we don't use Spectral for linting Spectral ruleset.

> Any other .json file I work on in studio will also complain about a lack of format.

True, but similarly to above, I can disable linting of such documents.

Either way, I hear your concerns.
Perhaps we make it opt-in (flag?) and exclusive to CLI?

marbemac:
Suggest add a method to public API along the lines `supportsFormat(id: string): boolean` - Studio (or other implementations) could use this to check before running spectral on a file. Thoughts?

nulltoken:
@P0lip Wow. Thanks for making that happening!

philsturgeon:
@P0lip yeah this could either be a CLI default, or opt-in for the CLI, or opt-in everywhere, but I don't think it should be default eerywhere.

I mentioned rulesets because in Studio we do seem to lint rulesets, and I asked about random json files because I noticed it linting random json files 😓 

Having studio only applying rules it knows to formats it recognizes is a good feature, but I can understand the CLI users wanting an error or warning if nothing is applied, that's a rather different use case.

P0lip:
@philsturgeon @marbemac 
so, what's the final call? :)
opt-in flag for CLI? and an extra method for JS usage?

nulltoken:
I'd say that opt-in for cli would actually defeat the purpose of the feature (which was to warn the user that something went sideways).

cf #678 _"Indeed, I may think that my file is perfect and that it passed all the many built-in checks Spectral comes up with. Whereas the sad reality is that it went through NO rule at all."_

philsturgeon:
Agreed @nulltoken 

CLI can be default, but what about JS? Default on with a "ignoreUnknownFormat" option?

P0lip:
> CLI can be default, but what about JS? Default on with a "ignoreUnknownFormat" option?

Sounds good to me. :+1:
I'd also make it enabled by default.

Are we all set? I wanna wrap this one so badly before 5.0 release :wink:

philsturgeon:
image

P0lip:
@philsturgeon @marbemac it's ready for another pass
cc @nulltoken

src/spectral.ts
```diff
@@ -240,6 +243,20 @@ export class Spectral {
...
     return resolveOpts;
   };
+
+  private static _generateUnrecognizedFormatError(parsedResult: IParsedResult): IRuleResult {
+    return {
+      range: parsedResult.getLocationForJsonPath(parsedResult.parsed, [], true)?.range || {
+        start: { character: 0, line: 0 },
+        end: { character: 0, line: 0 },
+      },
+      message: 'Provided file format does not match any of the registered formats',
```
marbemac: Reads a little weird - the user is not providing a file format (to me reads like it's some format option I provided) - they're providing a document (file, url, etc) that has a format. Perhaps the following?

`The provided document does not match any of the registered formats [openapi, markdown]` - bonus if can get the registered format names in there to give end user a bit more context.

src/spectral.ts
```diff
@@ -240,6 +243,20 @@ export class Spectral {
...
+      code: 'unrecognized-format',
+      severity: DiagnosticSeverity.Information,
```
marbemac: Info or error? E.g. should probably exit with a 0 if this occurs, so that tests fail? For example if user removes the `swagger` property from their openapi file, it is definitely invalid, but Spectral will not identify it as openapi at all and thus run no rules.

P0lip: You can make still make Spectral fail if you use `fail-severity-level` flag.
In any case, I agree that warning feels like a more appropriate severity. Will change.

philsturgeon: Yeah let’s make it a warning. The CLI is already set to panic about this so no worries there.

src/__tests__/linter.test.ts
```diff
@@ -144,6 +142,8 @@ describe('linter', () => {
   });
 
   test('should not report anything for disabled rules', async () => {
+    spectral.registerFormat('oas2', isOpenApiv2);
```
philsturgeon: Can you add a context and best then inside it?

describe(“when os2 and oas3 are registered as formats”

Then all tests which need that go inside

src/__tests__/linter.test.ts
```diff
@@ -41,20 +41,18 @@ describe('linter', () => {
...
     const fakeLintingFunction = jest.fn();
     spectral.setFunctions({
       [fnName]: fakeLintingFunction,
     });
     spectral.setRules(rules);
 
-    const result = await spectral.run('123');
-
-    expect(result).toHaveLength(0);
+    return expect(spectral.run('123', { ignoreUnknownFormat: true })).resolves.toBeTruthy();
```
philsturgeon: Do we need to talk about ignoreUnknownFormat in here? Seems like it's more about strings vs objects? 

Maybe it could be two different tests, one to prove the default and another further down where we do more testing of ignoreUnknownFormat?

src/__tests__/linter.test.ts
```diff
@@ -418,6 +429,7 @@ describe('linter', () => {
 
   test('should not run any rule with defined formats if some formats are registered but document format could not be associated', async () => {
     spectral.registerFormat('foo-bar', obj => typeof obj === 'object' && obj !== null && 'foo-bar' in obj);
+    spectral.registerFormat('markdown', () => false);
```
philsturgeon: This doesnt seem to be doing much?

P0lip: Actually, no, you are right. I'll remove it

src/__tests__/linter.test.ts
P0lip: Yeah, good point.

nulltoken:
@P0lip Thanks!

PR #807, opened 2019-11-25
feat(cli): implement HTML formatter

### LINKED ISSUE #389: HTML Format

P0lip:
Closes #389 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**

localhost_63342_spectral_test html__ijt=d1nrv39kkdr5jm5tebd0jvn5s5


package.json
```diff
@@ -43,9 +43,10 @@
     "inline-version": "./scripts/inline-version.js",
     "lint.fix": "yarn lint --fix",
     "lint": "tsc --noEmit && tslint 'src/**/*.ts'",
+    "copy.html-templates": "copyfiles -u 1 \"./src/formatters/html/*.html\" \"./dist/\"",
```
philsturgeon: Is there any other way to go about this? Is this standard? It feels a little clunky to me, but maybe its fine?

P0lip: > Is there any other way to go about this? Is this standard?

What do you mean? We can place it directly in `prebuild`, no need to have this extra script. I just wanted to make is slightly more readable.

philsturgeon: I just meant... this feels like an "asset pipeline" sort of thing, I wonder if there are more common ways people handle "not typescript" in typescript libraries than copying bits over. Nevermind :D

P0lip:
@philsturgeon could you re-approve? Had to normalize line endings for Windows (it used CRLF instead of LF).

Issue #823, opened 2019-11-30
Add intellij formatter


memelet:
**User story.**
As an api developer, I can do configure spectral to emit output from which intellij can navigate to the api spec.

**Describe the solution you'd like**
A new formatter that emits messages that include filename:line:column.

**Additional context**
For example:
```
src/api.yml:320:21 <similar output to stylish>
```

(PR forthcoming)

memelet:
Of course if stylish just added the file name to the existing line:column instead of just emitting it at the the top of the errors that also would work for idea.

memelet:
It would be nice to get a relative path to the api spec...

Issue #822, opened 2019-11-30
Add teamcity inspection formatter


memelet:
**User story.**
As an SRE, I can do configure spectral to emit teamcity inspection service messages, so that I can fail the build and navigate to errors.

**Describe the solution you'd like**
A new formatter that emits teamcity service messages.

**Additional context**
Teamcity service messages look like
```
##teamcity[inspection typeId='valid-example-in-schemas' file='src/api.yml' line='3970' message='\"hour-min.example\" property pattern should match pattern \"^(?:(|[01|]\\d||2|[0-3|]):(|[0-5|]\\d))$\"']
```

See also https://www.jetbrains.com/help/teamcity/build-script-interaction-with-teamcity.html#BuildScriptInteractionwithTeamCity-ReportingInspections

(PR forthcoming)

PR #825, opened 2019-11-30
Add two formatters: text, teamcity.

memelet:
Closes https://github.com/stoplightio/spectral/issues/823 and https://github.com/stoplightio/spectral/issues/822.

The text formatter is close to the stylish but with no special formatting and a source:line:column suitable for navigation within an IDE console (IntelliJ specifically).

The teamcity formatter emits teamcity inspection service messages that teamcity will automatically detect and add to the build inspection tab.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


src/formatters/teamcity.ts
```diff
@@ -0,0 +1,53 @@
+import { Dictionary } from '@stoplight/types';
+import { IRuleResult } from '../types';
+import { Formatter } from './types';
+import { getSeverityName, groupBySource, sortResults } from './utils';
+
+function escapeString(str?: string | number) {
+  if (!str) {
```
P0lip: What about `0`?
Let's adjust the condition a bit, suggest the following
```suggestion
  if (str === void 0) {
    return '';
  }
```

src/formatters/teamcity.ts
```diff
@@ -0,0 +1,53 @@
+import { Dictionary } from '@stoplight/types';
```
P0lip:
```suggestion
import { Dictionary, Optional } from '@stoplight/types';
```

src/formatters/teamcity.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IRuleResult } from '../types';
+import { Formatter } from './types';
+import { getSeverityName, groupBySource, sortResults } from './utils';
+
+function escapeString(str?: string | number) {
```
P0lip:
```suggestion
function escapeString(str: Optional<string | number>) {
```
Let's make the argument required.

src/formatters/teamcity.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!str) {
+    return '';
+  }
+  return str
```
P0lip: Being a bit paranoid :sweat_smile: 
```suggestion
  return String(str)
```

src/formatters/teamcity.ts
```diff
@@ -0,0 +1,53 @@
...
+    .toString()
```
P0lip:
```suggestion
```

.gitignore
```diff
@@ -26,6 +26,7 @@
 .env.production.local
 .vscode
 .idea
+.nvmrc
```
P0lip: Actually, go ahead and commit it.
I'm a happy user of nvm as well :slightly_smiling_face:.

memelet: fyi: version is 10.15.3, and all tests pass with that

PR #828, opened 2019-12-02
fix: resolve the source of errors for top-level references

P0lip:
It's a cherry-pick of the fix originally provided in https://github.com/stoplightio/spectral/pull/685/commits/05e909dcac1216e6634aac176fc79d1ea58680e7. Since it's unrelated, let's merge it ASAP, while I'm addressing the comments regarding the actual part of that PR. Will also reduce the noise in that PR making it easier to review.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Minor fix for top-level $refs. Right now, the source of error would not always be detected correctly for such a document
```yaml
$ref: ./remote-document.yaml#
```
Rather unlikely to be an often case, but it's still good detect it correctly.

### DIFF
--- a/src/cli/services/__tests__/linter.test.ts
+++ b/src/cli/services/__tests__/linter.test.ts
@@ -708,14 +708,26 @@ describe('Linter service', () => {
 
   describe('--resolver', () => {
     it('uses provided resolver for $ref resolving', async () => {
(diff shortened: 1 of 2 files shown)


P0lip:
The approach has been changed entirely in https://github.com/stoplightio/spectral/pull/685

PR #852, opened 2019-12-12
chore: test-harness codegen

P0lip:
**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Need to polish one thing and should be ready to go granted you all want it :sweat_smile: 
aaand some tests would be welcome too, I believe.

https://drive.google.com/file/d/1hds_ujaidN2QU_kztsVvkdJGrivCt5Z2/view?usp=sharing


P0lip:
@philsturgeon @marbemac and maybe @XVincentX what do you all think?
Nothing urgent, so you can take a look at it later.

philsturgeon:
What is the motivation for this? Tech debt?

P0lip:
@philsturgeon nope, not tech-debt :D
There is a video attached which hopefully explains pretty much everything.
Basically, what I did here is a small test generator, so that you can actually generate Jest-complian test for each test harness scenario.

philsturgeon:
It seems like it'll make things slightly quicker, but im not sure its particularly slow enough to justify working on it.

Are there any other benefits to this, like drastically improved diffs or failure detection? 

In general we should only be working on changes, features, bugs, etc which have an issue attached. This keeps us on track and makes it clear what the motivations are, otherwise we can find ourselves working on other things which might not be such high priority. 

Seeing as this is already done it's a bit late to worry about an issue, but... any more work on getting this done (like fixing conflicts) should be weighed against actual benefits of having this change done.

XVincentX:
Ok I finally had some time to watch the video.

This is definitely pretty cool, but I have the impression that it's a little bit an overkill for our use case. It certainly brings some value to the table, but the costs seem too much.

I'm not preventing this from happening in Spectral — I just do not think I'd back port it in Prism.

P0lip:
> It seems like it'll make things slightly quicker, but im not sure its particularly slow enough to justify working on it.

It's not about slowness. I don't know whether I said that during the video or not, but this is not what drives the change.
It's targeting DX, which is imperfect, at least I find it far from ideal which obviously affects my productivity.
At the moment, it's pretty annoying to run selective tests. You can provide that FILES env variable, but this is not very intuitive, so you end up running the whole suite for the most of time.

> Are there any other benefits to this, like drastically improved diffs or failure detection?

The detection is slightly improved, but I wouldn't say it's significantly better.
The only notable different is these color (or whatever they are called) chars.
"Colors" is something we don't really test at the moment anywhere and I strongly believe we should, as this is a crucial part of the stylish output, our default CLI formatter. For the time being, we have no way to assure the binary prints strings annotated with these color chars, but I suppose we could improve the existing solution to support that.

> In general we should only be working on changes, features, bugs, etc which have an issue attached. This keeps us on track and makes it clear what the motivations are, otherwise we can find ourselves working on other things which might not be such high priority.

It's actually a hobby project I did in my spare time. The fact I opened it does not mean I actually coded it when working for Stoplight. I was just bored.
Whether it brings value or not, it's a whole different topic, but it does make my life easier, so well, it does bring value for me.
That being said, we can close the PR, I don't really mind, since I can still keep the code in my own private project.

> This is definitely pretty cool

Thanks

> but I have the impression that it's a little bit an overkill for our use case. It certainly brings some value to the table, but the costs seem too much.

Could be. I don't really know what your use case is, so I cannot really elaborate much.
I didn't really think of Prism using this, to be frank, so it might be the change wouldn't satisfy your needs at all.
I did it primarily because I was bored and wanted to improve something that annoys me at times.
It would make my life easier, so I see the benefit of using it, but if everything one else is against, I don't really oppose. Although I spend the most time on Spectral, it's not my private project, so I don't want to bloat it with stuff nobody else wants or finds useful.
As said, it's a hobby project, so I don't really care whether we merge it here or not.

> I'm not preventing this from happening in Spectral — I just do not think I'd back port it in Prism.

Yeah, that's absolutely okay. I didn't really think of sharing that code with Prism, as I agree this would introduce an extra burden you wouldn't necessarily want to maintain.


---
Having said that, I'll close the PR for now.
If you change your mind, I can re-open it, update existing test cases, add some tests for the codegen and merge it.

PR #861, opened 2019-12-18
build: disable auto-merge for dependabot

P0lip:
Dependabot takes care of the PRs, so there is no need for github-actions to merge the develop branch.
The history quite a mess due to that + Dependabot is broken too when such a merge commit is pushed (see https://github.com/stoplightio/spectral/pull/832#issuecomment-566085874 as one of the examples).
In general, I believe automerging PRs bumping external packages may be unsafe, i.e what if it contains malicious code? we should at least briefly examine what the changes are all about.

According to Dependabot, it can even merge the PRs for you, so we don't need that action even for an actual merging.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -19,6 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: automerge
+        if: github.actor != 'dependabot-preview[bot]' && github.actor != 'dependabot[bot]'
         uses: "pascalgn/automerge-action@v0.6.0"
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"


philsturgeon:
I was not expecting `automerge` and dependabot to conflict with each other. Have you been applying `automerge` to dependabot PRs? 

Context for folks who didnt see @P0lip and I talking on slack, Dependabot is _meant_ to auto merge itself without this addon, but their "Allow dependabot preview to merge" feature does not actually work right now. It's being fixed but still https://github.com/dependabot/feedback/issues/86 

I guess either way we dont want automerge and dependabot getting confused so lets do this.

P0lip:
> I was not expecting automerge and dependabot to conflict with each other. 

Neither did I!

> Have you been applying automerge to dependabot PRs?

The label gets automatically added. This PR changes that and hopefully the action is no longer triggered for PRs created by Dependabot.

PR #865, opened 2019-12-18
Updating docs to contain examples for loading custom rulesets in the …

### LINKED ISSUE #864: Javascript API does not support extending rulesets
**Describe the bug**
I have a ruleset that extends the oas2 ruleset. This ruleset works perfectly from the command line but when I try to port that into my code with the JS API I cannot extend the oas2 rules. ~In the CLI the `RulesetProcessor` it is checking for `ruleset.extends` and processing those in turn. However neither `spectral.loadRuleset()` or `spectral.setRuleset()` take into account extends. This is made worse by the fact that there is a `spectral.mergeRules()` function that could be used to merge in my own rules if I wanted to take all of the rules from oas2, but there isn't an equivalent for functions to be able to add my custom functions.~

**To Reproduce**

Try to lint a ruleset with an extends using the JS API. Here is the one I am using (without my custom functions) https://gist.github.com/Jaredk3nt/6cf890fd26a77fced3dd602f11607c0e

**Expected behavior**
~There should be some way to extend a ruleset, either automatically or manually.~

This was caused by me misreading the code and the docs not really having a good example. I was able to get my extended ruleset to work in the JS API by using loadRuleset:

```js
spectral.registerFormat("oas2", isOpenApiv2);

function validate(spec) {
  return spectral.loadRuleset(__dirname + "/my-rules.json").then(() => {
    return spectral
      .run({
        getLocationForJsonPath,
        parsed: parseWithPointers(spec)
      })
      .then(console.log);
  });
}
```

Hopefully this example will save someone else!

Jaredk3nt:
…JS API

Fixes #864 .

**Checklist**

- [ ] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

I struggled through using a custom ruleset with extensions in the JS API due to confusing docs examples, so I wanted to update the docs with what worked for me.


docs/guides/javascript.md
```diff
@@ -114,10 +114,45 @@ spectral.loadRuleset('spectral:oas')
...
   });
-``` 
+```
 
 The OpenAPI rules are opinionated. There might be some rules that you prefer to change, or disable. We encourage you to create your rules to fit your use case, and we welcome additions to the existing rulesets as well!
 
+Custom rulesets can also be loaded using `spectral.loadRuleset()` by specifying the exact path to the ruleset file.
+
+```js
+const { Spectral, isOpenApiv2, isOpenApiv3 } = require('@stoplight/spectral');
```
P0lip:
```suggestion
const { Spectral, isOpenApiv2, isOpenApiv3 } = require('@stoplight/spectral');
const { join } = require('path');
```

docs/guides/javascript.md
```diff
@@ -114,10 +114,45 @@ spectral.loadRuleset('spectral:oas')
...
+const myOpenApiDocument = `
+openapi: 3.0.0
+# here goes the rest of document
+`
+
+const spectral = new Spectral();
+spectral.registerFormat('oas2', isOpenApiv2);
+spectral.registerFormat('oas3', isOpenApiv3);
+
+spectral.loadRuleset(__dirname + '/path/to/my-ruleset.yaml')
```
P0lip:
```suggestion
spectral.loadRuleset(join(__dirname './path/to/my-ruleset.yaml'));
```

Jaredk3nt:
Sorry for the delay, just saw this, I committed the suggested changes, glad to be able to help!

PR #867, opened 2019-12-19
feat: consistent paths in messages

P0lip:
Addresses https://github.com/stoplightio/spectral/issues/815

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/src/__tests__/linter.test.ts
+++ b/src/__tests__/linter.test.ts
@@ -642,7 +642,7 @@ responses:: !!foo
       }),
       expect.objectContaining({
         code: 'oas3-valid-schema-example',
-        message: '"foo.example" property type should be number',
+        message: 'property type should be number',
         path: ['components', 'schemas', 'foo', 'example'],
       }),
     ]);
@@ -659,7 +659,7 @@ responses:: !!foo
       expect.arrayContaining([
         expect.objectContaining({
           code: 'oas2-valid-parameter-example',
-          message: '"schema.example" property can\'t resolve reference #/parameters/missing from id #',
+          message: "property can't resolve reference #/parameters/missing from id #",
           path: ['paths', '/todos/{todoId}', 'put', 'parameters', '1', 'schema', 'example'],
         }),
       ]),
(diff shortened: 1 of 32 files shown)


src/__tests__/linter.test.ts
```diff
@@ -642,7 +642,7 @@ responses:: !!foo
       }),
       expect.objectContaining({
         code: 'oas3-valid-schema-example',
-        message: '"foo.example" property type should be number',
+        message: 'property type should be number',
```
P0lip: @philsturgeon is this something you expected to see?
I believe there is no point in printing paths when it exists in the document.
The only reasonable case is to print them when the property **does not** exist in the document and you want to indicate that, i.e. "XYZ property is not truthy".

We could keep it configurable though and let users decide, but leaving the call up to you all.

cc @marbemac

philsturgeon: I would expect to see "example property type should be number" or `"The property type of \"example\" should be number."`

marbemac: I agree that having just that little bit more context in the error message is helpful. Suggest we keep it in the pretty message.

P0lip:
@marbemac @philsturgeon 
the last remaining question is whether we are making paths configurable - right now we always try to display the property name, but some want to have the whole path printed.
What do you think?

philsturgeon:
If I had to guess at the use case for people who want the whole path, they are probably using it programmatically do if we make it available in the object (and it can be inserted into the message string via a {{path}} variable) then both JS and CLI users should be happy. Then we can keep it property whenever possible by default.

marbemac:
> whether we are making paths configurable

Suggest we punt this to later - not convinced we need it so might as well save the time now :).

P0lip:
> If I had to guess at the use case for people who want the whole path, they are probably using it programmatically do if we make it available in the object (and it can be inserted into the message string via a {{path}} variable) then both JS and CLI users should be happy. Then we can keep it property whenever possible by default.

We do expose `{{path}}` now, so your use case would be covered :)

PR #869, opened 2019-12-20
feat: switch default from warn to error

### LINKED ISSUE #817: Return error only for error level problems in linter
**User story.**
Integrating with CI/CD tooling I will fail spec validation only if there are error-level problems.

Our users of spectral are often not familiar with it and put the validations in CI/CD pipeline based on our internal rulesets. The fact that the default value fails on warnings is first - semantically doubtful, second - requires users to know the tooling better so that it doesn't fail on warnings as well.

Change the default fail level of problems to errors.

**Additional context**

philsturgeon:
We switched it to warn only in develop, master (4.2) is still essentially defaulting to hint.

Closes #817

**Checklist**

- [ ] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


jerzyn:
yay! :)

PR #875, opened 2019-12-23
fix: proper sources for ref resolving errors

P0lip:
~Needs https://github.com/stoplightio/json-ref-resolver/pull/158~

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

It slipped through the cracks when I was addressing the error source issues.
The lack of proper source of ref resolving error is fairly annoying in the context of Studio, since the reference error we scream about may not be included in the file, hence you might be unable to locate it easily if your document has plenty of references.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,7 +85,7 @@ jobs:
           else
             npm publish
           fi
-      - run: npx pkg . --targets linux,macos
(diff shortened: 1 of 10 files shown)


P0lip:
Merged in https://github.com/stoplightio/spectral/pull/878

PR #876, opened 2019-12-23
refactor: introduce Document

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

~Technically this is a breaking change, but I doubt anyone besides us actually passes `IParsedResult`, so we can consider it a fully Internal refactor that mostly addresses the issues with naming and makes certain parts of code a bit easier to reason about.~ not breaking anymore :blush:

I'm not planning to make it a part of 5.0.


src/functions/unreferencedReusableObject.ts
```diff
@@ -1,5 +1,6 @@
+import { isObject } from 'lodash';
```
philsturgeon: The docs talk about "Cherry-pick methods for smaller browserify/rollup/webpack bundles." where you `import isObject at from 'lodash/isObject';` instead. Is this something we should do too?

philsturgeon:
I am a bit concerned about releasing breaking changes in 5.1, is this necessary? Could we hold it off for a while until we have some bigger features which might warrant a 6.0? 

Or could it be made not breaking, by returning the thing we used to return too and deprecating it for now?

P0lip:
> Or could it be made not breaking, by returning the thing we used to return too and deprecating it for now?

Yeah, I can easily make it a non-breaking one and keep it like that.
The usage of that is mostly internal limited to Stoplight anyway, so users won't notice it, but I agree we should not be violating the our versioning model.

src/functions/unreferencedReusableObject.ts
P0lip: We used to have similar imports back in the day, but changed them in all projects across the whole Stoplight organization.


@marbemac was typings an issue? IIRC that was it, but don't remind myself exactly now.
If I am not mistaken, we had to use that awkward import-require syntax, `import isObject = require('lodash/isObject');`

marbemac: It was because tree shaking was not working quite correctly - importing from more specific `lodash/{foo}` type target doesn't rely on tree shaking. We have since moved away from the specific imports though, and to the nicer `import { foo } from 'bar'` style syntax.

philsturgeon: I dunno what tree shaking means I’ll have a google. The only reason I pointed this out is I’ve seen mention that this mixer syntax creates much larger webpack builds due to something something can’t remember why. Might be a worthwhile optimisation, might be a microoptimization.

marbemac: Yeah it's not worth it - in order to make the `lodash/isObject` type syntax work, we'd have to use that syntax consistently everywhere, and hope that our dependencies do as well. Better to try to get tree shaking working, but even then it's not a huge difference in output bundle size.

philsturgeon:
Great ok!

PR #1007, opened 2020-03-11
chore: add stoplight config file

marbemac:
Adds a stoplight config file, to help with our upcoming automation and docs functionality.

- [ ] Yes
- [x] No


.stoplight.json
```diff
@@ -0,0 +1,8 @@
+{
+  "formats": {
+    "markdown": {
```
P0lip: Ha, now that I think of Stoplight config we seem to be lacking the 'disable' option.
The following configuration will still match all openapi and json schema files

philsturgeon: I had expected that leaving them out would do that. If I say we only have markdown in this project we want to only have markdown. Is this something we could change?

P0lip: > I had expected that leaving them out would do that. If I say we only have markdown in this project we want to only have markdown. Is this something we could change?

I have a feeling that wouldn't be really very convenient after all, since you'd always need to provide all formats rather that the ones you want to configure.
What I mean is that let's say you are keen on working with all formats and are fine with the defaults, but want to change the default directory of images. In such case, you just pass `image` format to the config, and call it a day.

PR #1039, opened 2020-03-29
chore: only check docs/ markdown links

philsturgeon:
develop is failing due to this github action trawling through every issue/PR link in the CHANGELOG and hitting 429 rate limiting. It's slow and usually results in a failure, so let's just hit docs/ and see if that speeds things up.

### DIFF
--- a/.github/workflows/markdown-links.yml
+++ b/.github/workflows/markdown-links.yml
@@ -12,5 +12,5 @@ jobs:
     steps:
     - uses: actions/checkout@master
       with:
-        fetch-depth: 1
-    - uses: planetscale/github-action-markdown-link-check@master
+        folder-path: "./docs/"
+    - uses: gaurav-nelson/github-action-markdown-link-check@master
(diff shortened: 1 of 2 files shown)


.github/workflows/markdown-links.yml
```diff
@@ -13,4 +13,5 @@ jobs:
     - uses: actions/checkout@master
       with:
         fetch-depth: 1
-    - uses: planetscale/github-action-markdown-link-check@master
+        folder-path: docs/
+    - uses: gaurav-nelson/github-action-markdown-link-check@v0.6.0
```
philsturgeon: swapped back from the fork we were using after the bug got fixed. This is the same one we are using in master anyhow.

XVincentX:
```suggestion
    - uses: gaurav-nelson/github-action-markdown-link-check@master
```

philsturgeon:
@XVincentX I just checked prism to see how you made it work, but it seems like you've not got this folder-check in place? 

This build is failing because its still checking CHANGELOG.md

```
FILE: ./CHANGELOG.md
```

and its failing. How do I make it only check docs/ and where is it in Prism?

Issue #403, opened 2019-07-23
Invalid schema causes "should have required property '$ref'" error


AnthonyPorthouse:
**Describe the bug**
When an OpenAPI v3 Response Object contains a `schema` which has invalid schemas, an error is thrown saying: should have a required `$ref` property.

This is probably coming from better-ajv-errors, which is deduping errors which are saying "should be a valid sub schema, or should be a $ref" and its just showing the $ref bit. 

**To Reproduce**

1. Given this OpenAPI document 
```yaml
openapi: 3.0.1
info:
  title: Example $ref error
  version: 1.0.0
paths:
  /user:
    get:
      operationId: getUser
      responses:
        "200":
          description: An Example
          content:
            application/json:
              schema:
                type: object
                properties:
                  user_id: 12345
```

2. Run this CLI command `spectral lint openapi.yml`

3. See error

**Expected behavior**

Show a more appropriate validation error. In this instance, other validators return this error:

> Structural error at paths./user.get.responses.200.content.application/json.schema.properties.user_id should be object

**Environment:**
 - Library version: 4.0.1
 - OS: Mac OS 10.14.5

mburtless:
Adding exact error code, so this can be found by other folks that run into this issue.  For the schema above, error code would likely be something like:
```
error  oas3-schema                     /paths//user/get/responses/200 should have required property '$ref'
```
I see this error even if I have ref inside the `schema` object.

tbarn:
@AnthonyPorthouse I need to double check if that's a valid OAS document. I changed the last part to this: 

```
                  user_id:
                    type: integer
```

And it passed. If you are still wanting the example there, I would normally do: 

```
                  user_id:
                    type: integer
                    example: 12345
```

tbarn:
The error message is definitely misleading, but I think the initial document is not valid either way.

AnthonyPorthouse:
@tbarn perhaps the issue then is that it is eating errors saying it should be a `$ref` instead of _why_ it's invalid?

Agreed my initial PoC spec was rushed as it was towards the end of my work day and as you've pointed out is in fact invalid.

Would you prefer I open a new issue describing these issues more correctly, or should we simply go with this issue?

tbarn:
@AnthonyPorthouse We can totally stick to this issue, but could you edit it so it accurate?

philsturgeon:
Got an easily repeatable one: 

[...]

PR #1071, opened 2020-04-07
fix: improve schema validation

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/403

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

Besides addressing 403, it also improves the general readability and consistency of error messages generated by `schema` function.


marbemac:
Prove it - let's see that before and after screenshot for whatever cases were improved :P.

P0lip:
Given 
```yaml
openapi: 3.0.1
info:
  title: Example $ref error
  version: 1.0.0
paths:
  /user:
    get:
      operationId: getUser
      responses:
        "200":
          description: An Example
          content:
            application/json:
              schema:
                type: object
                properties:
                  user_id: 12345
```

5.3.0
image


PR
image

Given
```yaml
openapi: 3.0.1
info:
  title: Example $ref error
  version: 1.0.0
paths:
  /user:
    get:
      operationId: getUser
      responses:
        "200":
          description: An Example
          content:
            application/json:
              schema:
                type: object
                properties:
```

5.3.0
image

PR
image

Given
```yaml
openapi: 3.0.1
info:
  title: Example $ref error
  version: 1.0.0
paths:
  /user:
    get:
      operationId: getUser
      responses:
        "200":
          description: An Example
          parameters:
```

5.3.0
image

PR
image

Given

```yaml
openapi: 3.0.1
info:
  title: Example $ref error
  version: 1.0.0
paths:
  /user:
    get:
      operationId: getUser
      parameters:
        - type: string
          name: module_id
          in: path
          required: true
          schema:
            type: string
          description: Module ID
      responses:
```

5.3.0
image

PR
image


Etc, etc.

src/functions/schema.ts
```diff
@@ -100,26 +105,28 @@ const validators = new (class extends WeakMap<JSONSchema, ValidateFunction> {
   }
 })();
 
-const replaceProperty = (substring: string, group1: string) => {
-  if (group1) {
-    return 'property ';
+const replaceProperty = (substring: string, _: Optional<number | string>, propertyName: Optional<string>) => {
+  if (typeof _ === 'string' && propertyName !== void 0) {
```
nulltoken, line 109: @P0lip As it's actually used in the code, can't we give this `_` param a more meaningful name?

src/rulesets/oas/functions/__tests__/oasDocumentSchema.test.ts
```diff
@@ -0,0 +1,77 @@
...
+          info: {
+            title: 'response example',
+            version: '1.0',
+          },
+          paths: {
+            '/user': {
+              get: {
+                operationId: 'd',
+                responses: {
+                  200: {},
```
nulltoken, line 60: @P0lip Although this is not caught by the schema validation, how about changing the key `"200"` to more compliant to the spec?

src/rulesets/oas/functions/oasDocumentSchema.ts
```diff
@@ -0,0 +1,29 @@
+import * as AJV from 'ajv';
+import { ISchemaOptions } from '../../../functions/schema';
+import { IFunction, IFunctionContext } from '../../../types';
+
+// /shrug
+function prepareResults(errors: AJV.ErrorObject[]) {
```
nulltoken, line 6: @P0lip As this is about to become a somewhat cornerstone of schema validation, how about covering it with targeted unit tests to better express what we expect from this function?

src/functions/schema.ts
```diff
@@ -14,6 +14,10 @@ export interface ISchemaOptions {
   schema: object;
   // The oasVersion, either 2 or 3 for OpenAPI Spec versions, could also be 3.1 or a larger number if there's a need for it, otherwise JSON Schema
   oasVersion?: Optional<number>;
+  allErrors?: boolean;
+
+  // this is used by oasDocumentSchema function, to be removed once we sort out
```
nulltoken, line 19: @P0lip What's the foreseen plan here with regards to `schema` vs `oasDocumentSchema`? Should the later eventually replace the former? Do you foresee to keep them both?

src/functions/schema.ts
```diff
@@ -45,6 +49,7 @@ function getAjv(oasVersion?: Optional<number>): AJV.Ajv {
   const ajvOpts: AJV.Options = {
     meta: true, // Add default meta schemas (draft 7 at the moment)
     schemaId: 'auto',
+    allErrors,
```
nulltoken, line 52: @P0lip When would we want to not retrieve all the errors?

philsturgeon:
@P0lip this is amazing! It's the biggest issue we have with Spectral and it looks like you've nailed it.

I know it might feel annoying and redundant, but can you make a test harness for each of those examples you put in? As it's such a complicated and important thing, I want to smother it with coverage so it doesn't rear its ugly head again.

src/functions/schema.ts
P0lip: The default is false, so I simply retained it.
We'll want to report all errors in most cases, I believe, but I didn't want to introduce that change as a part of this PR, as this would cause even more noise.

src/rulesets/oas/functions/__tests__/oasDocumentSchema.test.ts
P0lip: This is already a string, as only string and symbols as recognized as valid property keys.
Everything else automatically coerced to a string. To illustrate, let's take the following object literal:
```
200: {},
"200": {}
```
In such chase, the object have 2 duplicate properties, both string 200.
Does it make sense?

src/functions/schema.ts
P0lip: > @P0lip What's the foreseen plan here with regards to schema vs oasDocumentSchema? Should the later eventually replace the former? Do you foresee to keep them both?

I'm eyeing replacing ajv with some other schema validator, that, hopefully, produces more human-friendly errors.
For the time being, we need to prepare the results on our own.
If you need some dedicated treatment in case of errors reported by asyncapi-schema rule (assuming we have such), feel free to spin `asyncDocumentSchema` fn.

src/rulesets/oas/functions/oasDocumentSchema.ts
philsturgeon: Agreed. Also more comments on _why_ would help future maintainers, our future selves (and my current self)  understand whats going on better.

src/rulesets/oas/functions/__tests__/oasDocumentSchema.test.ts
nulltoken: @P0lip Yep! Thanks!

test-harness/scenarios/oas3-schema.scenario
```diff
@@ -0,0 +1,63 @@
+====test====
+The amount of enabled rules is printed in a parenthesis.
```
nulltoken, line 2: @P0lip Copy/paste issue?

src/functions/schema.ts
nulltoken: Thx!

src/functions/schema.ts
nulltoken: Ok.

P0lip:
@nulltoken @philsturgeon do you have any additional comments or suggestions?

nulltoken:
> @nulltoken @philsturgeon do you have any additional comments or suggestions?

To be honest, I've read the code of `prepareResults` at least three times, I still have no idea what it exactly does. Would it be very time consuming to export it and cover it with tests putting under the lights the input AJV-like error objects and the expected results?

By taking a look at the results of the integration tests, I can see that it indeed works, however, I'd really like to get a better understanding at how it works and solves the initial issue.

If we ever need to extend/patch this function, I'd feel more at ease with a set of tight unit tests that precisely express the behavioral contract of the function.

P0lip:
> To be honest, I've read the code of prepareResults at least three times, I still have no idea what it exactly does. Would it be very time consuming to export it and cover it with tests putting under the lights the input AJV-like error objects and the expected results?

Yep, I'll prepare errors. It'll certainly make it easier for everyone to reason about that function when you show what kind of errors it filters out and how it does it.

P0lip:
@nulltoken added some. LMK if it's more clear now. In general all it does is removal of "evil" errors.

src/rulesets/oas/functions/__tests__/oasDocumentSchema.test.ts
```diff
@@ -189,4 +190,158 @@ describe('oasDocumentSchema', () => {
...
+          dataPath: '/paths/test/post/parameters/0/schema',
+          schemaPath: '#/properties/schema/oneOf',
+          params: { passingSchemas: null },
+          message: 'should match exactly one schema in oneOf',
+        },
+      ];
+
+      prepareResults(errors);
+
+      expect(errors).toStrictEqual([
```
nulltoken, line 222: @P0lip In that case, I wonder if we shouldn't pick the `oneOf` error. It gives a proper direction to work with and doesn't lead the user into fixing a specific branch of the `oneOf`, no?

P0lip: Actually `oneOf` is pretty useless here as we have 2 options for most of the time.
There are a few exceptions, i.e. security components I covered manually, but apart from them we usually deal with a simple case that can be literally expressed as: "either proper value of $ref property". That being said, I strip both `oneOf` and `$ref`, since we are always interested in the first error. Why? Both oas-schema rules operate on resolved content, so it's actually not really `oneOf` anymore - it's always the first member of oneOf we match against.
Does it make sense?

src/rulesets/oas/functions/oasDocumentSchema.ts
```diff
@@ -0,0 +1,64 @@
...
+    path: /^components\/securitySchemes\/[^/]+$/,
+    message: 'Invalid security scheme',
+  },
+  {
+    path: /^securityDefinitions\/[^/]+$/,
+    message: 'Invalid security definition',
+  },
+];
+
+// /shrug
```
nulltoken, line 24: @P0lip I think that you can drop this, now ;-)

src/rulesets/oas/functions/__tests__/oasDocumentSchema.test.ts
nulltoken: @P0lip Yep! Pretty clear! The "already resolved" part indeed sells it.

Could you add that explanation as comments in the error filtering function?

P0lip:
@nulltoken done!

huksley:
Tried to use latest develop branch with this fix applied, have a different error from my issue I have
https://github.com/maasglobal/maas-tsp-api/issues/70#issuecomment-614221659

```
/Users/user/src/maas-tsp-api/schemas/core/components/travel-mode.json
 1:1  error  oas3-schema  Property `$id` is not expected to be here.
```

I see no problem with this JSON schema file, see https://github.com/maasglobal/maas-schemas/blob/develop/maas-schemas/schemas/core/components/travel-mode.json

P0lip:
Is `$id` considered a valid property in OAS-flavor of JSON Schema?
I'm not sure and cannot seem to find anything about that in docs.
I've checked the JSON Schema [model](https://github.com/stoplightio/spectral/blob/develop/src/rulesets/oas/schemas/schema.oas3.json) we use for validation and it has no trace of `$id`.

@philsturgeon any clue?

Issue #1130, opened 2020-04-28
Equivilent paths points at paths not the bad path


philsturgeon:
If there are two paths that are causing a problem, we should point to one of those problematic paths, not just `paths:`.

<img width="1116" alt="image" src="https://user-images.githubusercontent.com/67381/80498624-dd8d1600-8963-11ea-8a49-4e7fc8018bc5.png">

Clicking that problem on the right "The paths foo and bar are equivalent" takes me to the paths object, and then I have to scroll for ages to find the problematic ones. That's if I'm in code view anyway. In Form view it just does nothing.

P0lip:
> Clicking that problem on the right "The paths foo and bar are equivalent" takes me to the paths object, and then I have to scroll for ages to find the problematic ones. That's if I'm in code view anyway. In Form view it just does nothing.

Which one is the problematic one? I assume is the latter one in the document, but just want to make sure we are on the same page.

PR #1135, opened 2020-05-01
fix(oas): path-params reports actual endpoint path

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1130

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Make sure to answer my question before reviewing :sweat_smile: 
https://github.com/stoplightio/spectral/issues/1130#issuecomment-622588402


src/rulesets/oas/functions/__tests__/oasPathParam.test.ts
```diff
@@ -284,15 +288,15 @@ describe('oasPathParam', () => {
       {
         code: 'path-params',
         message: `The paths \`/foo/{boo}\` and \`/foo/{bar}\` are equivalent.`,
-        path: ['paths'],
+        path: ['paths', '/foo/{bar}'],
         range: {
           end: {
```
nulltoken, line 293: @P0lip How come the end block of `['paths']` is the same for `['paths', '/foo/{bar}']`?
As we're getting one "step" within, I would have expected it to change.

P0lip: I guess I'd need to pass a string to `run` rather than an actual object literal.
Otherwise ranges are, well, not really precise, because we stringify the object again.
I'll do it.

P0lip:
This is the range for `paths` image

PR #1171, opened 2020-05-19
chore(repo): make it possible run test-harness on Windows

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

Pretty sure @nulltoken will appreciate the change in particular, as azure-pipelines job was never triggered for his commits :wink:

Note - the windows job does not perform browser testing, as it doesn't seem like headless browsers are supported yet :confused: 

@nulltoken any clue why test-harness step fails for test-windows job?
I have no access to Windows instance at the moment, so cannot verify it.
I suppose changing the shell to `bash` would do the trick, but I don't want to do it since powershell is 
likely to be more popular among Windows users.


nulltoken:
@P0lip Amazing PR! <3

>any clue why test-harness step fails for test-windows job?

@P0lip Unfortunately, it's a known issue (cf. https://github.com/stoplightio/spectral/issues/919#issuecomment-575371411). 

I've reduced the pain and the feedback loop by introducing https://github.com/stoplightio/spectral/pull/999 as a workaround.

However the issues you're seeing look completely different from the one I was suffering from. Those look related to path issues whereas mine where more of the "AJV ref resolving" kind. 

Can this be stalled for a few days? I'll take a look at it during the week-end.

P0lip:
> Can this be stalled for a few days? I'll take a look at it during the week-end.

Yep, sure thing.  No rush, feel free to take a look at it at your leisure.
I'll try to investigate the issue either once I boot Windows.

nulltoken:
@P0lip Below some findings:

- `cli acceptance tests › exceptions.oas3.stdin.scenario file › Do not report exceptions when document comes through stdin`
   - Fails on Circle-CI with `'c:' is not recognized as an internal or external command, operable program or batch file.`
   - Fails on my box with `'cat' is not recognized as an internal or external command, operable program or batch file.` (which makes more sense in a Windows env)
- `cli acceptance tests › oas3-schema.scenario file › Prints reasonable error messages for oas3-schema errors.`
  - Similar results on both CircleCI and my box. That somehow reminds of the previous errors I was running when running harness. All of them were AJV related. And it that case, it looks like the whole oas3-schema rule is ignored.
- `cli acceptance tests › help-no-document.scenario file › Errors when no document is provided`
  - I've managed to move it a bit forward by applying the following patch. But we're still trying to run a bash script on a Windows box ;-)

```
diff --git a/test-harness/scenarios/help-no-document.scenario b/test-harness/scenarios/help-no-document.scenario
index ebb2005..a2cbb1c 100644
--- a/test-harness/scenarios/help-no-document.scenario
+++ b/test-harness/scenarios/help-no-document.scenario
@@ -1,7 +1,7 @@
 ====test====
 Errors when no document is provided
 ====command====
-./scripts/faketty.bash {bin} lint
+{cwd}/scripts/faketty.bash {bin} lint
 ====stdout====
 spectral lint [documents..]
 ```

```
diff --git a/test-harness/index.ts b/test-harness/index.ts
index 541c3de..0691d03 100644
--- a/test-harness/index.ts
+++ b/test-harness/index.ts
@@ -20,6 +20,7 @@ describe('cli acceptance tests', () => {
     const replacements: Dictionary<string> = {
       __dirname: normalize(__dirname),
       bin: spectralBin,
+      cwd
     };
 
     const tmpFileHandles = new Map<string, tmp.FileResult>();
```

**Comment by nulltoken** [comment]
@p0lip With regards to the ´cat’ on Windows, maybe could we create ´dump’, ´dump.cmd’ and ‘dump.ps1’ under the ´scripts’ folder. With a little luck calling ´{cwd}/scripts/dump’ would work everywhere (similar hack to what npm/yarn does with the ´.bin’ folder)

**Comment by nulltoken** [comment]
> @P0lip With regards to the ´cat’ on Windows, maybe could we create ´dump’, ´dump.cmd’ and ‘dump.ps1’ under the ´scripts’ folder. With a little luck calling ´{cwd}/scripts/dump’ would work everywhere (similar hack to what npm/yarn does with the ´.bin’ folder)

@P0lip After some further testing, it doesn't seem to be that straightforward (differences between the way `cmd` and `powershell` deal with pipelines, more or less required quoting in cmd, ...)

I wonder if it wouldn't be easier to accept, beside `====command====`, something like `====command-nix====` and `====command-win====`? This would make easier to deal with subtleties of each OS (and potentially find a different approach for the `faketty` won windows).

Proposal would be
-> `command` cannot be used along `====command-nix====` and `====command-win====` and should error the harness test
-> only `command` -> run it everywhere
-> only `====command-nix====` -> skip the test on Windows
-> only `====command-win====` -> skip the test on *nix platforms

Thoughts?

**Comment by P0lip** [comment]
> @P0lip With regards to the ´cat’ on Windows, maybe could we create ´dump’, ´dump.cmd’ and ‘dump.ps1’ under the ´scripts’ folder. With a little luck calling ´{cwd}/scripts/dump’ would work everywhere (similar hack to what npm/yarn does with the ´.bin’ folder)


Sounds good to me.

> I wonder if it wouldn't be easier to accept, beside ====command====, something like ====command-nix==== and ====command-win====? This would make it easier to deal with subtleties of each OS (and potentially find a different approach for the faketty on windows).

Agreed. I'm convinced the commands look differently in real-life too, as Windows folks certainly account for the differences and commands vary.

**Comment by P0lip** [comment]
@nulltoken going to implement your proposal re commands.

**Comment by P0lip** [comment]
@nulltoken we could potentially have this in, but we have 2 failing oas-schema tests we know about.
Any chance you looked into it?

**Comment by nulltoken** [comment]
> @nulltoken we could potentially have this in, but we have 2 failing oas-schema tests we know about.
> Any chance you looked into it?

@P0lip Awesome work! I'll give it a look tomorrow.

**Comment by nulltoken** [comment]
@P0lip It seems the root cause is similar to https://github.com/stoplightio/spectral/issues/919#issuecomment-586628906

Applying the following patch

```diff
diff --git a/src/functions/schema.ts b/src/functions/schema.ts
index 71ec795..ecde87d 100644
--- a/src/functions/schema.ts
+++ b/src/functions/schema.ts
@@ -111,6 +111,10 @@ const validators = new (class extends WeakMap<JSONSchema, ValidateFunction> {
 })();
 
 export const schema: ISchemaFunction = (targetVal, opts, paths, { rule }) => {
+  console.log('targetVal', targetVal);
+  console.log('opts', opts);
+  console.log('paths', paths);
+  console.log('rule', rule);
   const path = paths.target ?? paths.given;
 
   if (targetVal === void 0) {
```

triggers the following output

```
expect(received).toEqual(expected) // deep equality

    - Expected  - 11
    + Received  + 27

      OpenAPI 3.x detected
    -
    - c:/Users/fermaem/AppData/Local/Temp/asset--17156-0IM9WnfhxsL6-.yml
    -   5:11  error  oas3-schema  Property `foo` is not expected to be here.
    -  12:11  error  oas3-schema  Property `type` is not expected to be here.
    -  23:18  error  oas3-schema  `type` property type should be string.
    -  32:28  error  oas3-schema  `user_id` property type should be object.
    -  36:28  error  oas3-schema  `properties` property type should be object.
    -  38:23  error  oas3-schema  `description` property type should be string.
    -  41:17  error  oas3-schema  `responses` property should not have fewer than 1 properties.
    -
    - ✖ 7 problems (7 errors, 0 warnings, 0 infos, 0 hints)
    + targetVal {
    +   openapi: '3.0.1',
    +   info: {
    +     title: 'Example $ref error',
    +     version: '1.0.0',
    +     contact: { foo: 'me' }
    +   },
    +   paths: { '/user': { get: [Object] }, '/address': { get: [Object] } }
    + }
    + opts {
    +   allErrors: true,
    +   schema: { '$ref': './schemas/schema.oas3.json' },
    +   prepareResults: [Function: n]
    + }
    + paths { given: [], target: [] }
    + rule Rule {
    +   name: 'oas3-schema',
    +   description: 'Validate structure of OpenAPI v3 specification.',
    +   message: '{{error}}.',
    +   severity: 0,
    +   resolved: true,
    +   formats: [ 'oas3' ],
    +   then: [ { function: 'HCldYUWtY-PvymVRih4D0', functionOptions: [Object] } ],
    +   given: [ '$' ]
    + }
    + applyManualReplacements []
    + No results with a severity of 'error' or higher found!
```

I'm going to dig a little deeper this evening. However, it those above ring a bell to you, please chime in ;-)

P0lip:
@nulltoken schema is not resolved for some reason, and this is causing the issue.
I have no clue why exactly, but seems like something is off in `src/rulesets/reader.ts`.

nulltoken:
@P0lip Made some progress. Looks like the issue originates from

https://github.com/stoplightio/spectral/blob/develop/src/functions/schema.ts#L105

Where ajv, given `{"$ref":"./schemas/schema.oas3.json"}` doesn't seem to be able to properly deal with the "faked" pkg filesystem.

It throws an exception (`{"message":"can't resolve reference ./schemas/schema.oas3.json from id #","missingRef":"schemas/schema.oas3.json","missingSchema":"schemas/schema.oas3.json"}`), which is eventually silenced in the catch clause.

Looks like there's an async version of `compile()` that accepts a `loadSchema()` [callback](https://github.com/ajv-validator/ajv#asynchronous-schema-compilation). This might be a way to inject some resolving sense into it.

Do you want me to try and work this out further (or is the `loadSchema()` thingie totally not the direction which should be taken)?

P0lip:
> Where ajv, given {"$ref":"./schemas/schema.oas3.json"} doesn't seem to be able to properly deal with the "faked" pkg filesystem.

AJV should not be given it at all in this case - the $ref should be resolved during the ruleset loading process.
This is what happens in case of *NIX. Not sure why that $ref is left when running the test in Windows.

nulltoken:
> the $ref should be resolved during the ruleset loading process.

That's a good hint!

So schema() should be given an already fully resolved schema. I can't start from this assumption and work up the stack and see where that fails to happen (and with a little luck, why).

P0lip:
> So schema() should be given an already fully resolved schema.

Yep, that's right. This is exactly what happens on my machine - ajv is given an already resolved schema.

image

nulltoken:
@P0lip 

>I ~~can't~~ can start from this assumption and work up the stack and see where that fails to happen (and with a little luck, why).

Edited: sorry. Typo (Duh)

P0lip:
> Edited: sorry. Typo (Duh)

Yeah, I assumed you meant to type "can" :D

nulltoken:
> Yeah, I assumed you meant to type "can" :D

`<relieved />` ;-)

nulltoken:
@P0lip For some reason, I can't seem to properly run the harness test through a docker on my box without suffering from 

```
 ● cli acceptance tests › oas3-schema.scenario file › Prints reasonable error messages for oas3-schema errors.

    : Timeout - Async callback was not invoked within the 240000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 240000ms timeout specified by jest.setTimeout.Error:
```

So, I'm going to spam that CI and work the issue through `printf` based debugging...

nulltoken:
@P0lip Made some minor progress.

Surfacing the resolving errors gives some hint about the issue (which _seems_ pkg related. 

pkg debug log seems to show that the missing files have been embedded
image

`resolveFile` relies on `readFile` to read those (which pkg seems to support).

Next step would to try and `readdir` the content of the virtual `snapshot` folder on Windows in order to see what's really in there.

P0lip:
@nulltoken oh, interesting find.
LMK what the result of `readdir` is.

src/fs/reader.ts
```diff
@@ -14,34 +14,21 @@ export interface IReadOptions {
...
-
-    stream.on('readable', () => {
-      let chunk: string | null;
-
-      while ((chunk = stream.read()) !== null) {
-        result += chunk;
-      }
-    });
+    // Piped through STDIN
+    return readFrom(name, opts.encoding);
```
nulltoken, line 18: @P0lip This doesn't fail on windows, whereas the `createReadStream` cringes with "EOF found"

src/resolvers/http-and-file.ts
```diff
@@ -13,11 +15,24 @@ export const httpAndFileResolver = createHttpAndFileResolver();
 export function createHttpAndFileResolver(opts?: IHttpAndFileResolverOptions): Resolver {
   const resolveHttp = createResolveHttp({ ...DEFAULT_REQUEST_OPTIONS, ...opts });
 
+  // Note: this should be better fixed at @stoplight/json-ref-readers level
+  const patchedResolved = (ref: URI): Promise<unknown> => {
+    return new Promise((resolve, reject) => {
+      const path = ref.valueOf();
```
nulltoken, line 21: given `c:/toto/file.txt`, `pathname()` returns `/toto/file.txt`. This is was causing the main resolving havoc.

test-harness/scenarios/help-no-document.scenario
```diff
@@ -1,6 +1,6 @@
 ====test====
 Errors when no document is provided
-====command====
+====command-nix====
```
nulltoken, line 3: standard win shell doesn't know bash ;-)

nulltoken:
@P0lip The situation is a little bit better on Windows:

- schema based harness tests do not fail any more (pathname vs valueOf)
- `cat|type tot.yml | spectral lint` works in a Powershell command line (readFile vs createReadStream)

However, `cat tot.yml | spectral lint` still doesn't work when spawned from the test harness.
 - default shell is CMD (and doesn't know `cat`)
 - CMD supports (but doesn't REALLY like forward slashes)
 - ...

nulltoken:
@P0lip I've made some progress on the harness side.

Seems I've broken other things. Will take a look at this tomorrow.

nulltoken:
@P0lip Fixed the node tests. Harness tests seem to be more or less working (still have to tackle some additional newlines in the output and minor other things).

Before going any further, does this seem a proper approach to you?

src/rulesets/reader.ts
```diff
@@ -79,12 +79,16 @@ const createRulesetProcessor = (
         dereferenceInline: false,
         uriCache,
         async parseResolveResult(opts) {
-          opts.result = parseContent(opts.result, opts.targetAuthority.pathname());
+          opts.result = parseContent(opts.result, opts.targetAuthority.valueOf());
```
P0lip, line 82: We could also use `normalize` from `@stoplight/path` to make sure paths are in tact.
```suggestion
          opts.result = parseContent(opts.result, normalize(opts.targetAuthority.valueOf()));
```

P0lip (commented):
Looks good to me in general
I'm totally fine with the direction you took here

src/rulesets/reader.ts
P0lip: and potentially in other spots we care about in `reader.ts`

test-harness/helpers.ts
```diff
@@ -44,15 +44,21 @@ export function parseScenarioFile(data: string): IScenarioFile {
   const env = getItem(split, 'env');
 
   if (command === void 0) {
-    if (commandWindows !== void 0 && commandUnix !== void 0) {
-      throw new Error('No ====command==== provided');
+    if (commandWindows === void 0 && commandUnix === void 0) {
+      throw new Error(
+        'When no ====command==== is provided, both ====command-nix==== and ====command-win==== should be defined',
```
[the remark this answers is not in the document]
nulltoken: @P0lip Updated with a different approach

nulltoken:
@P0lip Still making progress and reducing the number of issues.

However, I'm starting to wonder if that PowerShell bridge is reliable enough (additional starting newlines I can't really make sense of for now, encoding, ...). I may have another idea I'll try if I don't feel the code gets too wonky.

test-harness/helpers.ts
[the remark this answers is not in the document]
P0lip: Well, I actually like the change, since it might be easy to forget to specify a Windows/Unix counterpart.
To reduce the severity of that decision, we could potentially issue a warning instead of error out the whole suite.

test-harness/spawn.ts
```diff
@@ -44,7 +44,51 @@ function stringifyStream(stream: Transform) {
...
+export const spawnPowershell: SpawnFn = async (command, env, cwd): Promise<SpawnReturn> => {
+  const ps = new Shell({
+    executionPolicy: 'Bypass',
+    noProfile: true,
+    inputEncoding: 'utf8',
+    outputEncoding: 'utf8',
+  });
+
+  const winCommand = command.replace(/\/binaries\/(spectral\.exe|spectral)/, '/binaries/spectral.exe');
+  const wrappedCommand = `cd '${cwd}';${winCommand};echo LASTEXITCODE=$LASTEXITCODE`;
```
P0lip, line 59: `$LASTEXITCODE` is always empty :/ 
@nulltoken any clue?
no hurry, take your time to respond :)

PR #1183, opened 2020-05-25
chore: nix/win specific commands in test-harness

P0lip:
As suggested in https://github.com/stoplightio/spectral/pull/1171

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


test-harness/index.ts
```diff
@@ -17,6 +17,9 @@ describe('cli acceptance tests', () => {
   describe.each(files)('%s file', file => {
     const data = fs.readFileSync(path.join(cwd, file), { encoding: 'utf8' });
     const scenario = parseScenarioFile(data);
+
+    if (scenario.command === void 0) return;
```
nulltoken, line 21: @P0lip Will this mark the test as 'skipped' in the console? Wouldn't that be the case, how complex would it be to make it happen? That would be pretty handy to detect scenario that are not fully xplat...

P0lip: Nah, it won't. Can make it marked though.

P0lip:
image
image

nulltoken:
> Marked them as 'todo'

Perfect!

PR #1188, opened 2020-05-29
perf: improve ruleset loading

P0lip:
`copyRule` unnecessarily performs a deep clone that is expensive at times due to the presence of oas2 & oas3 JSON Schema that need to be copied over.
We can perform a shallow copy here, as we do not alter the value of `functionOptions` while merging rules.
`then` is mutated however [here](https://github.com/stoplightio/spectral/blob/develop/src/rulesets/mergers/rules.ts#L98), so we need to account for this.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**
Before:
image

After:
image
zoomed in :P
image

**Additional context**

Not a big deal in case of CLI users, but fairly important for Studio users, as the Spectral activation is slower than it has to be.
Bonus change - we prefer `JSON.parse` when the filepath has `.json` extension or the url's pathname ends with `.json`.

Last but not least - a couple of ESLint warnings addressed.


src/rulesets/reader.ts
```diff
@@ -15,6 +15,14 @@ export interface IRulesetReadOptions {
   timeout?: number;
 }
 
+function parseContent(content: string, source: string): unknown {
+  if (extname(source) === '.json') {
```
P0lip, line 19: `stoplight/path#extname` handles URLs.
image

PR #1239, opened 2020-06-18
test(reader): try to reduce flakiness under Windows

P0lip:
Whether it helps or not - we shall see, but I hope so.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/src/fs/__tests__/reader.jest.test.ts
+++ b/src/fs/__tests__/reader.jest.test.ts
@@ -30,7 +30,7 @@ end
     });
 
     it('throws when fd cannot be accessed', () => {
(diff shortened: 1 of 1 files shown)


src/fs/__tests__/reader.jest.test.ts
```diff
@@ -30,7 +30,7 @@ end
     });
 
     it('throws when fd cannot be accessed', () => {
-      return expect(readFile(2, { encoding: 'utf8' })).rejects.toThrow();
+      return expect(readFile(2147483647, { encoding: 'utf8' })).rejects.toThrow();
```
P0lip, line 33: This is the highest valid number.

PR #1247, opened 2020-06-23
chore(dependabot): use version 2

P0lip:
Upgrades to new dependabot and fixes_ https://github.com/stoplightio/spectral/issues/1206, but let's have the issue open.

If I am not mistaken, security updates should still be delivered by [GitHub](https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#about-the-dependabotyml-file), because we have that option enabled, so we are good here.
Prod deps are locked anyway, and we rarely update them, so no harm is done, however, it would be freaking awesome if Dependabot folks added support for multiple projects, so we can eventually rely on Dependabot in case of prod deps.
There are several requests for it:
https://github.com/dependabot/feedback/issues/966
https://github.com/dependabot/dependabot-core/issues/1865
https://github.com/dependabot/dependabot-core/issues/317

The documentation for the new config can be found [here](https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates)


philsturgeon:
The docs you linked to seem to suggest this file be renamed to `.github/dependabot.yml`?

P0lip:
Oh, right. Nice catch. Let me move it.

Issue #901, opened 2020-01-10
oas3-valid-...-example: Only checks example, but not examples


m-mohr:
**Describe the bug**

This is somewhat related to #882: All the oas-valid-...-example rules only check the examples in `example`, but not the examples in `examples.*.value`. 

**To Reproduce**

OAS3 file adapted from the second test in src/rulesets/__tests__/templates/_oas-examples.ts:
```js
    {
      openapi: '3.0.0',
      headers: {
        xoxo: {
          schema: {
            type: 'string',
          },
          examples: {
            test1: {
              value: 123
            },
          },
        },
      },
    }
```

If I add to the test:
```js
  test('will fail when simple example is invalid (examples)', async () => {
    const results = await s.run({
      openapi: '3.0.0',
      [path]: {
        xoxo: {
          schema: {
            type: 'string',
          },
          examples: {
            test1: {
              value: 123
            },
          },
        },
      },
    });
    expect(results).toEqual([
      expect.objectContaining({
        severity: DiagnosticSeverity.Error,
        code: ruleName,
        message: '`examples.*.value` property type should be string',
      }),
    ]);
  });
```

and run: `npm test "oas3-valid-oas-header"`, it doesn't check the examples.test.value value.

**Expected behavior**

All examples should be checked.

**Environment (remove any that are not applicable):**
 - Library version: 5.0.0
 - OS: Windows 10

PR #1284, opened 2020-07-16
fix(oas): improve examples validation

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/901

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


src/rulesets/oas/functions/oasExample.ts
```diff
@@ -0,0 +1,78 @@
+import type { IFunction, IFunctionContext, IFunctionResult } from '../../../types';
+import type { Optional, Dictionary } from '@stoplight/types';
+import type { ISchemaOptions } from '../../../functions/schema';
+
+function isObject(value: unknown): value is Dictionary<any> {
+  // todo: expose a util, so it can be used everywhere
+  return value !== null && typeof value === 'object';
```
nulltoken, line 7: @P0lip or maybe just use lodash `isOjbectLike`?

src/rulesets/oas/index.json
```diff
@@ -418,10 +419,9 @@
       "type": "validation",
       "given": "$..parameters..[?(@.in == 'body')]..[?(@property !== 'properties' && @.example && ( @.type || @.format || @.$ref ))]",
       "then": {
-        "function": "schemaPath",
+        "function": "oasExample",
```
nulltoken, line 422: @P0lip My understanding is that oas2 only supports `examples` on responses. Woudn't this try and validate an `examples` field here would it exist?

P0lip: Yeah, it'd, but we can ignore it out if needed.

nulltoken: @P0lip maybe passing, per rule, a whitelist of valid fields as functionOptions?

philsturgeon:
These rules have always been hell, due to the _incredibly confusing_ nature of all the different types of examples that can show up anywhere in OAS2, OAS3, and JSON Schema.

It honestly might need custom functions.

Using https://github.com/stoplightio/studio/issues/381 as a reference to all the bajillion different combinations of rule, can we think about how we'd implement what combination of examples to handle:

Schema Objects (for OAS2, OAS3, JSON Schema) when there could be x-example or example or examples or x-examples depending on whatever type of thing is there, and the value is always a bare value, or an array of bare values.

Then there's Media Types, Parameters, etc which are sometimes "Schema like" but not, or might actually have a "Schema" inside them...

No one rule could do everything, and we keep getting bug reports when we miss a certain specific combination, so whatever we do needs to work for everything. 

P.S No need to validate externalValue.

src/rulesets/oas/functions/oasExample.ts
P0lip: Yeah, I considered doing it, but hm, tree-shaking feels shaky :laughing: 
Let me see how it impacts the size of the bundled fn.
It's the difference is negligible, I'll go for isObjectLike

src/rulesets/oas/index.json
P0lip: Hmmm, so if I understand it correctly, we'd only need to support `responses` in case of OAS2.
@philsturgeon correct?

philsturgeon: @P0lip sorry friend I don't understand the question.

P0lip: so what I meant was whether OAS2 allows you to place `examples` under other spots than just responses.

philsturgeon: I think this thread makes it about as clear as I could hope to make it, but it’s a very confusing situation.

philsturgeon: https://github.com/stoplightio/studio/issues/381#issuecomment-626944301

P0lip: Yeah, I went through it, but still feels like a nightmare to maintain :sweat_smile: 
In any case, I'll try to come up with something decent enough that should cover most of cases we have to support.

src/rulesets/oas/functions/oasExample.ts
```diff
@@ -0,0 +1,127 @@
...
+  schemaField: string;
+  type: 'media' | 'schema';
+}
+
+const ORDER_CHECK = {
+  media: {
+    2: [
+      {
+        field: 'examples',
+        multiple: true,
```
philsturgeon, line 20: Media `examples` in OAS2 are bare values, and are usually a MIME type as the name:

```yaml
examples:
  application/json: "the json example"
  application/xml: "the xml example"
```

In OAS3 they have a `value` key:

```yaml
examples:
  foo: 
    value: "the foo example"
  bar:
    value: "the bar example"
```

A more full example:

```
    paths:
      /users:
        post:
          summary: Adds a new user
          requestBody:
            content:
              application/json:     # Media type
                schema:             # Request body contents
                  $ref: '#/components/schemas/User'   # Reference to an object
                examples:    # Child of media type
                  Jessica:   # Example 1
                    value:
                      id: 10
                      name: Jessica Smith
                  Ron:       # Example 2
                    value:
                      id: 11
                      name: Ron Stewart
          responses:
            '200':
              description: OK
```

This example is here. https://swagger.io/docs/specification/adding-examples/

We'll need to make sure we not only say "multiple: true/false" but... is the value keyed behind something or right there?

src/rulesets/oas/functions/oasExample.ts
```diff
@@ -0,0 +1,127 @@
...
+      for (const exampleKey of Object.keys(value)) {
+        const exampleValue = value[exampleKey];
+        if (!isObject(exampleValue)) {
+          // should be covered by oas3-examples-value-or-externalValue
+          continue;
+        }
+
+        const result = this.functions.schema.call(
+          this,
+          exampleValue.value,
```
philsturgeon, line 95: Related to my other comment: schema `examples` can be `multiple: true` but do not have a `value` key.

```
{
  "title" : "Match anything",
  "description" : "This is a schema that matches anything.",
  "default" : "Default value",
  "examples" : [
    "Anything",
    4035
  ]
}
```

src/rulesets/oas/index.json
```diff
@@ -594,14 +595,15 @@
       "message": "{{error}}",
       "recommended": true,
       "severity": 0,
+      "formats": ["oas3"],
       "type": "validation",
-      "given": "$..parameters..[?(@.example && @.schema)]",
+      "given": "$..parameters..[?(@.schema && (@.example || @.examples))]",
```
philsturgeon, line 600: See the OAS3 spec for Parameter objects, they can taker example or examples at that top level, which is correct with what you've got here. 

image

They can also take the OAS `example` or the "JSON Schema" `examples`:

There can be an `example` inside the schema for a parameter or next to it.

image

The example key could be inside or next to schema, and it works the same. The examples key inside schema would be a "JSON Schema" examples which is a bare array (no values key), and the examples next to schema would be an OAS examples, which has the `values` key inside.

src/rulesets/oas/functions/oasExample.ts
P0lip: @philsturgeon do we actually have any rule matching these OAS2 media examples?
In case of OAS2 and its examples, we only have the following two rules
```
"oas2-valid-parameter-example": {
      "description": "Examples must be valid against their defined schema.",
      "message": "{{error}}",
      "recommended": true,
      "formats": ["oas2"],
      "severity": 0,
      "type": "validation",
      "given": "$..parameters..[?(@.in == 'body')]..[?(@property !== 'properties' && (@.example || @['x-example']) && (@.type || @.format || @.$ref ))]",
      "then": {
        "function": "oasExample",
        "functionOptions": {
          "schemaField": "$",
          "oasVersion": 2,
          "type": "schema"
        }
      }
    },
    "oas2-valid-definition-example": {
      "description": "Examples must be valid against their defined schema.",
      "message": "{{error}}",
      "recommended": true,
      "formats": ["oas2"],
      "severity": 0,
      "type": "validation",
      "given": "$..definitions..[?(@property !== 'properties' && (@.example || @['x-example']) && (@.type || @.format || @.$ref))]",
      "then": {
        "function": "oasExample",
        "functionOptions": {
          "schemaField": "$",
          "oasVersion": 2,
          "type": "schema"
        }
      }
    }
```
Seems like we might need a new rule?

philsturgeon: I suppose we will, otherwise we’ll get more complaints that we’re not validating examples.

P0lip: bah, alright.
FWIW, I aimed at supporting oas3 examples (as stated in the PR title) and not oas2, but let's get both done while we are at it. :sweat_smile:

P0lip:
@philsturgeon I added oas2-valid-example rule.

philsturgeon:
Excellent! I'm going to try this out on https://github.com/philsturgeon/examples-example and see how far we get. I'd suggest we use these as a basis for test harness.

Once we've got the ones which are meant to fail to actually fail, we can try them with $ref variants too.

philsturgeon:
Amazing work here, running this branch on from https://github.com/philsturgeon/examples-examples (the oas3.yaml) I am seeing these errors:

```
 145:30  error  oas3-valid-content-schema-example    `example` property type should be boolean
 149:30  error  oas3-valid-content-schema-example    `example` property should match format `date-time`
 153:30  error  oas3-valid-content-schema-example    `example` property should match format `url`
 176:20  error  oas3-valid-oas-parameter-example     `example` property should be equal to one of the allowed values: foo, bar
 192:22  error  oas3-valid-parameter-schema-example  `example` property should be equal to one of the allowed values: foo, bar
 204:22  error  oas3-valid-oas-parameter-example     `value` property type should be string
 206:21  error  oas3-valid-oas-parameter-example     `value` property type should be string
 269:25  error  oas3-valid-oas-content-example       `value` property should have required property `id`
 289:25  error  oas3-valid-oas-content-example       `value` property should have required property `user`
```

The first 5 are in v5.5 but the last 4 are only in this branch. The errors are not wildly useful as they are, but they are catching onto the fact that there are problems in the examples. 

>  204:22  error  oas3-valid-oas-parameter-example     `value` property type should be string
>  206:21  error  oas3-valid-oas-parameter-example     `value` property type should be string

```
        - name: multiple-examples
          description: Valid to its schema
          in: query
          schema: 
            type: string
            enum: [foo, bar]
          examples: 
            the-good:
              value: foo
            the-bad:
              value: 123                       # line 204: it's noticed this isn't a string!
            the-ugly:
              value: [an, array]            # line 206: it's noticed this isnt a string either
```

It's also noticed some response examples using the `examples` keywords:

```yaml
        "500":
          description: Fail, because the example is nonsense
          content:
            application/json:
              schema:
                properties:
                  id:
                    type: integer
                  name:
                    type: string
                  completed:
                    type: boolean
                  completed_at:
                    type: string
                    format: date-time
                required:
                  - id
                  - name
                  - completed
              examples:
                response:
                  value:
                    completed: Not a boolean
                    completed_at: Not a date
```

So that's cool.

It's still missing this:

```
      "500":
        description: Fail, because the example is nonsense
        content:
          application/json:
            schema:
              properties:
                id:
                  type: integer
                name:
                  type: string
                completed:
                  type: boolean
                completed_at:
                  type: string
                  format: date-time
              required:
                - id
                - name
                - completed
              # notice how im in the schema object, not next to it
              example:
                id: 1
                completed: Not a boolean
                completed_at: Not a date
```

I'll have a look at OAS2 too.

philsturgeon:
For oas2.yaml in that repo I'm seeing these two errors:

```
 176:30    error  oas2-valid-response-example  `application/json` property should have required property `name`
 193:30    error  oas2-valid-response-example  `application/json` property should have required property `user`
```

That's two more than the latest release! 

Similarly to oas3 feedback, if a response example is an object that sits inside a schema, it doesnt seem to get picked up. This example should fail, but doesn't:

```
        '500':
          description: 'Fail, because the example is nonsense'
          schema:
            properties: 
              id:
                type: integer
              name:
                type: string
              completed: 
                type: boolean
              completed_at:
                type: string
                format: date-time
            required:
              - id
              - name
              - completed
            # notice how im in the schema object, not next to it
            example:
              id: 1
              completed: Not a boolean
              completed_at: 'Not a date'
```

OAS2 "property examples" should also fail, but currently do not. These do work in OAS3.

```
        '501':
          description: 'Fail, because the property example is nonsense'
          schema:
            properties: 
              some-bool:
                type: boolean
                example: not a bool
              some-date:
                type: string
                format: date-time
                example: not a date
              some-url:
                type: string
                format: url
                example: not a URL
```

P0lip:
@philsturgeon I think this should be ready for another round.

P0lip:
@philsturgeon ah, seems like the same rule targets the same portion of the spec.
I'll tweak them.

P0lip:
@philsturgeon how do you feel about merging some of the rules?
Since `given` might be an array now, we do not need to have plenty of rules such as `oas2-valid-response-schema-example` or `oas2-valid-parameter-example`. What do you think?
As this is a potential breaking change, we might postpone it until 6.0

philsturgeon:
Yeah I thought we could probably get away with some rule merges so long as we can get the right function options into place. I let seems like different targets will need different function options, but if you can get the tests to pass with the changed rules then... go ahead! 

Maybe we can do the minimal change now and merge the rules as a follow up in 6.0, like you said.

philsturgeon (approved):

P0lip:
> Maybe we can do the minimal change now and merge the rules as a follow up in 6.0, like you said.

I'd need to drop a rule or two and merge them into some existing one.
I'll create a follow-up PR, so that 5.5 can be released without any breaking changes included.

PR #1301, opened 2020-08-03
feat: expose resolvedPath

P0lip:
This is somewhat related to https://github.com/stoplightio/spectral/issues/1280.

It's primary needed by Studio (diagnostics inside of forms), but I think I might actually go ahead and implement 1280 as a part of this PR.

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [ ] No


philsturgeon:
@P0lip is this going to happen or should we close it?

P0lip:
Yup, it's going to happen!

philsturgeon:
It seems like this is no longer important as it's been sat in draft a while. Closing it for now, but if you want to bring it back lmk.

P0lip:
Yeah, we're going to work on this. However, to get this right, we'll need a change of resolver, or/and, resolving process.
We haven't put much effort into this yet, so that's why it was kept open as a draft. 
I don't see much value building this functionality around the existing $ref resolver we have in Spectral, but at the same time, there's no other reasonable json $refs resolver we could use in Spectral. 
All of them share the same flaws, and therefore don't fit the Spectral use case particularly well, so swapping one with another, is likely switching from one car to another, while in reality, we need a different kind of transportation. 😆
Perhaps when we revisit $ref resolving in our own platform, we can port some ideas over to Spectral, and then we can re-open this PR.

philsturgeon:
@P0lip @mnaumanali94 I understand switching from one unideal solution to another unideal solution doesn’t seem overwhelmingly fun or worthwhile, but currently we’ve got three different ref resolvers floating around that need maintaining:

- our old thing only used by spectral
- json-schema-ref-parser
- our fork of json-schema-ref-parser

Having several broken down old cars to maintain just means spreading resources thinly across them all. If we could send one of those cars off to the scrapyard then we can focus our efforts on the remaining two. I’d rather we focused on making one excellent package, as it’s the only way out of this situation. 

Maintenance of [json-schema-reader](https://github.com/APIDevTools/json-schema-reader) is up for grabs and it would be a perfect place for us to create the best ref resolver which is fast, flexible, has the right extension points, etc. If we can make something that works for our use cases, you can imagine it would work for most folks too.

Maybe we take on finishing that tool, so we can make it our new standard, so we can delete our in house library, deprecate json-schema-ref-parser, delete our fork, and get all of our tooling (and everyone else’s) to support $id and OAS3.1 properly as a reward for doing it.

P0lip:
Well, both `@stoplight/json-schema-ref-parser` and `@stoplight/json-ref-resolver` don't require much attention from our side.
They work somewhat OK, so we don't need to touch them.
However, I totally agree that maintaining 3 or more libraries is no fun, and it's far from the point we want to be at.

The thing is that Spectral will probably have a more tailored solution rather than something generic that fits many use cases. 
This means that the code will live here, and not in some other package.
Nothing is set in the stone, though. We shall see how the situation evolves.

PR #1305, opened 2020-08-19
build(assets): $refs under functionOptions cannot be relative to the document

P0lip:
Fixes the issue with oas2/oas3-schema rules in Studio.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Why is it broken?
Some time ago we switched from dereferencing to bundling strategy when resolving $refs in rulesets during the Spectral offline assets generation.
This had one unintended side effect - as of that change all $refs are scoped relatively to the whole document.
This causes troubles, since in certain cases such has we pass a portion of the entire ruleset.
Having rewritten inline $refs inside of functionOptions is not really desired, because we always pass a given portion of the ruleset, therefore the $refs won't be resolvable, and eventually Ajv will refuse to validate the document.

```
  MissingRefError {
      message: "can't resolve reference #/rules/oas3-schema/then/functionOptions/schema/definitions/Info from id https://spec.openapis.org/oas/3.0/schema/2019-04-02#",
      missingRef: 'https://spec.openapis.org/oas/3.0/schema/2019-04-02#/rules/oas3-schema/then/functionOptions/schema/definitions/Info',
      missingSchema: 'https://spec.openapis.org/oas/3.0/schema/2019-04-02'
    }
```

The following PR makes sure we bundle only external $refs, while all inline $refs are kept untouched.

### DIFF
--- a/scripts/generate-assets.ts
+++ b/scripts/generate-assets.ts
@@ -12,7 +12,10 @@ import * as path from '@stoplight/path';
 import * as fs from 'fs';
 import { promisify } from 'util';
 import * as $RefParser from '@apidevtools/json-schema-ref-parser';
-import { KNOWN_RULESETS } from "../src/formats";
+import { KNOWN_RULESETS } from '../src/formats';
+import { Dictionary } from '@stoplight/types';
+import { isLocalRef, pointerToPath } from '@stoplight/json';
+import { get } from 'lodash';
 
 const readFileAsync = promisify(fs.readFile);
 const writeFileAsync = promisify(fs.writeFile);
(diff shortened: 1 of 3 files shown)


marbemac:
If you have not already, could you yalc into Studio and give it a thorough testing?

P0lip:
> If you have not already, could you yalc into Studio and give it a thorough testing?

I've already yalced it, but my testing was rather brief and focused primarily on the oas2/oas3-schema rules.
image

philsturgeon (approved):

PR #1317, opened 2020-08-27
chore(deps): bump @stoplight/lifecycle

wmhilton:
related to https://github.com/stoplightio/platform-internal/pull/4132

Right now I'm using a resolution to force 2.3.1 but that would break Spectral since it relied on the `createEventEmitter` hack.

**Checklist**

- [x] Tests added / updated NA
- [x] Docs added / updated NA

- [ ] Yes
- [x] No

### DIFF
--- a/src/runner/runtime.ts
+++ b/src/runner/runtime.ts
@@ -34,6 +34,7 @@ export class RunnerRuntime extends EventEmitter<SpectralEvents> {
   public spawn(): Pick<RunnerRuntime, 'on'> {
     return this.persist(
       Object.freeze({
+        // eslint-disable-next-line @typescript-eslint/unbound-method
         on: this.hijackDisposable(this.on),
       }),
     );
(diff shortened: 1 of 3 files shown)


src/runner/runtime.ts
```diff
@@ -34,6 +34,7 @@ export class RunnerRuntime extends createEventEmitter<SpectralEvents>() {
   public spawn(): Pick<RunnerRuntime, 'on'> {
     return this.persist(
       Object.freeze({
+        // eslint-disable-next-line @typescript-eslint/unbound-method
         on: this.hijackDisposable(this.on),
```
wmhilton, line 38: On my machine, eslint was throwing an `@typescript-eslint/unbound-method` error. So I added this, since I assume you know what you're doing and are being intentional.

wmhilton:
Out of curiosity, I put together a couple examples showing why the new typings are better...

Here's the errors you'd get with the old typings:
image
image


And here's what you get with the new typings:
image
image

P0lip (approved):

wmhilton:
> Right now I'm using a resolution to force 2.3.1 but that would break Spectral since it relied on the `createEventEmitter` hack.

As discussed with @P0lip on Slack, since Spectral doesn't expose its EventEmitter, we should be fine to remove the resolution and just let Studio install both versions.

PR #1330, opened 2020-09-10
Removed wrong example for `casing`

hpatoio:
`casing` is not good to check paths. See #1327

**Checklist**

- [ ] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No

**Additional context**
I think is a good idea to remote the example because `casing` is not good to check `paths`. See #1327 

Since we agreed that `casing` is not good to validate paths, does it makes sense to have `separator.allowLeading` ? I can'ts think to a good use case.


philsturgeon (approved):

philsturgeon:
This is a good change but let's leave allowLeading in there for now. Properties that are camelCase but can also have _ which can maybe even go before like _FooBar_BazBan are a use case we can support, as awkward and gross as that is.

PR #1371, opened 2020-10-01
feat(cli): introduce --fail-on-unmatched-globs flag

P0lip:
Closes https://github.com/stoplightio/spectral/issues/1117

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


docs/guides/2-cli.md
```diff
@@ -35,6 +35,7 @@ Other options include:
                                                                            [boolean] [default: false]
   --ignore-unknown-format      do not warn about unmatched formats         [boolean] [default: false]
   --show-unmatched-globs       show unmatched glob patterns                [boolean] [default: false]
+  --fail-on-unmatched-globs    fail on unmatched glob patterns             [boolean] [default: false]
```
philsturgeon, line 38: Should we deprecate show-unmatched-globs and have this be the new "more better" way of doing it? 

Correct me if im wrong but is it useful to have a single result show up saying couldnt find the files, instead of having the CLI explain it couldnt find the files?

P0lip: > Should we deprecate show-unmatched-globs and have this be the new "more better" way of doing it?

Yeah, that'd be a good idea.
By the way, ESLint exits with non-zero code by default if any of the provided glob patterns is not matched.
Example:
```
$ eslint --cache --cache-location .cache/ --ext=.js,.ts src test-harness b

Oops! Something went wrong! :(

ESLint: 6.8.0.

No files matching the pattern "b" were found.
Please check for typing mistakes in the pattern.

error Command failed with exit code 2.
```
IMHO Spectral could do the same.

philsturgeon:
Ok let’s do things one step at a time. Deprecate old flag and introduce this flag, then remove the old way in the next major, then flip new behavior default on in the major after that. OSS! 😅

--
Phil Sturgeon
DevRel @ Stoplight.io

philsturgeon (approved):

PR #1375, opened 2020-10-05
 fix(cli): JUnit formatter respects fail-severity 

### LINKED ISSUE #935: Warnings are not being shown in junit output
> **Update from Stoplight/Phil:** Let's go with the "Convert anything that meets the fail-severity to an error in the case of JUnit" option. It makes sense.

We use the junit output to show errors in Gitlab's Merge Requests, and it's worked great until now.  After upgrading to v5, we were trying to setup our builds to fail when spectral catches an `error`, but would also like to show warnings.  However, it appears that the `junit` output doesn't display warnings, but other formatters do (checked with `stylish` and `json`, but haven't rigorously tested all the others).

With `stylish` formatter, warnings show:
image

With `junit` formatter, no warnings shown:
image

I'd be happy to try and cook up an example if it's needed, but my hope is that it was just either a) an oversight on our part, b) something that's easily spotted to fix or c) intentional but undocumented behavior.

Let me know if you need more information!
Thanks!

P0lip:
Fixes #935 and the lack of choices as reported in https://github.com/stoplightio/spectral/issues/935#issuecomment-698332260 by @fgabolde

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


philsturgeon (approved):
Looks good to me but please check with the folks reporting the issue so see if it works for them.

Issue #1359, opened 2020-09-30
Take x-nullable into consideration for oas2 valid examples


igas:
**Describe the bug**

`oas2-valid-response-schema-example` does not consider `x-nullable`

**To Reproduce**

1. clone https://github.com/igas/examples-examples
2. `spectral lint oas2.yaml`
3. See error `63:24    error  oas2-valid-response-schema-example  `nickname` property type should be string`

**Expected behavior**
Because nickname has x-nullable I expect it not to raise error

**Additional context**
I believe it's related to #1284 as it started failing with 5.5.0 upgrade attempt

igas:
Thanks @P0lip!

corroded:
@P0lip Curious why this was reverted in master? https://github.com/stoplightio/spectral/commit/621c02505e1178a4aef7817c138ac45e7a88287d

P0lip:
@corroded Hey. I had to revert it since patch-package didn't work too well with npm installations of Spectral.
Binaries worked just fine. I'll revisit it soon - I'm hoping to get it out in one of the next releases.

PR #1376, opened 2020-10-05
fix(typed-enum): respect nullable in case of OAS3 document

### LINKED ISSUE #1353: Nullable enum warning for null value in example section
I receive warning when trying to use null value in example section for property with enum.

**To Reproduce**

1. Given this OpenAPI/AsyncAPI document:
[nullable_enum.yaml](https://github.com/stoplightio/spectral/files/5274847/nullable_enum.txt)

2. Run this CLI command:
> spectral lint -v nullable_enum.yaml

3. See error
> Found 82 rules (66 enabled)
> Linting nullable_enum.yaml
> OpenAPI 3.x detected
> 
> nullable_enum.yaml
>  39:23  warning  typed-enum  Enum value `null` does not respect the specified type `string`.
> ✖ 1 problem (0 errors, 1 warning, 0 infos, 0 hints)
> 

**Expected behavior**
No warnings produced

**Environment:**
 - Library version: 5.5.0
 - OS: Debian 9 (image node:14.11-slim from Docker Hub)

**Additional context**
FROM node:14.11-slim
RUN yarn global add @stoplight/spectral
ADD nullable_enum.yaml nullable_enum.yaml
RUN spectral lint -v nullable_enum.yaml

P0lip:
Fixes #1353.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


philsturgeon:
I'm not sure I understand this implementation. It seems like OAS2 says no this is not ok (because it doesnt support nullable) and OAS3 says "yeah this is fine" because it does support nullable, and that enum is accepting null as a deviation to "type: string" because nullable is there... 

I dont think this is the right solution.

I think the enum should not have `null` in it, that seems more like user error. The enum constricts the strings that can be used as a value, and nullable makes null be another value that could be used, but that doesn't mean null is now a valid enum value. It can't be, because everything in that enum should be a string...

P0lip:
> I'm not sure I understand this implementation. It seems like OAS2 says no this is not ok (because it doesnt support nullable) and OAS3 says "yeah this is fine" because it does support nullable, and that enum is accepting null as a deviation to "type: string" because nullable is there.

This is what we do in case of `schema` function.

> I think the enum should not have null in it, that seems more like user error

Why so? AFAIK `type` cannot be an array in case of OAS3, therefore user has to rely on `nullable` as there is no way to specify `"type": ["null", "string"]`.

I'm not sure how `nullable` works in case of `enum`, however, so I might be wrong.

philsturgeon:
I've read a whole lot of back and forth on this issue but it seems like what you've done is correct, but keep in mind this isn't anything to do with type arrays. 

image

From here: https://github.com/OAI/OpenAPI-Specification/blob/master/proposals/003_Clarify-Nullable.md

It seems like `nullable: true` will only work for a `type: string` with an enum if null is specified as an option. Who knew?! 

If that is what this code is doing, please carry on, but the OAS2 example should maybe use x-nullable as thats a very common vendor extension in OAS2. OAS2 is barely usable without it.

philsturgeon (approved):

philsturgeon:
We can go ahead with this. OAS2 users with `x-nullable: true` might notice trouble here, but it might not be a thing.

P0lip:
> If that is what this code is doing, please carry on, but the OAS2 example should maybe use x-nullable as thats a very common vendor extension in OAS2. OAS2 is barely usable without it.

Ah yeah, I think we actually even have a bug report for it. 
https://github.com/stoplightio/spectral/issues/1359
I'll fix it in a separate PR, since I'll need to touch other parts of code.

PR #1378, opened 2020-10-08
feat(schema): supports x-nullable for OAS2

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1359

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/patches/ajv+6.12.5.patch
+++ b/patches/ajv+6.12.5.patch
@@ -1,0 +1,25 @@
+diff --git a/node_modules/ajv/lib/ajv.d.ts b/ajv.d.ts
+index cc2881b3..a5ef48ce 100644
+--- a/node_modules/ajv/lib/ajv.d.ts
++++ b/node_modules/ajv/lib/ajv.d.ts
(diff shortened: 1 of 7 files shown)


patches/ajv+6.12.5.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/node_modules/ajv/lib/ajv.d.ts b/ajv.d.ts
+index cc2881b3..a5ef48ce 100644
+--- a/node_modules/ajv/lib/ajv.d.ts
```
P0lip, line 3: patch-package was recommended to me by @wmhilton.
He uses it in isomorphic-git with success, so decided to give it a try.
IMHO it made much more sense to apply these 2 small changes to the code of ajv rather than trying some nasty workarounds in our code.

I attempted to register a custom keyword, but unfortunately it doesn't seem that implementing a feature such as nullable is possible in a reasonable way.

philsturgeon (approved):

PR #1382, opened 2020-10-23
feat: Add JSON path to stylish formatter

aburgel:
When running the linter against a generated OpenAPI document, line numbers aren't all that helpful in seeing where the source of an issue is coming from. In my case, we generate the doc in a CI build, so don't easily have access to the generated file. We do have access to the CI output, and having a JSON path means we can see what path is producing the error.

So this PR adds the JSON path to the stylish formatter makes it easier to see the cause. 

**Checklist**

- [X] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [X] No

**Screenshots**
Screen Shot 2020-10-23 at 3 57 01 PM


aburgel:
Just realized I didn't update all the scenario tests. There's a lot, so I'll hold off on that until there's some indication that this PR may be merged.

P0lip:
@philsturgeon what do you think about this one?

philsturgeon:
Do we want `,` in there? Is this just javascript turning an array into a string and defaulting to using , or is it intentional?

aburgel:
> Do we want , in there? Is this just javascript turning an array into a string and defaulting to using , or is it intentional?

Yeah, it looks like it's an array getting turned into a string. I'm using `IRuleResult.path` which is of type `JsonPath` from `@stoplight/types`, which is an array of `string | number`.

Looks like there's [a helper function that I can use: `printPath`](https://github.com/stoplightio/spectral/blob/7f110eb59513b027f7e8a24a07806e7eda6c0338/src/utils/printPath.ts#L41). Any preference on the `PrintStyle` to use?

aburgel:
@philsturgeon @P0lip updated to use `printPath`

philsturgeon (approved):

src/cli/formatters/stylish.ts
```diff
@@ -69,6 +70,7 @@ export const stylish: Formatter = results => {
       getMessageType(result.severity),
       result.code ?? '',
       result.message,
+      printPath(result.path, PrintStyle.Dot),
```
P0lip, line 73: This style is fine, I think. :+1:

P0lip (approved):

aburgel:
@philsturgeon @P0lip what's the release process for spectral? i'm realizing now that this got merged into the develop branch, so it didn't get into the most recent release. when does develop get promoted to master and shipped?

i'd love to start using this feature.

P0lip:
@aburgel The change you implemented will be included in the 5.8.0 version. I'm planning to get it out before the end of 2020.

P0lip:
@aburgel it's in! Enjoy.

Issue #1402, opened 2020-11-09
Linter is missing non existing enums in the default


savage-alex:
Spectral 5.6.0

Given the following schema property
```yaml
property:
          type: string
          enum: [s, b, a, c]
          description: 'example schema for bug report'
          default: ''
          example: 'a'
```

Note that the validation on examples works great but no validation happens on default so I can leave it empty or put anything inside without spectral noticing.

P0lip:
Hey!
I'm on it. I'm not sure yet whether a new rule will be introduced for this one or not, but regardless of the choice, Spectral will be able to validate some of the patterns you mentioned.

P0lip:
The PR is up.

PR #1429, opened 2020-12-29
feat(oas): validate 'default' in schemas

P0lip:
Closes https://github.com/stoplightio/spectral/issues/1402

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

I'm very torn when it comes to the solution chosen here.

Besides what's implemented here, I considered the following:
- 2 completely new rules called `oas2-valid-schema-default` and `oas3-valid-schema-default` - this might cause redundant fragmentation
- renaming existing rules to something more generic, i.e. `oas2-valid-schema` (wouldn't it be confusing, though?) - this one would be breaking and therefore defer the release of the fix

I still think that ideally we should rename the rules, so that we can validate more things in the future using the same rule.
I'm not fully certain, however, so hm... @philsturgeon ideas?


philsturgeon (approved):
I think its ok to have it in default for now, its really similar. Only example and default work this way, so making a more generic oasX-valid-schema would just end up causing future bloat I think. If people report being confused by this we can split the rules but there's far bigger fish to fry I think.

PR #1435, opened 2020-12-31
Add a note about Desktop Studio not recognizing global tags

jpswade:
See https://github.com/stoplightio/studio/issues/347

### DIFF
--- a/docs/reference/openapi-rules.md
+++ b/docs/reference/openapi-rules.md
@@ -314,6 +314,8 @@ Operation should have non-empty `tags` array.
 
 Operation tags should be defined in global tags.
 
+Note: [Desktop Studio does not recognize global tags](https://github.com/stoplightio/studio/issues/347)
+
 **Recommended:** Yes
 
 ### path-declarations-must-exist


philsturgeon:
@jpswade thanks for the PR but that's not quite what I was trying to say. I was hoping for some basic "this is what tag descriptions are, why they're useful, and this is how they can be added with YAML" and then anyone ending up there can switch to code view and solve the problem

> Tags are used by various tools to group operations together. For example, documentation tools like Stoplight Docs and ReDoc group endpoints together in the navigation bar to make things easier to find. Adding descriptions to these tags is a great way to explain what these concepts are. For example, the Strava API has "Streams" but never explains what Streams are. Instead of explaining them in every single endpoint, you could explain them in the Streams tag description.

```yaml
tags:
  - name: Streams
     description: |
       A short or long explanation of what the _heck_ Streams are, with 
       any [CommonMark](https://commonmark.org/) syntax you like.
```

P0lip:
FWIW, I'm working on adding support for global tags in Studio.
If everything goes smoothly, the change should be out by the end of this week.

PR #1465, opened 2021-01-09
fix: nested extends broken in rulesets

### LINKED ISSUE #1352: Error disabling an extended ruleset in an extended ruleset
**Describe the bug**
When I structure my rules like so:

index.json
```json
{
  "extends": [
    "rules-one.json",
    "rules-two.json",
  ]
}
```

rules-one.json
```json
{
  "extends": ["spectral:oas", "off"],
  "rules": {
    "path-params": true
  }
}
```

...and then load them like so:
```ts
spectral.registerFormat("oas3", isOpenApiv3);
await spectral.loadRuleset(path.resolve("***/index.json"));
```

I get the following error:
```
    Error: Provided ruleset is not an object
        at Object.assertValidRuleset (/***/node_modules/@stoplight/spectral/dist/rulesets/validation.js:23:15)
        at processRuleset (/***/node_modules/@stoplight/spectral/dist/rulesets/reader.js:62:38)
        at process._tickCallback (internal/process/next_tick.js:68:7)
```

This does not occur if:
* I don't disable the `spectral:oas` rules
* I extend and disable them in `index.json` instead of `rules-one.json`

I also tried to extend/disable the spectral rules in `index.json` and then enable specific ones in `rules-one.json` but that didn't work (they don't get enabled).

I printed the ruleset that fails to validate and it's an IIFE that starts with:
`(function (root, factory) { if (typeof define === \"function\" && define.amd) {`

**To Reproduce**

1. Set up and load rules as described above
1. Observe error

**Expected behavior**
The rules load successfully and allow me to disable an extended ruleset, in an extended ruleset, and then selectively enable them.

 - Library version: 5.5.0
 - OS: macOS 10.15.6

**Additional context**
I organised my rules to be aligned with the OAS schema objects (e.g. `operation-object.json`) to keep things manageable. I want to enable specific `spectral:oas` rules alongside related, custom ones.

Workaround is to extend/disable in `index.json` and then enable specific rules there too.

### LINKED ISSUE #1380: Multiple extends with enabled extra rules does not work
When using multiple extends turning ON specific rules in the "base" ruleset does not work in the extended one, but turning rules OFF does.

**To Reproduce**

Have a `base.yaml` like this (where by default `tag-description` is turned off and `operation-operationId` is turned on)

```
extends: spectral:oas

rules:
  tag-description: true
  operation-operationId: off
```

and a `ruleset.yaml` like this:

```
extends: ./base.yaml
```

with the following test OpenAPI document in `api.yaml`

```
openapi: 3.0.0
servers:
  - description: Test API
    url: https://example.com/api
info:
  description: 'Test API'
  version: 1.0.0
  title: Test API
  contact:
    email: contact@example.com
  license:
    name: Apache 2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
tags:
  - name: test                # !!! note tag not having description !!!
paths:
  /test:
    get:                      # !!! note endpoint not having operationId !!!
      tags:
        - test
      summary: 'Test Operation'
      description: 'Test Operation'
      responses:
        '200':
          description: successful operation
```

When running `npx spectral lint api.yaml -r ruleset.yaml` the result is:

```
OpenAPI 3.x detected
No results with a severity of 'error' or higher found!
```

**Expected behavior**

Since in `base.yaml` the rule `tag-description` is turned ON, I'd expect the command above to fail on the missing tag description. Please note that on the other hand `operation-operationId: off` does work (the example above missed `operationId` so without explicitly turning it off the command should fail).

 - Library version: [e.g. 3.0.0] - 5.6.0
 - OS: [e.g. Windows 7] - macOS 10.15.6

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1352
Fixes https://github.com/stoplightio/spectral/issues/1380

I'm going to split this PR into 2 smaller ones.
I cleaned up all tests we had, so that they don't rely on OAS ruleset as well as moved a few files around.

Prior to that, I spent a bit of time improving the current ruleset API we have - at the moment it's not the most convenient for JS consumers.
The work can be found on [feat/nicer-rulesets](https://github.com/stoplightio/spectral/tree/feat/nicer-rulesets) branch

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


src/rule.ts
```diff
@@ -3,26 +3,26 @@ import { JSONPathExpression } from 'nimma';
...
 
-  public get enabled(): boolean {
-    return this.severity !== -1;
-  }
+  public readonly enabled: boolean;
 
-  constructor(name: string, rule: IRule) {
+  constructor(name: string, rule: IProcessedRule) {
     this.name = name;
+    this.enabled = rule.enabled !== false;
```
domagojk, line 25:
```suggestion
    this.enabled = rule.enabled ?? true;
```

seems more consistent with other rules, and IMO it's easier to read 🤷‍♂️

domagojk (dismissed):
Looks good!
I left a "code style" comment, but you choose if you want to change it :)

PR #1470, opened 2021-01-12
chore: split actual rulesets and ruleset core logic

P0lip:
Extracted from https://github.com/stoplightio/spectral/pull/1465

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

PR #1472, opened 2021-01-12
test(oas): reduce the amount of boilerplate

P0lip:
Extracted from https://github.com/stoplightio/spectral/pull/1470
Related to #1082 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


domagojk (commented):
Other than naming comment, seems good 👍 

Btw is the reason for adding props like `openapi: '3.0.2'` requried so that `isOpenApiv2`/`isOpenApiv3` can work?

P0lip:
> Btw is the reason for adding props like openapi: '3.0.2' requried so that isOpenApiv2/isOpenApiv3 can work?

Yeah, `isOpenApiv2` and `isOpenApiv3` are used to determine whether given document is of particular format.
`isOpenApiv2` looks for `swagger: 2.0.`, while the latter searches for `openapi`

domagojk (approved):

PR #1559, opened 2021-03-25
feat: js-ish rulesets

### LINKED ISSUE #1615: Future-proofing rulesets
The current design of rulesets is slightly troublesome.
The list of user-facing problems includes, but not limits itself to:

- no feature parity between CLI <-> browser usage of Spectral
- if custom functions are in use, the distribution becomes very tricky, as there's no way to generate a single file Spectral could import. This will likely become important for shared style guides
- versioning is somewhat troublesome
- they're vulnerable to breaking changes because there's no way they could lock a version of Spectral util, etc.
- difficult/impossible to use external libraries in a browser env
- not being able to specify your own format (if you're a CLI user)

The other side is us, engineers, that have quite a bit of burden to deal with:

- separate flows for CLI / browsers
- shenanigans needed to load a custom function

Moreover, current rulesets have a set of features that are very uncommonly used, but do add a notable amount of complexity:

- $refs (we do use them internally, but I doubt others do)
- `functionOptions` can be validated using a JSON schema

I could probably find more issues and flaws with the current state of things, but these are all that come to my mind at this very moment.

I'd like to propose a revision of the current design, that would address most of the issues described above, and at the same time would reduce the burden we have to deal with.

```json
{
  "rules": {
    "foo": {
      "then": {
        "function": "truthy"
      }
    }
  }
}
```

```json
{
  "extends": ["https://stoplight.com/spectral-rulesets/oas2.json"]
}
```

```js
import { isObject } from "https://cdn.jsdelivr.net/npm/lodash-es/+esm";
// obviously we could serve these from our own server / tenant server (on-prem installations)
import { truthy, schema } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-functions/+esm";
import { alphabetical } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-functions@1.0.4/+esm"; // you can stick to an older version if you want to for some reason. That's fine
import { oasRuleset } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-rulesets/+esm";
import { oas2 } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-formats/+esm";

import { verifyType } from './verifyType.mjs';

const $SCHEMA_DRAFT_2020_XX_REGEX = /^https?:\/\/json-schema.org\/draft\/2020-\d\d\/(?:hyper-)?schema#?$/;

const JSONSchemaDraft2020_XX = document => isObject(document) && "$schema" in document && $SCHEMA_DRAFT_2020_XX_REGEX.test(document.$schema);

export default {
  formats: [oas2, oas3],
  // this would no longer accept functions & functionOptions, but the rest would be almost the same (with the exception it does not support $refs, and functions have to be provided)
  extends: [oasRuleset],
  rules: {
    "valid-rule": {
      given: "$.info",
      then: {
        function: truthy, // instead of 'truthy'
      },
    },
    "only-new-json-schema": {
      formats: [JSONSchemaDraft2020_XX],
      given: "$..type",
      then: {
        function: verifyType,
      },
    }
  },
};
```

As you can see, the syntax is pretty much the same.
What's different is that you need to explicitly load formats or rulesets, but thanks to that you get control over versioning.
All the inheritance rules would be kept untouched, so `all/off/recommended`, etc. would all stay the same

Proposed file extension: `/^\.?spectral\.mjs$/`. `mjs` cause of Node.js.

##### Breaking Changes

- no custom functions in JSON/YAML rulesets
- custom functions need to be valid ESM

#### Benefits

With such an approach, any time we make a change to a ruleset or/and a function, we can quite easily roll it out to everyone.
If someone does want to stick to a particular version, it's as easy as specifying a version in the import.
Furthermore, most of the issues I listed above are solved.

#### Potential Challenges

Adjusting Studio Spectral ruleset form - I'll be able to do without any issues by working on ESTree rather than trying to dispatch JSON patches.
So basically, the flow would be:
- parse JS file with acorn
- work on the tree
- use `astring` and write contents to disk

Some migration for existing users.

#### Links

https://github.com/stoplightio/spectral/issues/845
https://github.com/stoplightio/spectral/issues/1329
https://github.com/stoplightio/spectral/issues/1561

P0lip:
Closes #1615 

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated (we still need to update it, but perhaps a bit later once the migration is a bit better fleshed out)

- [x] Yes
- [ ] No

**Additional context**

high-level overview of changes

- `src/`
  - `src/**/__tests__/**` - it's similar to changes that were applied under test-harness
  - `src/ruleset` - contains the core of this whole task
  - `src/rulesets` - rulesets rewritten to JS. Some of their tests had to be adjusted.

- `test-harness/`
  there are 3 notable changes in most of tests that a result of:
  - we no longer register a default ruleset (oas + asyncapi) therefore each test that linted something comes up with a ruleset now
  - rulesets written in YAML/JSON were rewritten to JS since we have no migrator yet
  - "XYZ spec detected" gone because we don't report it anymore


src/functions/pattern.ts
```diff
@@ -13,7 +13,9 @@ export interface IRulePatternOptions {
 // the available flags are "gimsuy" as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
 const REGEXP_PATTERN = /^\/(.+)\/([a-z]*)$/;
 
-function getFromCache(cache: Map<string, RegExp>, pattern: string): RegExp {
+const cache = new Map();
```
yordis, line 16: I have written so many freaking time the exact same code about get or set caching that I ended up publishing https://www.npmjs.com/package/@straw-hat/fancy-map

Feel free to ignore or copy the code, or use the package 😄 just funny to see it again, and again and again LOL

src/cli/services/linter/utils/getRuleset.ts
```diff
@@ -1,29 +1,28 @@
...
-async function loadRulesets(cwd: string, rulesetFiles: string[], opts: IRulesetReadOptions): Promise<IRuleset> {
-  if (rulesetFiles.length === 0) {
-    return {
-      functions: {},
-      rules: {},
-      exceptions: {},
-    };
+async function getDefaultRulesetFile(): Promise<Optional<string>> {
+  const cwd = process.cwd();
+  for (const filename of await fs.promises.readdir(cwd)) {
```
yordis, line 9: Holy shit, I didn't know about `fs.promises` #TodayILearned

src/functions/pattern.ts
P0lip: Thanks!
This cache won't stay for too long here, at least not in its current form.
I just moved it out, so that things continue to work, but I'm going to do it a bit differently.
My plan is to move it to the runtime package, buuuut prior to that I need to convert Spectral to monorepo (on our list!)

scripts/strap.sh
```diff
@@ -0,0 +1,14 @@
+mkdir -p ./node_modules/@stoplight/spectral-functions && cp -r ./dist/functions/* ./node_modules/@stoplight/spectral-functions
```
P0lip, line 1: This is temporary until monorepo is a thing.
Needed by test-harness

package.json
```diff
@@ -138,12 +130,11 @@
     "prettier": "^2.2.1",
     "recast": "^0.20.4",
     "rimraf": "^3.0.2",
-    "rollup": "^2.40.0",
```
P0lip, line 141: We used Rollup to building static assets. Static assets was a dirty workaround to serve rulesets in Studio and tests in Spectral.
Luckily, we no longer need these static assets as we're all based on CommonJS now (ES Modules soonish, as a part of monorepo).

tsconfig.json
```diff
@@ -4,7 +4,9 @@
     "baseUrl": ".",
     "paths": {
       "@stoplight/spectral-core": ["src/index.ts"],
-      "@stoplight/spectral-functions": ["src/functions/index.ts"]
+      "@stoplight/spectral-functions": ["src/functions/index.ts"],
+      "@stoplight/spectral-formats": ["src/formats/index.ts"],
```
P0lip, line 8: more packages exposed.
These paths won't be needed as soon as we have monorepo.

scripts/generate-assets.ts
```diff
@@ -1,97 +0,0 @@
-/**
```
P0lip, line 1: So glad to see this gone :heart_eyes:

src/cli/commands/lint.ts
```diff
@@ -73,10 +73,6 @@ const lintCommand: CommandModule = {
           alias: 'r',
           description: 'path/URL to a ruleset file',
           type: 'string',
-          coerce(value): Optional<string[]> {
```
P0lip, line 76: We allow only one ruleset to be specified at a time.

domagojk (approved):
LGTM 😄 

but yea it does look good. I went trough code as much as I could. It's hard to review so I missed some things for sure but 🤷‍♂️

src/cli/commands/lint.ts
philsturgeon: Does this mean `spectral lint --ruleset foo --ruleset=bar` will still work, we just removed `spectral lint --ruleset foo,bar`?

src/meta/rule.schema.json
```diff
@@ -1,18 +1,23 @@
 {
-  "$id": "http://stoplight.io/schemas/rule.schema.json",
   "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://github.com/stoplightio/spectral/blob/develop/src/meta/rule.schema.json",
```
philsturgeon, line 3: Seeing as $id does not need to resolve, could we a) remove this, or b) set it to something that we control so if people do $ref using it its stable? `http://stoplight.io/schemas/rule.schema.json` with a redirect would be ideal if we need to make it resolve for some reason.

P0lip: yeah, b) would be ideal. I'll try to see if that's possible.

philsturgeon: Also maybe namespace under spectral. `http://stoplight.io/schemas/spectral/rule.json` so other tools can define schemas.

PR #1609, opened 2021-05-10
fix(oas): misleading error for missing paths/webhooks/components

### LINKED ISSUE #1604: Misleading error for missing paths/webhooks/components
**Chore summary**
OAS 3.1 requires one of paths/webhooks/components as a top level entry. Spectral gives a `missing path object` error in case none of these is present.

**Tasks**
The error should specify that either path, webhooks or components need to be present.

P0lip:
Fixes #1604 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

We can probably vastly simplify the `schema` fn and rely on `ajv-errors`.
Something to consider for v6.


domagojk (approved):
LGTM

PR #1620, opened 2021-05-12
feat(cli): require valid commands and options

### LINKED ISSUE #1586: Running spectral with unknown flags gives exit code of 0
**Describe the bug**
When running the linter, if you pass it unknown arguments it does not run the linter and instead writes out help docs.
This behavior is fine but it's problematic because it returns an exit code of 0 - For the past 2 weeks, we erroneously thought the linter was running because it was giving a successful status code.

**To Reproduce**
`npx @stoplight/spectral lint descriptions/v1/api.yaml -rip && echo "FAIL"`

You will see the word `FAIL` printed to your console, indicating the status code was successful

**Expected behavior**
I expected spectral to return a non zero status code and my echo statement to not run

 - Library version: 5.9.1

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1586

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


paulatulis (approved):
@P0lip looks good to me! But because you know better what you're doing than I do :grimacing:, I'll also tag @domagojk @mnaumanali94 or @wmhilton if they should take a look as well.

PR #1625, opened 2021-05-12
feat: add option to pass a list of additional paths for resolving

barrett-schonefeld:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Purpose:
- Allow additional paths from which Spectral may resolve `require`s in custom functions.

Changes:
- Add a `-p` option in Spectral, so users may provide the paths through the CLI.
- Add a `paths` parameter to `loadRuleset` to add this feature to the javascript API.
- Pass paths to `evaluators.proxyRequire` to use for resolving `require`s.

Tests:
- Additional paths helps resolve file locations  when otherwise not enough information
- `-p` works and data is passed to `lint`

Additional Considerations:
- May extend so the additional paths are used to resolve files in the http-and-file-resolver.

### DIFF
--- a/docs/guides/2-cli.md
+++ b/docs/guides/2-cli.md
@@ -29,6 +29,7 @@ Other options include:
   --output, -o                 output to a file instead of stdout                                      [string]
   --resolver                   path to custom json-ref-resolver instance                               [string]
   --ruleset, -r                path/URL to a ruleset file                                              [string]
+  --paths, -p                  path from which custom functions `require` statements may be resolved   [string]
   --fail-severity, -F          results of this level or above will trigger a failure exit code
                                          [string] [choices: "error", "warn", "info", "hint"] [default: "error"]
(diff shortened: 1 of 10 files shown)


barrett-schonefeld:
Hey, folks, want to check in again here.

Any thoughts on moving forward with this PR?

philsturgeon:
@P0lip @mnaumanali94 what do you folks think?

P0lip:
I believe this may be no longer required given we support JS rulesets now.
This is one of the benefits of having that additional ruleset format - we offer the support of such cases out of the box.

philsturgeon:
That’s a really good point. I think we’d probably want to move towards bundling them up as NPM instead of floaty file system things, this didn’t work so well for PEAR in PHP world and it won’t be great for us.

I’ve got some blog posts and examples coming for JS+NPM rulesets so I’ll try and remember to post them back here.

PR #1627, opened 2021-05-12
chore(docker): build docker image in Circle CI

wmhilton:
Moves the docker image builds from Docker Hubs CI back to Circle CI, because apparently this:

image

is an [ancient issue they don't care enough to fix](https://github.com/docker/hub-feedback/issues/1777#issuecomment-835426290). I've tried everything I can think of (turning off build caches, making a new branch with a different name) and no, Docker Hub is just incapable of running "git clone" on an empty directory, as incredibly logical as that behavior would seem for generating reliable, immutable builds. And no they don't let you override the git clone step in their hooks and stuff.

This is an impediment from us releasing v6 Docker images, because presumably to release v6 we'll merge `develop` into the `master` branch. At which point DockerHub will give up and say "nope, I'm broken on master branch sorry too bad".

This is basically a reversal of https://github.com/stoplightio/spectral/pull/343 although I kept Vincenzo's dockerfile improvements.

No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,14 +153,11 @@ jobs:
             echo "GIT_SHA=${GIT_SHA}";
             # Login to DockerHub
             echo $DOCKER_PASSWORD | docker login -u $DOCKER_USER --password-stdin
-            echo "CIRCLE_TAG=${CIRCLE_TAG}";
             DOCKER_BRANCH="${CIRCLE_BRANCH//\//-}";
             echo "DOCKER_BRANCH=${DOCKER_BRANCH}";
             if [[ -n "$DOCKER_BRANCH" ]]; then
               docker tag stoplight/spectral:${GIT_SHA} stoplight/spectral:${DOCKER_BRANCH};
             fi
-            echo "Setting a fake value for CIRCLE_TAG";
-            CIRCLE_TAG="0.0.0";
             if [[ -n "$CIRCLE_TAG" ]]; then
               FULL_VERSION="${CIRCLE_TAG/v/}";
               MINOR_VERSION="${FULL_VERSION/%.+([0-9])/}";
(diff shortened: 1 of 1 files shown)


wmhilton:
Alrighty. I think I have all the kinks out except we need to replace Vincenzo's personal Dockerhub account with the company one in the CircleCI env vars.

image

.circleci/config.yml
```diff
@@ -175,6 +231,12 @@ workflows:
       - test-windows
       - test-browser
       - lint
+      - build-docker:
+         filters:
+            branches:
+              only:
+                - develop
+                - chore/can-we-build-docker-in-circle
```
P0lip, line 239:
```suggestion
```

wmhilton: We currently build it for the `develop` branch though @P0lip

P0lip: Hmm, do we need to, tho? I'd say we should only run it when a given commit is tagged appropriately

P0lip: at least line 239 should be removed :laughing:

wmhilton: I mean, honestly though line no 239 is totally harmless and there's no reason to remove it. :D

Buuuut yeah honestly I'm OK with no branch builds either. The only question then is do we always tag every new release with `latest` instead of doing a separate branch build on `master` or `main` for that.

wmhilton: The more I think about it, the more inclined I am to simply remove the `latest` tag. :/ No one in their right mind would use it anyway since it jumps across major versions.

wmhilton:
screenshot of it working on the previous commit, before I removed the branches filter in the config so now it'll only build on tags.

image

P0lip (approved):
Thanks mate!

PR #1633, opened 2021-05-21
chore: drop schemaPath function

### LINKED ISSUE #1621: Drop `schemaPath` function
`schemaPath` function relies on Spectral's internals and is used only by AsyncAPI ruleset.
 
- [ ] Remove the SchemaPath usage in AsyncAPI ruleset.
- [ ] Create a custom function if we cannot use any other spectral function to achieve the same outcome. 
- [ ] Use the custom function in the AsyncAPI ruleset.

P0lip:
Closes #1621 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


src/rulesets/asyncapi/functions/utils/isObject.ts
```diff
@@ -0,0 +1,5 @@
+import type { Dictionary } from '@stoplight/types';
+
+export function isObject(value: unknown): value is Dictionary<unknown> {
+  return value !== null && typeof value === 'object';
+}
```
domagojk, line 5: Should we also check if `!Array.isArray(value)`? Because `isObject([1,2])` would return `true` and array can't be a `Dictionary` type, right?

Also, is there some lodash equivalent we could use?

domagojk (changes requested):
One question but it looks good!

domagojk (approved):

Issue #1636, opened 2021-05-24
Spectral yells at you if there are x-extensions on operation level


mnaumanali94:
OpenAPI allows x-extensions on the operation level: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#operationObject

Spectral yells at you if x-internal or any other x-extension is added to the operation level. This shouldn't happen as it's valid OpenAPI

![image.png](https://images.zenhubusercontent.com/5ea9a7f489044245dbd23d46/b7c92a41-db5b-4b50-93e1-43238800af28)

- [ ] Fix the issue above
- [ ] Find and fix other places where x-extensions can be used but aren't validating rn.

PR #1640, opened 2021-05-25
fix(oas): unevaluated x- extensions

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1636

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

Temporary workaround until https://github.com/ajv-validator/ajv/issues/1625 is fixed.
This PR makes it so the property gets evaluated ("type" keyword), but since every type is allowed, the behavior is almost identical to a booleanish true schema.


wmhilton (approved):
LGTM

Issue #1618, opened 2021-05-12
Standardized error handling


P0lip:
Logging in Spectral is not particularly standardized now and needs an overall improvement.
At present, we can distinguish at least 3 approaches:
- using any `console.*` calls (sometimes silenced if --quiet flag is passed, but not always)
- throwing an error and not handling it too well
- on rare occasions, we push a validation result.

Here's what I propose.

For actual logging let's setup up something similar to `@stoplight/reporter`. I'm not saying we should include the project, as it's tied to Sentry and an array of other libraries, but we can incorporate the same pattern.

If something is considered to be an exception, let's not use `console.warn` or `console.error`, but simply abort the process the exception occurred in, regardless of its place - this means linting can also be canceled and result in an error.

Examples:

- https://github.com/stoplightio/spectral/blob/develop/src/runner/lintNode.ts#L30 & https://github.com/stoplightio/spectral/blob/develop/src/ruleset/readRuleset.ts#L178 - these are actual errors. Instead of logging a message, let's refuse to process the ruleset and persuade the user to address the issues.

- https://github.com/stoplightio/spectral/blob/develop/src/runner/lintNode.ts#L52, https://github.com/stoplightio/spectral/blob/develop/src/runner/lintNode.ts#L73 & https://github.com/stoplightio/spectral/blob/develop/src/runner/runner.ts#L151 - ditto, it's an error-ish scenario. In this particular case, we could potentially push a diagnostic error saying "xyz rule failed to execute", etc.

Related issues: https://github.com/stoplightio/spectral/issues/1096

PR #1644, opened 2021-05-31
feat: do not swallow runtime exceptions

P0lip:
Addresses certain chunks of #1618 

The doc is here https://github.com/stoplightio/spectral/pull/1648

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

**Additional context**

I know some tests were not located in the right spot, or could be simplified thus I went ahead and shuffled things around a bit.


src/rulesets/oas/index.json
```diff
@@ -21,7 +21,7 @@
       "description": "Operation must have at least one `2xx` or `3xx` response.",
       "recommended": true,
       "type": "style",
-      "given": "$.paths.*[?( @property === 'get' || @property === 'put' || @property === 'post' || @property === 'delete' || @property === 'options' || @property === 'head' || @property === 'patch' || @property === 'trace' )]",
+      "given": "$.paths[*][?( @property === 'get' || @property === 'put' || @property === 'post' || @property === 'delete' || @property === 'options' || @property === 'head' || @property === 'patch' || @property === 'trace' )]",
```
P0lip, line 24: `[*]` is slightly more standard

src/rulesets/oas/index.json
```diff
@@ -437,9 +437,9 @@
       "severity": 0,
       "type": "validation",
       "given": [
-        "$..definitions..[?(@property !== 'properties' && (@.example !== void 0 || @['x-example'] !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
-        "$..parameters..[?(@property !== 'properties' && (@.example !== void 0 || @['x-example'] !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
-        "$..responses..[?(@property !== 'properties' && (@.example !== void 0 || @['x-example'] !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]"
+        "$..definitions..[?(@property !== 'properties' && @ && (@.example !== void 0 || @['x-example'] !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
```
P0lip, line 440: This was actually throwing an exception at the runtime we silently ignored.
It's related to the latest jsonpath-plus change, where they started to evaluate these script filter expressions also in case of the falsy values. Previously, if the value was `null`, it wouldn't evaluate the expression entirely, thus the error would not appear.

paulatulis:
@P0lip I've read over this and the issue, but I am still not quite sure what I am looking at, so I think someone else from @stoplightio/void-crew should look as well. And if you have any extra time tomorrow in your early afternoon before standup, maybe we could jump on a quick call or talk in Slack about what is happening?

P0lip:
>  And if you have any extra time tomorrow in your early afternoon before standup, maybe we could jump on a quick call or talk in Slack about what is happening?

Sure! Let's chat

paulatulis (approved):
Approving. Thanks for walking me through @P0lip!

PR #1665, opened 2021-06-09
test(oas): use a rule tester

### LINKED ISSUE #1082: Reduce the test maintainance

P0lip:
This is most the copy-paste of #1650 with added support for fs mocks to dedupe certain tests.
Closes #1082 

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


P0lip:
I'm merging this as this doesn't involve any meaningful change anyway and the approach was validated in the other PR.

PR #1672, opened 2021-06-14
feat: improve ruleset validation

### LINKED ISSUE #1637: Improve Ruleset validation
## To Reproduce

Use `.spectral.yml`:

```yml
extends: 'spectral:oas'
rules:
  se_oas3_components_must_exist:
    type: style
    given: $
    formats:
      - oas3
    message: '{{error}}'
    then:
      field: components
      function: schema
      functionOptions:
        type: object
        minItems: 1
```

Look at validations for spectral file.

## Expected Results

I would expect one error relating to the lack of a `schema` object under `functionOptions`.

## Actual Results

image

This made fixing a simple problem a goose chase. The CLI at least returns the actual error, though a ton of extra garbage:

```bash
$ spectral lint -r ~/Downloads/se-spectral.yml ~/Documents/Stoplight\ Studio/test/test.v1.json
/rules/se_oas3_components_must_exist/then/functionOptions should NOT have additional properties
/rules/se_oas3_components_must_exist/then/functionOptions should have required property 'schema'
/rules/se_oas3_components_must_exist/then should match "then" schema
/rules/se_oas3_components_must_exist/then should be array
/rules/se_oas3_components_must_exist/then should match some schema in anyOf
/rules/se_oas3_components_must_exist should be string
/rules/se_oas3_components_must_exist should be equal to one of the allowed values
/rules/se_oas3_components_must_exist should be boolean
/rules/se_oas3_components_must_exist should be array
/rules/se_oas3_components_must_exist should match exactly one schema in oneOf
```

## Expected

Improve the ruleset validation to show better errors for spectral rulesets

P0lip:
Closes #1637 

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [x] Yes
- [ ] No

**Additional context**

A high-level list of changes:

Rulesets:
- dropped the dynamic `functionOptions` declaration you could specify on each `function` declaration or when overriding a rule - cherry-pick from #1559 
- moved the responsibility of validation of `functionsOptions` from a ruleset onto each function respectively. This is to ensure the integrity of options when the function is called by a different function. Right now we don't quite validate such scenarios which may lead to unexpected situations and cause confusion. Moreover, there's a way to validate the input too. I observed a pretty common pattern in each function where they validate the input first. I decided to add `errorOnInvalidInput` option which is not consumed yet. As a part of that, I also strengthened the typings a tiny bit. Lastly, the behavior is a bit more consistent since some functions tended to return an error message upon invalid input, while others not. Right now the default is no error and  `errorOnInvalidInput` can be set to alter the behavior.
- rewrote the validation to provide better messages that are easier to understand

Custom Functions:
- `IFunctionContext` is gone - cherry-pick from #1559 . It's expected to use functions from `@stoplight/spectral-functions` now instead of that obscure syntax.

In tests:
- tests for core functions leverage Spectral from now on meaning they no longer call core functions directly as they used to. This spares us from all these `#{{print('value)}}` messages we had to use as a result of not executing the whole linting pipeline.
- `test` -> `it` - we already use `it` pretty much everywhere, but there were places where `test` was used. We use Jasmine in Karma and it doesn't support `test`, therefore we have to mock it.
- added validation tests
- some wording here and there

Misc:
- setup basic aliases (monorepo)
- removed terser - not needed anymore (in fact, the whole rollup won't be needed as of #1559)
- some minor bugfixes that new core function tests exposed


package.json
```diff
@@ -54,7 +54,7 @@
     "pretest.karma": "node ./scripts/generate-karma-fixtures.mjs && yarn pretest",
     "pretest": "yarn generate-assets",
     "test.harness": "jest -c ./test-harness/jest.config.js",
-    "test.karma": "karma start",
+    "test.karma": "karma start || exit 0",
```
P0lip, line 57: This is temporary until #1559 is merged.

rollup.config.js
```diff
@@ -40,8 +41,8 @@ module.exports = functions.map(fn => ({
     resolve(),
     commonjs(),
     json(),
-    terser(),
   ],
+  external: id => id.startsWith('ajv') || id.startsWith('@stoplight/') || deps.includes(id),
```
P0lip, line 45: ^ added due to this particular change

src/utils/printValue.ts
```diff
@@ -11,6 +11,10 @@ export function printValue(value: unknown): string {
       return 'Array[]';
     }
 
+    if (value instanceof RegExp) {
```
P0lip, line 14: This one is an enhancement

src/runner/runner.ts
```diff
@@ -50,7 +46,7 @@ const runRule = (
         rule,
         exceptRuleByLocations,
       );
-    } else {
+    } else if (isObject(target)) {
```
P0lip, line 49: a bugfix.
I realized that `given: $` would never be executed for primitive values, which is totally not intended.
An edge case, but still.

src/rulesets/oas/functions/typedEnum.ts
```diff
@@ -1,7 +1,9 @@
-import type { IFunction, IFunctionContext } from '../../../types';
+/* eslint-disable @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-assignment */
```
P0lip, line 1: The rule doesn't quite like `@stoplight/spectral-functions` for some reason :man_shrugging:
Had worked perfectly prior to the addition of aliases

mnaumanali94 (approved):

Issue #1455, opened 2021-01-05
[Proposal] Refactor schema to accept both required and x-example parameter in all -in (header, query param, uri param etc) parameter


singhtejpal11:
This refactor will help us to allow below schema:
	- name: customerId
          in: query
          description: customer id
          required: false
          type: number
          x-example: 10		  

As of now when we try with both, we are getting below exception:
Error: schema is invalid: data/required should be array
    at Ajv.validateSchema (/*/node_modules/ajv/lib/ajv.js:178:16)
    at Ajv._addSchema (/*/node_modules/ajv/lib/ajv.js:307:10)
    at Ajv.compile (/*/node_modules/ajv/lib/ajv.js:113:24)
    at WeakMap.get (/*/dist/functions/schema.js:69:29)
    at Object.exports.schema (/*/dist/functions/schema.js:89:87)
    at Object.oasExample (eval at exports.evaluateExport (/*/dist/rulesets/evaluators.js:89:80), <anonymous>:3:1163)
    at Object.exports.lintNode (/*/dist/runner/lintNode.js:30:33)
    at callback (/*/dist/runner/runner.js:38:32)
    at JSONPath._handleCallback (/*/node_modules/jsonpath-plus/dist/index-umd.js:626:7)
    at JSONPath._trace (/*/node_modules/jsonpath-plus/dist/index-umd.js:658:12)

P0lip:
Yeah, it's a bug. Valid JSON Schema demands `required` to be an array, while we deal with `boolean` on this particular occasion.

singhtejpal11:
Thanks for the update. Could you please confirm ,what will be the next step for this? Is it going to be fixed in new release or its already planed?

P0lip:
> Is it going to be fixed in new release or its already planed?

It's going to be fixed, but unfortunately I cannot promise any specific date of delivery at this point.
That said, it's certainly something that'll be included in one of the next releases.
I'll keep you updated.

singhtejpal11:
If I put x-example inside schema, it is working fine with both required and x-example parameter. Like below:
- name: customerId
  in: query
  description: customer id
  required: false
  type: number
  schema:
    x-example: 10

Also If I put type parameter inside schema and don't make any changes with required and x-example parameter, it is still working. Like below:
- name: customerId
  in: query
  description: customer id
  required: false
  schema:
    type: number
  x-example: 10

[...]

PR #1674, opened 2021-06-16
Fixes for x-example schema issue - (issue no -1455)

singhtejpal11:
[Proposal] Refactor schema to accept both required and x-example parameter in all -in (header, query param, uri param etc) parameter

 (issue no -1455)


philsturgeon:
Thanks for this PR @singhtejpal11! Could you please provide at least one success test and one failure test showing this functionality works as intended, and what problem it was resolving in the first place?

P0lip:
#1455 has been already fixed by #1781, so I think we're good here 🎉.
Feel free to reach out in case you still experience issues.

PR #1675, opened 2021-06-16
chore: drop exceptions

P0lip:
Will reduce the noise in #1559 

Do note that although exceptions are gone, they'll be replaced with #1021 which'll be actually a superset of present exceptions, so we're not losing any functionality.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No


paulatulis (approved):
This looks good to me since it's just removing exceptions everywhere. However, if someone else from @stoplightio/void-crew wants to also take a look, I won't stop you :) 

Only thing I would add is that the documentation should also be removed from [meta](https://meta.stoplight.io/docs/spectral/docs/guides/6-exceptions.md).

P0lip:
> Only thing I would add is that the documentation should also be removed from meta.

This is going to happen automatically thanks to web hooks :)

paulatulis:
> > Only thing I would add is that the documentation should also be removed from meta.
> 
> This is going to happen automatically thanks to web hooks :)

neat!

PR #1679, opened 2021-06-18
chore: monorepo

### LINKED ISSUE #1614: Spectral v6 - monorepo
@P0lip commented on [Fri May 07 2021](https://github.com/stoplightio/platform-internal/issues/6225)

The primary goal is to slice Spectral into smaller chunks / make it more modular to ensure that introducing changes is more straightforward and less fragile.
Moreover, I'd love to make Spectral _truly_ spec agnostic, meaning we don't have any OAS/AAS exclusive code directly anywhere else than in the rulesets packages.
This work has already started some time ago when we decided to gradually rewrite tests that relied on the OAS ruleset.
While significant progress on that front has been made, there's certainly a decent amount of work left.
Oh, and we could probably forget about cherry-picking anything, and such. We'd just merge the PR, and ideally, ship it right away to customers/users.

Example layout:

- cli (similar to @stoplight/prism-cli)
- core - everything related to the linting process, spec-agnostic.
- runtime - src/runner/runtime.ts, but only the public-api part of the runner
- rulesets (src/rulesets -> packages/rulesets) - this could be a separate repo if the product team permits it.
- functions (src/functions -> packages/functions)
- formats (src/formats -> packages/formats)
- parsers (src/parsers -> packages/parsers) - it's probably worth keeping them out of the core package, because Spectral does not rely on them explicitly, and there is no need to force users to import yaml-ast-parser and jsonc-parser if they only intend to lint JSON files, for example
- resolvers? I don't know, it might be useful to have them separately, as this is something one may not need too if we make Spectral independent

Independent versioning is becoming increasingly important due to a number of issues related purely to rulesets. In such a setup, rulesets would have their own release cycle. This would eventually lead to faster iterations since a ruleset breaking change wouldn't imply a CLI / core breaking change if you know what I mean.

In terms of implementation, we can try to follow Prism, albeit with the difference that we need independent versioning.

Initially posted [here](https://github.com/stoplightio/spectral/issues/1507#issuecomment-789635794)

P0lip:
Closes #1614 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

**Additional context**

I suggest to review the PR commit by commit.
There's a bit of noise that's a result of src -> packages move, but this took place in the first commit.

I introduced 2 breaking changes I was planning to do it anyway.

Note on the release - it's still not quite set up yet.
I'm planning to add a cron job to CircleCI to release some packages with a certain cadence.
However, to accomplish that, I'll need to integrate Lerna with semantic-versioning to make sure the process is automated and we don't need to do anything manually.
Ideally, rulesets and related are released weekly (assuming we have any code in place to publish), while Core / CLI are released on demand using the approval workflow in CircleCI


domagojk (approved):
image

PR #1692, opened 2021-06-24
feat(core): support path aliases in rulesets 

### LINKED ISSUE #1680: Path aliases
**User story.**
As a ruleset write, I often have to come up with complex JSON Path expressions in order to pick a certain fragment of my document I'd like my rule to be run on.
This is both tedious and error-prone, in particular when targeting specific specs like OpenAPI.

I'd like to have a way to specify these complex path expressions in one spot to be able to re-use them later on.
Ideally, the rulesets Stoplight provides have an array of paths I could choose from, so that I don't need to author my own paths anymore (assuming I write a ruleset based on OpenAPI / AsyncAPI)

**Additional context**

This issue just intend to add the functionality to specify those aliases which can then be used in the ruleset and rulesets extending it.

Stoplight ruleset having aliases would be tackled in a separate issue.

P0lip:
Closes #1680

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/meta/ruleset.schema.json
+++ b/packages/core/src/meta/ruleset.schema.json
@@ -100,6 +100,9 @@
               },
               "parserOptions": {
                 "$ref": "#/properties/parserOptions"
+              },
+              "aliases": {
+                "$ref": "#/properties/aliases"
               }
             },
             "anyOf": [
@@ -114,8 +117,22 @@
         ]
       },
       "errorMessage": {
-        "minItems": "must not be empty",
-        "type": "must be an array"
+        "minItems": "must not be empty"
+      }
+    },
+    "aliases": {
+      "type": "object",
+      "propertyNames": {
+        "pattern": "^[A-Za-z][A-Za-z0-9_-]*$",
+        "errorMessage": {
+          "pattern": "to avoid confusion the name must match /^[A-Za-z][A-Za-z0-9_-]*$/ regular expression",
+          "minLength": "the name of an alias must not be empty"
+        }
+      },
+      "additionalProperties": {
+        "type": "string",
+        "pattern": "^[$#]",
+        "errorMessage": "the value of an alias must be a valid JSON Path expression or a reference to the existing Alias optionally paired with a JSON Path expression subset"
       }
     }
   },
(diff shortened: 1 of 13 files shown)


packages/functions/src/schema/ajv.ts
```diff
@@ -37,9 +37,18 @@ function createAjvInstance(Ajv: typeof AjvCore, allErrors: boolean): AjvCore {
 }
 
 function createAjvInstances(Ajv: typeof AjvCore): { default: AjvCore; allErrors: AjvCore } {
+  let _default: AjvCore;
```
P0lip, line 40: note for myself: extract to a separate PR.

packages/core/src/meta/ruleset.schema.json
```diff
@@ -114,8 +117,22 @@
...
       "errorMessage": {
-        "minItems": "must not be empty",
-        "type": "must be an array"
+        "minItems": "must not be empty"
+      }
+    },
+    "aliases": {
+      "type": "object",
+      "propertyNames": {
+        "pattern": "^[A-Za-z][A-Za-z0-9_-]*$",
```
wmhilton, line 126: 🤔 Pattern aliases would also be cool, so you could have a re-usable pattern called `c-style-identifier` etc.

packages/core/src/ruleset/__tests__/__fixtures__/overrides/aliases/scope.ts
```diff
@@ -0,0 +1,58 @@
+import { DiagnosticSeverity } from '@stoplight/types';
+import {falsy, pattern, truthy} from '@stoplight/spectral-functions';
+import { RulesetDefinition } from '@stoplight/spectral-core';
+
+export { ruleset as default };
```
wmhilton, line 5: wtf I did not know you could call `export` on an identifier before the identifier is declared. Makes sense, but man that's gotta be a pain on the module loader devs.

packages/core/src/ruleset/__tests__/validation.test.ts
```diff
@@ -255,6 +255,72 @@ Error at #/rules/rule/formats/1: must be a valid format`,
...
+          },
+        }),
+      ).toThrow(
+        new RulesetValidationError(
+          'Error at #/aliases: to avoid confusion the name must match /^[A-Za-z][A-Za-z0-9_-]*$/ regular expression',
+        ),
+      );
+    });
+
+    it.each(['', 'foo'])('given %s value used as an alias, throws', value => {
```
wmhilton, line 308: wait `foo` throws? what is my ☕ not revealing to my 🧠

wmhilton: Ah duh, needs a `#`.

billiegoose (dismissed):
yeah!

PR #1698, opened 2021-06-28
feat(ruleset-migrator): implement ruleset migrator 

P0lip:
Closes https://github.com/stoplightio/platform-internal/issues/6763

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
HonestWork

- [x] No

**Additional context**

Make sure to read https://github.com/stoplightio/spectral/pull/1698/commits/f90d2c8cdd404865db67d56eea093b735af65637

Tests account for ~50% of the whole diff

```
 .../services/__tests__/__fixtures__/ruleset.json   |  17 +++
 packages/cli/src/services/__tests__/linter.test.ts |  14 ++
 .../__fixtures__/apisyouwonthate/output.cjs        | 149 +++++++++++++++++++++
 .../__fixtures__/apisyouwonthate/output.mjs        | 143 ++++++++++++++++++++
 .../__fixtures__/apisyouwonthate/ruleset.yaml      | 133 ++++++++++++++++++
 .../__fixtures__/extends-variant-1/output.cjs      |   4 +
 .../__fixtures__/extends-variant-1/output.mjs      |   4 +
 .../__fixtures__/extends-variant-1/ruleset.yaml    |   1 +
 .../__fixtures__/extends-variant-2/output.cjs      |   4 +
 .../__fixtures__/extends-variant-2/output.mjs      |   4 +
 .../__fixtures__/extends-variant-2/ruleset.yaml    |   1 +
 .../__fixtures__/extends-variant-3/output.cjs      |   4 +
 .../__fixtures__/extends-variant-3/output.mjs      |   4 +
 .../__fixtures__/extends-variant-3/ruleset.yaml    |   1 +
 .../__fixtures__/extends-variant-4/output.cjs      |   4 +
 .../__fixtures__/extends-variant-4/output.mjs      |   4 +
 .../__fixtures__/extends-variant-4/ruleset.yaml    |   1 +
 .../__fixtures__/extends-variant-5/output.cjs      |   4 +
 .../__fixtures__/extends-variant-5/output.mjs      |   4 +
 .../__fixtures__/extends-variant-5/ruleset.yaml    |   1 +
 .../__fixtures__/formats-variant-1/output.cjs      |  10 ++
 .../__fixtures__/formats-variant-1/output.mjs      |   5 +
 .../__fixtures__/formats-variant-1/ruleset.yaml    |   2 +
[...]

PR #1699, opened 2021-06-29
feat(cli): demand some ruleset to be present

P0lip:
Related to #1698 

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [x] Yes
- [ ] No

**Additional context**

Bear in mind that this only relevant to the situation where you do not provide a ruleset and have no `.spectral.{js,json,yaml,yml}` at the same time.
We no longer load any default ruleset, thus it's on user now to specify some.

### DIFF
--- a/packages/cli/src/services/__tests__/linter.test.ts
+++ b/packages/cli/src/services/__tests__/linter.test.ts
@@ -94,7 +94,9 @@ describe('Linter service', () => {
 
   it('demands some ruleset to be present', () => {
     processCwdSpy.mockReturnValue(join(__dirname, '__fixtures__/resolver'));
-    return expect(run(`lint stoplight-info-document.json`)).rejects.toThrow('No ruleset has been provided');
+    return expect(run(`lint stoplight-info-document.json`)).rejects.toThrow(
+      'No ruleset has been found. Please provide a ruleset using the --ruleset CLI argument, or make sure your ruleset file matches .?spectral.(js|ya?ml|json)',
+    );
   });
 
   describe('when document is local file', () => {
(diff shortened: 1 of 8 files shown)


billiegoose (changes requested):
Can we change the error message for the CLI to say something like "Please provide a ruleset using the --ruleset CLI argument" or maybe just print the standard "Usage: " when it's used incorrectly? (So that users are told how to use it correctly instead of just told an error)

billiegoose (approved):

wmhilton:
Thanks!

dolmen:
My CI build that use the latest version of Spectral to validate OpenAPI specs, now fails because of this change.

Where can I find the upgrade guide that will tell me where I can find a generic ruleset (which was not needed before)? There is nothing helpful in the [CHANGELOG](https://github.com/stoplightio/spectral/blob/develop/CHANGELOG.md#600---2021-07-14).

P0lip:
Hey @dolmen!
It's in the readme 
image
```
echo '{\n\t"extends": ["spectral:oas"]\n}' > .spectral.json
```

and we have a more extensive guide about style guides over here https://github.com/stoplightio/spectral/blob/develop/docs/getting-started/3-rulesets.md

There's also a migration guide available https://github.com/stoplightio/spectral/blob/develop/docs/migration-guides/6.0.md

dolmen:
I did this because that's what is written in the [ruleset documentation](https://meta.stoplight.io/docs/spectral/ZG9jOjYyMDc0NA-rulesets#extending-rulesets):

```console
echo 'extends: "spectral:oas"' > .spectral.yml
```

I notice that here `extends` takes a string while you wrote above like it takes an array. Which is right? Is there a built-in ruleset to validate Spectral rulesets (edit: added #1853)?

PR #1684, opened 2021-06-22
feat(core): support overrides in rulesets

P0lip:
Closes #1021

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

**Additional context**

How to test this PR?

- clone the repo
- `npx lerna bootstrap
- `yarn build`
- `lerna run build.binary`

Then, add some rulesets, files, etc.
The binary is located under `packages/cli/binaries`.

PR #1700, opened 2021-06-29
chore(test-harness): use random names for tmp files

P0lip:
**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

We started to see some random failures as of #1684.
I had to adjust the temporary files creation logic in order to make the needed scenarios testable.
Unfortunately, the side effect of that change was that we didn't generate unique file names which at times lead to errors like these
https://app.circleci.com/pipelines/github/stoplightio/spectral/6521/workflows/c02de2fb-6117-4e43-83d0-48ff81039c64/jobs/19367
This PR addresses that issue, and while we need to use a slightly different syntax, all names are unique now which ultimately fixes the issues we started to observe.


test-harness/index.ts
```diff
@@ -6,35 +6,47 @@ import * as fs from 'fs';
 import * as tmp from 'tmp';
 import { applyReplacements, normalizeLineEndings, parseScenarioFile, tmpFile } from './helpers';
 import { spawnNode } from './spawn';
+import nanoid = require('nanoid/non-secure');
```
domagojk, line 9: why not import from? 🤔

domagojk (changes requested):

domagojk (approved):
I'm changing it to approve since the require thing is the only thing I have and auto-merge is not enabled.

So, you either used it for a reason or you can address it :)

test-harness/index.ts
P0lip: doesn't work unfortunately, would need to enable default module interop which I don't want to do.

P0lip: doesn't work unfortunately, would need to enable default module interop which I don't want to do.

PR #1709, opened 2021-07-02
feat: scheduled release

domagojk:
For https://github.com/stoplightio/platform-internal/issues/6819

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


domagojk:
@P0lip I've replaced lerna with yarn commands and created scheduled release which triggers every monday at 10am.

Two questions:
- lint job is failing in circleCi and it works locally, any idea why?
- I tried `semantic-releases` and monorepo plugin [works](https://app.circleci.com/pipelines/github/stoplightio/spectral/6586/workflows/c5ebca49-3619-430b-8224-a09f646916da/jobs/19754) but it skipped publishing because it looks for `develop` branch. Still, in order it to work we need github token and npn token env vars, do we have similar things defined anywhere else?

domagojk:
The lint error was gone after merging `develop`, and npm and github tokens are already configured in CircleCi, so all that is left is actually testing `semantic-release`. 
Running it locally seems ok (it prints out versions and release notes), but it wont publish it in dry-mode. One thing I noticed is that versions starts with `1.0.0` since every version would be a breaking one in `0.x`. Should we than follow sem release recommendations and stay with `1.x`?

Also, it looks like releasing it from a pull request is not possible so it must be merged 🤷

packages/cli/package.json
```diff
@@ -25,6 +25,8 @@
   },
   "scripts": {
     "build.binary": "pkg . --output ./binaries/spectral",
+    "build.windows": "pkg . --targets windows --out-path ./binaries",
+    "build.linuxMacAlpine": "pkg . --targets linux,macos,alpine --out-path ./binaries",
```
P0lip, line 29:
```suggestion
    "build.nix": "pkg . --targets linux,macos,alpine --out-path ./binaries",
```
how about that? :D

.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ commands:
       - run:
           name: Install Dependencies
           command: |
-            npx lerna bootstrap
+            yarn
             ! git diff --name-only | grep yarn.lock
```
P0lip, line 22:
```suggestion
            yarn --frozen-lockfile
```
I added `! git diff --name-only | grep yarn.lock` because lerna didn't have the equivalent of `frozen-lockfile`

.circleci/config.yml
```diff
@@ -203,15 +203,8 @@ jobs:
...
-          if [[ "$CIRCLE_TAG" == *beta* ]] ; then
-            npm publish --tag beta
-          elif [[ "$CIRCLE_TAG" == *alpha* ]] ; then
-            npm publish --tag alpha
-          else
-            npm publish
-          fi
-      - run: lerna exec "pkg . --targets linux,macos,alpine --out-path ./binaries" --scope=@stoplight/spectral-cli
+      - run: yarn release
+      - run: yarn workspace @stoplight/spectral-cli build.linuxMacAlpine
```
P0lip, line 207:
```suggestion
      - run: yarn workspace @stoplight/spectral-cli build.nix
```

.circleci/config.yml
```diff
@@ -273,3 +266,36 @@ workflows:
           requires:
             - release
             - build-windows-binary
+
+  scheduled:
```
P0lip, line 270:
```suggestion
  scheduled-released:
```

P0lip (commented):
Looks good!

P0lip (approved):

PR #1729, opened 2021-07-07
chore(test-harness): smarter date replacement 

P0lip:
Hopefully fixes https://app.circleci.com/pipelines/github/stoplightio/spectral/6596/workflows/10bb9be8-7603-42e5-a2db-1c48da25f1be/jobs/19804.
This one is also not particularly common, but occurs from time to time.
Since we have these automated releases now, it makes sense to address it to ensure the release is not blocked by that random failure.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/test-harness/index.ts
+++ b/test-harness/index.ts
@@ -2,6 +2,7 @@ import * as path from '@stoplight/path';
 import { normalize } from '@stoplight/path';
 import { Dictionary } from '@stoplight/types';
 import * as fg from 'fast-glob';
+import { escapeRegExp } from 'lodash';
 import * as fs from 'fs';
 import * as tmp from 'tmp';
 import { applyReplacements, normalizeLineEndings, parseScenarioFile, tmpFile } from './helpers';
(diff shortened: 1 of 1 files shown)


test-harness/index.ts
```diff
@@ -82,7 +83,19 @@ describe('cli acceptance tests', () => {
...
+      // executing Date() before or after spawnNode were constantly leading to occasional mismatches,
+      // as the success of that approach was highly bound to the time spent on the actual spawnNode call
+      // this is a tad smarter, because instead of naively hoping the date will match, we try to extract the date from the actual output
+      // this regular expression matches "00:43:59" in "Thu Jul 08 2021 00:43:59 GMT+0200 (Central European Summer Time)"
+      const date = RegExp(escapeRegExp(String(Date())).replace(/(\d\d:){2}\d\d/, '(\\d\\d:){2}\\d\\d'));
+      Reflect.defineProperty(replacements, 'date', {
+        configurable: true,
+        enumerable: true,
+        get(): string {
+          return stdout.match(date)?.[0] ?? String(Date());
```
domagojk, line 96: Date without new returns a string, right? Then I guess we can just do:

```suggestion
          return stdout.match(date)?.[0] ?? Date();
```

or leave it how it was before:

```suggestion
          return stdout.match(date)?.[0] ?? String(new Date());
```

domagojk (changes requested):
Nice! 💪

test-harness/index.ts
P0lip: True, you're right.

domagojk (approved):

PR #1736, opened 2021-07-08
fix(repo): disable husky on release

domagojk:
Disables husky when releasing with semantic-release

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "test.jest": "jest --silent",
     "test.karma": "karma start",
     "prepare": "husky install",
-    "release": "yarn workspaces run semantic-release -e semantic-release-monorepo"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "test.jest": "jest --silent",
     "test.karma": "karma start",
     "prepare": "husky install",
-    "release": "yarn workspaces run semantic-release -e semantic-release-monorepo"
+    "release": "HUSKY=0 yarn workspaces run semantic-release -e semantic-release-monorepo"
```
P0lip, line 41: do we run this command on Windows?
if so, this won't work and we may need to use `cross-env` or something similar.

P0lip (approved):
honest

PR #1756, opened 2021-07-23
fix(ruleset-migrator): invalid values for JSON Schema drafts 2019 & 2020

P0lip:
Fixes issue found by Will in https://github.com/stoplightio/platform-internal/pull/7085

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

>


packages/ruleset-migrator/src/validation/schema.ts
```diff
@@ -59,8 +59,8 @@ const schema: JSONSchema4 = {
           'json-schema-draft4',
           'json-schema-draft6',
           'json-schema-draft7',
-          'json-schema-2019-09',
-          'json-schema-2020-12',
+          'json-schema-draft-2019-09',
```
P0lip, line 62: fwiw this is a "soft" breaking change.
In v5 it's called `json-schema-2019-09`, but it was an oversight that I never really addressed back then because this would be a breaking change.
Now, it's time to make it more consistent.

P0lip: hmm or maybe we keep it as is?
seems like the authors don't use the word draft here https://json-schema.org/draft/2019-09/release-notes.html
@wmhilton thoughts?

mnaumanali94: I think we should keep it as is. @P0lip I don't think its worth the breaking change

P0lip: I can still alias the old value to that new one, so folks can use both

wmhilton (changes requested):
Handle "async-api2" as well?

P0lip:
>  "async-api2" 

Nah, this one is purely on the Studio side of things.

packages/ruleset-migrator/src/validation/schema.ts
P0lip: > I think we should keep it as is. @P0lip I don't think its worth the breaking change

The primary motivation I have here is that in JS rulesets you'd refer to these formats differently. In JS they're named "jsonSchemaDraft2019_09" and `2020_12` respectively.

AFAIK json-schema-draft-2020-12 doesn't even exist in Spectral v5, so it'd be only 2019-09 that's changed.

packages/ruleset-migrator/src/transformers/formats.ts
```diff
@@ -2,32 +2,34 @@ import { builders as b, namedTypes } from 'ast-types';
...
 function transform(ctx: TransformerCtx, input: unknown): namedTypes.ArrayExpression {
   assertArray(input);
 
   return b.arrayExpression(
-    input.map(format => {
-      assertString(format);
-      return ctx.tree.addImport(REPLACEMENTS[format], '@stoplight/spectral-formats');
-    }),
+    Array.from(
+      new Set(
```
P0lip, line 26: Although is not invalid for formats to be listed more than once, it feels nice to have them de-duped.

packages/ruleset-migrator/src/__tests__/__fixtures__/formats-variant-2/output.cjs
```diff
@@ -8,7 +8,7 @@ const {
 } = require('@stoplight/spectral-formats');
 const { truthy: truthy } = require('@stoplight/spectral-functions');
 module.exports = {
-  formats: [oas2, oas3_1, oas3_0, jsonSchemaLoose, jsonSchemaLoose],
+  formats: [oas2, oas3_1, oas3_0, jsonSchemaLoose],
```
P0lip, line 11: This is the aforementioned deduplication.

wmhilton (approved):

wmhilton:
@P0lip - Nauman is OOO today so if you think you've addressed his existing comments feel free to merge.

PR #1758, opened 2021-07-29
feat(ruleset-migrator): assert safe & unique identifiers 

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

This is just to make sure we don't generate an invalid identifier OR that we don't redeclare the const, etc.
Haven't seen such a usage yet as some of these cases are truly edge, but let's account for them to future-proof the migrator.


packages/ruleset-migrator/src/__tests__/__fixtures__/functions-variant-3/output.cjs
```diff
@@ -0,0 +1,32 @@
+const uppercase = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/upper-case.js'));
```
P0lip, line 1: prior to this PR, you'd get `const upper-case` which is obviously invalid.

packages/ruleset-migrator/src/__tests__/__fixtures__/functions-variant-3/output.cjs
```diff
@@ -0,0 +1,32 @@
...
+const no = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/no-@.js'));
```
P0lip, line 2: `no-@` here.

packages/ruleset-migrator/src/__tests__/__fixtures__/functions-variant-3/output.cjs
```diff
@@ -0,0 +1,32 @@
...
+const _400response = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/400-response.js'));
+const import$0 = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/import.js'));
+const require$0 = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/require.js'));
```
P0lip, line 5: aaand
```ts
const no-@ = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/no-@.js'));
const 400response = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/400-response.js'));
const import = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/import.js'));
const require = _interopDefault(require('/.tmp/spectral/functions-variant-3/functions/require.js')); // this, while a valid identifier, would override `require` leading to broken CJS
```

packages/ruleset-migrator/src/__tests__/__fixtures__/functions-variant-4/output.cjs
```diff
@@ -0,0 +1,18 @@
+const { oas3: oas3, oas3_1: oas3_1 } = require('@stoplight/spectral-formats');
+const oas3$0 = _interopDefault(require('/.tmp/spectral/functions-variant-4/functions/oas3.js'));
```
P0lip, line 2: before the change, you'd get
```ts
const oas3 = 
```
which presumably could lead to an error if an OAS format was used too because the output would look as follows if the aforementioned circumstance was met.

```
const { oas3: oas3 } = require('@stoplight/spectral-formats');
const oas3 = _interopDefault(require('/.tmp/spectral/functions-variant-4/functions/oas3.js'));
```

packages/ruleset-migrator/src/__tests__/__fixtures__/rules-variant-1/output.cjs
```diff
@@ -11,7 +11,7 @@ module.exports = {
       given:
         "$.paths.*[?( @property === 'get' || @property === 'put' || @property === 'post' || @property === 'delete' || @property === 'options' || @property === 'head' || @property === 'patch' || @property === 'trace' )]",
       then: {
-        function: oasOpFormDataConsumeCheck,
```
P0lip, line 14: this is correct, because `oasOpFormDataConsumeCheck` is not declared

packages/ruleset-migrator/src/tree/index.ts
```diff
@@ -91,14 +94,15 @@ export class Tree {
...
-  public identifier(name: string): namedTypes.Identifier {
-    let uniqName = name;
+  public static identifier(name: string, scope: Scope): namedTypes.Identifier {
+    const baseName = name.replace(/[^$_0-9A-Za-z]/g, '').replace(/^([0-9])/, '_$1');
+    let uniqName = baseName;
     let i = 0;
-    while (this.scope.has(uniqName)) {
-      uniqName = `${name}_${i++}`;
+    while (scope.has(uniqName)) {
+      uniqName = `${baseName}$${i++}`;
```
lott-ai, line 102: Curious, why don't we scope the entire identifier for imports? Then could just replace invalid characters instead of matching registered words.

```suggestion
      uniqName = `import${i++}_${baseName}`;
```

packages/ruleset-migrator/src/tree/scope.ts
```diff
@@ -0,0 +1,102 @@
+const REGISTERED_WORDS = [
```
lott-ai: Is there a list/package of reserved words we can pull from so we don't have to maintain it here?

lott-ai (dismissed):
Code looks good! I haven't been super involved in this project so only had a couple questions. Nothing blocking other than the failing test 👍

packages/ruleset-migrator/src/tree/scope.ts
P0lip: Seems like https://www.npmjs.com/package/reserved, will use it, looks cool.

packages/ruleset-migrator/src/tree/index.ts
P0lip: Mostly for the sake of readability.
I try to make the output as clean as possible.
Basically if one has such a function declared
```json
"functions": ["test"]
```
we want to output

```js
import test from '/functions/test.js';
```
This is all for the sake of trying to be as least destructive as we can.

PR #1771, opened 2021-08-10
fix(functions): schema should respect "auto" dialect

P0lip:
Currently when you pass "auto" we don't run `detectDialect` function, so nothing useful happens.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/functions/src/schema/index.ts
+++ b/packages/functions/src/schema/index.ts
@@ -25,11 +25,13 @@ export default createRulesetFunction<unknown, Options>(
         },
         dialect: {
           enum: ['auto', 'draft4', 'draft6', 'draft7', 'draft2019-09', 'draft2020-12'],
+          default: 'auto',
         },
         allErrors: {
           type: 'boolean',
+          default: false,
         },
-        prepareResults: {},
+        prepareResults: true,
       },
       required: ['schema'],
       type: 'object',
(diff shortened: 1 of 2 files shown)


packages/functions/src/schema/index.ts
```diff
@@ -25,11 +25,13 @@ export default createRulesetFunction<unknown, Options>(
         },
         dialect: {
           enum: ['auto', 'draft4', 'draft6', 'draft7', 'draft2019-09', 'draft2020-12'],
+          default: 'auto',
```
P0lip, line 28: unrelated, but I thought it'd be actually cool to express what the default is.
Going to be useful later on in Studio when building forms.

packages/functions/src/schema/index.ts
```diff
@@ -56,7 +58,9 @@ export default createRulesetFunction<unknown, Options>(
 
     try {
       let validator;
-      const dialect = opts?.dialect ?? detectDialect(schemaObj) ?? 'draft7';
+      const dialect =
```
P0lip, line 61: this is the only actual change

packages/functions/src/schema/index.ts
```diff
@@ -25,11 +25,13 @@ export default createRulesetFunction<unknown, Options>(
...
         dialect: {
           enum: ['auto', 'draft4', 'draft6', 'draft7', 'draft2019-09', 'draft2020-12'],
+          default: 'auto',
         },
         allErrors: {
           type: 'boolean',
+          default: false,
         },
-        prepareResults: {},
+        prepareResults: true,
```
lott-ai, line 34: I'm assuming this is the same as passing an empty object?

packages/functions/src/schema/index.ts
```diff
@@ -56,7 +58,9 @@ export default createRulesetFunction<unknown, Options>(
 
     try {
       let validator;
-      const dialect = opts?.dialect ?? detectDialect(schemaObj) ?? 'draft7';
+      const dialect =
+        (opts.dialect === void 0 || opts.dialect === 'auto' ? detectDialect(schemaObj) : opts?.dialect) ?? 'draft7';
```
lott-ai, line 62: If `opts` is not defined, it will throw on L56

image


```suggestion
        (opts.dialect === void 0 || opts.dialect === 'auto' ? detectDialect(schemaObj) : opts.dialect) ?? 'draft7';
```

lott-ai (approved):

PR #1787, opened 2021-08-24
fix(functions): schema fn should allow redundant escapes

### LINKED ISSUE #1782: Certain valid patterns are not accepted by schema function
**Describe the bug**

In some cases, Ajv bails upon a valid pattern.
It's presumably an issue with Ajv, or, with the way we use Ajv.

**To Reproduce**

1. Run the following snippet
```js
const Ajv = require('ajv');
const ajv = new Ajv();
const assert = require('assert');

const schema = {
  type: 'string',
  pattern: '^[A-Za-z0-9\\-\\_]',
};

assert.ok(ajv.validate(schema, 'test-case-1'));
```
2. Exits with non-zero code

**Expected behavior**

Exits with a zero-code

Related to https://github.com/stoplightio/platform-internal/issues/7541

P0lip:
Fixes #1782 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

This is not a bug in Ajv or anything, it's just how the unicode flag works.

### DIFF
--- a/packages/functions/src/schema/ajv.ts
+++ b/packages/functions/src/schema/ajv.ts
@@ -23,7 +23,15 @@ const logger = {
 };
 
 function createAjvInstance(Ajv: typeof AjvCore, allErrors: boolean): AjvCore {
-  const ajv = new Ajv({ allErrors, meta: true, messages: true, strict: false, allowUnionTypes: true, logger });
+  const ajv = new Ajv({
+    allErrors,
+    meta: true,
+    messages: true,
+    strict: false,
+    allowUnionTypes: true,
+    logger,
+    unicodeRegExp: false,
+  });
   addFormats(ajv);
   if (allErrors) {
     ajvErrors(ajv);
--- a/packages/functions/src/__tests__/schema.test.ts
+++ b/packages/functions/src/__tests__/schema.test.ts
@@ -42,6 +42,34 @@ describe('Core Functions / Schema', () => {
     expect(await runSchema(2, { schema, dialect: 'auto' })).toStrictEqual(result);
   });
 
+  it('allows redundant escapes', async () => {
+    let schema = {
+      $schema: 'http://json-schema.org/draft-07/schema#',
+      type: 'string',
+      pattern: '[\\-_]', // this one is good and would work with unicodeRegExp: true
+    };
+
+    expect(await runSchema(2, { schema })).toEqual([
+      {
+        message: 'Value type must be string',
+        path: [],
+      },
+    ]);
+
+    schema = {
+      $schema: 'http://json-schema.org/draft-06/schema#',
+      type: 'string',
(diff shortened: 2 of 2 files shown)


packages/functions/src/schema/ajv.ts
```diff
@@ -23,7 +23,15 @@ const logger = {
...
 function createAjvInstance(Ajv: typeof AjvCore, allErrors: boolean): AjvCore {
-  const ajv = new Ajv({ allErrors, meta: true, messages: true, strict: false, allowUnionTypes: true, logger });
+  const ajv = new Ajv({
+    allErrors,
+    meta: true,
+    messages: true,
+    strict: false,
+    allowUnionTypes: true,
+    logger,
+    unicodeRegExp: false,
```
mmiask, line 33: Are you sure we want to remove Unicode-related features? This definitely fixes the issue, but introduces some limitation to the shape of regex (according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode#description))

WDYT about somehow passing `String.raw`$stringliteral`` instead of `string` when we create `new RegExp()`? (although this might require pushing changes to `Ajv` itself)

Or just changing `\\` to `'\\\\` whenever we see `\\` in a string literal pattern?

P0lip: > WDYT about somehow passing String.raw$stringliteral`` instead of string when we create `new RegExp()`? (although this might require pushing changes to `Ajv` itself)
> Or just changing \\ to '\\\\ whenever we see \\ in a string literal pattern?

That won't fix the issue, as this changes the semantics of the regular expression.
Replacing`\\` with `\\\\` leads to a different regular expression.
For instance

```ts
new RegExp('[\\_-]').test('_'); // true
new RegExp('[\\_-]').test('\\'); // false
new RegExp('[\\\\_-]').test('\\'); // true
```

In a nutshell, the problem here is that when a `u` (unicode) flag is supplied, the behavior of `\` changes a bit, and it's up to a end user to account for that.
Unfortunately, this means that the only choices we're left here include
a) tell users to write more proper regular expressions
b) get rid of redundant escapes - very tricky do to right
c) switch to non-unicode regexps

In other words, there's no good solution, each of them has some trade-offs.
Luckily, I haven't really seen much usage of unicode, the vast majority of users doesn't use it at all. The the spec itself [does not](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-6.4
) enforce the unicode support, it recommends it.

mmiask: @P0lip you convinced me. We'll always have to pick lesser evil

mmiask (approved):

Issue #1796, opened 2021-08-31
Running `spectral` 6.0.0 command without ` --ruleset` CLI option fails


DavidBiesack:
**Describe the bug**

Using 6.0.0 installed from the installation page, running lint returns an error `No ruleset has been found`

**To Reproduce**
```bash
# Assuming user has write access to /usr/local/bin ....
$ curl -L https://raw.github.com/stoplightio/spectral/master/scripts/install.sh | sh
$ spectral --version
6.0.0
$ head -1 openapi.yaml
openapi: 3.1.0
$ type spectral
spectral is a tracked alias for /usr/local/bin/spectral
$ spectral lint openapi.yaml
spectral lint openapi.yaml
No ruleset has been found. Please provide a ruleset using the --ruleset CLI argument, or make sure your ruleset file matches .?spectral.(js|ya?ml|json)
```

**Expected behavior**

As per the [doc page](https://meta.stoplight.io/docs/spectral/ZG9jOjI1MTg1-spectral-cli#using-a-ruleset-file) , use the default ruleset (i.e. `spectral:oas` in this case)

> **Using a Ruleset File**
>If you don't specify a ruleset file with the --ruleset parameter, the Spectral CLI will look for a ruleset file called .spectral.yml, .spectral.yaml, or .spectral.json in the current working directory. If no ruleset is specified and no default ruleset file is found, the built-in rulesets will be used.

**Environment:**

 - spectral CLI 6.0.0
 - OS: Mac OS

P0lip:
Hey!
That is expected - I updated the documentation to reflect the change in behavior.

DavidBiesack:
Ouch! It is _expected_ to make a command line option required where one was not required before?
As a user, I don't think that is very user friendly.  I respectfully ask that you revert the decision to make this spectral's expected behavior.

Also note that the CLI help does _not_ document this as required either, just `spectral lint [documents..]` :

```bash
$ /usr/local/bin/spectral lint --help
spectral lint [documents..]

lint JSON/YAML documents from files or URLs

Positionals:
  documents  Location of JSON/YAML documents. Can be either a file, a glob or fetchable resource(s) on the web.                                                                                  [array] [default: []]

Options:
      --version                  Show version number                                                                                                                                                         [boolean]
      --help                     Show help                                                                                                                                                                   [boolean]
  [...]
  -r, --ruleset                  path/URL to a ruleset file                                                                                                                                                   [string]
...
```

[...]

PR #1797, opened 2021-08-31
docs(repo): update rulesets loading procedure

P0lip:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


docs/guides/2-cli.md
```diff
@@ -44,7 +44,8 @@ The Spectral CLI supports loading documents as YAML or JSON, and validation of O
 
 ## Using a Ruleset File
 
-If you don't specify a ruleset file with the `--ruleset` parameter, the Spectral CLI will look for a ruleset file called `.spectral.yml`, `.spectral.yaml`, or `.spectral.json` in the current working directory. If no ruleset is specified and no default ruleset file is found, the built-in rulesets will be used.
+If you don't specify a ruleset file with the `--ruleset` parameter, the Spectral CLI will look for a ruleset file called `.spectral.yml`, `.spectral.yaml`, `.spectral.json` or `.spectral.js` in the current working directory.
+Spectral will refuse to lint the document if no ruleset is specified and no default ruleset file is found.
```
philsturgeon, line 48: If this is meant to be a new paragraph it will need another line.

P0lip: We can have a separate paragraph, sure.
```suggestion

Spectral will refuse to lint the document if no ruleset is specified and no default ruleset file is found.
If you want to have a simple basic ruleset for OpenAPI, this command will let you get started:
`echo "extends: spectral:oas" > .spectral.yaml` 
```

@philsturgeon thoughts?

philsturgeon: Is this new behavior in Spectral v6? We used to autodetect and apply OAS if we thought it was OAS, but now folks have to specify that for theirselves?

dijitali: I just ran up against this change and found this confusing behaviour.  A few other people in https://github.com/stoplightio/spectral/issues/1796 had a similar experience from the sound of things.

P0lip: Ooops, I lost sight of these comments.

> Is this new behavior in Spectral v6? We used to autodetect and apply OAS if we thought it was OAS, but now folks have to specify that for theirselves?

Yeah, it's a deliberate change we've made.
Overall, from a technical standpoint it is quite trivial to load some default ruleset - would probably be like a one-liner kind of a change if we were to make it, so totally straightforward to do.
Our goal was to convince users to write their own style guides, but I'm all in for constant reiteration of the standpoint if this happens to be truly annoying - I believe that most of folks do have some kind of a ruleset, so this change is  seamless for them.

philsturgeon: Ok, thanks for clarifying.

mnaumanali94 (commented):
Screenshot 2021-09-15 at 9 19 44 PM

Lets add a note here too please.

mnaumanali94 (approved):

README.md
```diff
@@ -31,6 +31,18 @@ yarn global add @stoplight/spectral-cli
...
+Spectral, being a generic YAML/JSON linter, needs a ruleset in order to be able to lint files.
+You can learn more about the rulesets [here](./docs/getting-started/3-rulesets.md).
+
+If you intend to lint an OpenAPI or AsyncAPI document, we have a few predefined rulesets you can extend to get Spectral up and running.
+To reference them, you can run the following command:
+
+```bash
+echo '{\n\t"extends": ["spectral:oas", "spectral:asyncapi"]\n}' > .spectral.json
+```
+
```
mnaumanali94, line 45: Also add a callout to say that this is a breaking change moving to V6

PR #1819, opened 2021-09-06
feat(ruleset-bundler): initial release

P0lip:
Closes https://github.com/stoplightio/platform-internal/issues/7847

Here's the integration of it -> https://github.com/stoplightio/spectral/pull/1824

I'd suggest reviewing this in some editor - the set of changes is relatively small, but tests add noise and make it a bit difficult to follow the actual code in the UI.

It's pretty much a refactored copy of https://github.com/stoplightio/platform-internal/tree/v2/packages/studio/src/stores/spectral/service/worker/rollup.

I sliced the [resolver](https://github.com/stoplightio/platform-internal/blob/v2/packages/studio/src/stores/spectral/service/worker/rollup/plugins/resolver.ts) plugin into smaller chunks

Certain aspects have been slightly adjusted, however, namely:

- https://github.com/stoplightio/platform-internal/blob/v2/packages/studio/src/stores/spectral/service/worker/rollup/index.ts#L47 is now a standalone plugin -> packages/ruleset-bundler/src/plugins/stdin.ts
- builtins plugin replaced `createGlobalModules`, it's quite similar in its behavior, though the implementation is a tad different.

I also ensured we cover more cases, as this code is now supposed to be executed in Node.js as well, meaning certain transformations are redundant.

I also added a few presets which come handy depending on the environment you execute your code in.
Studio / Shared Style Guides API can make use of `browser` to make sure external imports are retained, while Spectral CLI can make use of the fact it has access to FS.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


packages/ruleset-bundler/README.md
```diff
@@ -0,0 +1,15 @@
+# @stoplight/spectral-ruleset-bundler
+
+**WARNING** - for the time being, the following package is meant to be used internally.
```
marbemac, line 3: Any reason to actually publish this then? Versus just not treat it as external dep and bundle it into the other spectral packages that use it in this workspace?

packages/ruleset-bundler/package.json
```diff
@@ -0,0 +1,52 @@
...
+      "default": "./dist/plugins/*.js"
+    }
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/stoplightio/spectral.git"
+  },
+  "dependencies": {
+    "@stoplight/path": "1.3.2",
+    "@stoplight/spectral-core": ">=1",
```
marbemac, line 34: Should we at least peg it to a major? Or is there some reason it's OK in this case to use such a wide version range?

packages/ruleset-bundler/README.md
marbemac: OH, I see I think - do we need to be able to use this package from our ecosystem standalone?

marbemac (approved):
Looks great, nice work! One small concern re the dep ranges.

packages/ruleset-bundler/README.md
P0lip: > OH, I see I think - do we need to be able to use this package from our ecosystem standalone?

yeah, we're going to use this in platform-internal (styleguides service, Studio) & spectral itself

packages/ruleset-bundler/package.json
P0lip: We don't consume the API of these packages, so it doesn't matter.
The only package we do rely on is `spectral-runtime` and this one is indeed pegged.
These packages are only here for the sake of https://github.com/stoplightio/spectral/pull/1819/files#diff-8938b40bb5a35f8551f5d19ba496c77a0e75c44e23c3992d3a8c869ee51b5d1aR1.
Technically, I could make them peerDependencies or something, but I'm not a huge fan of peerDependencies.

PR #1830, opened 2021-09-14
chore(core): remove inline-version.mjs script

P0lip:
`SPECTRAL_PKG_VERSION` is no longer needed.
It was essential when we loaded rulesets in a different way, but now it's redundant.
This PR removes the script that inserted the correct version after we built Spectral.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


domagojk (approved):
image

PR #1839, opened 2021-09-20
chore(core): validate scoped aliases

P0lip:
Adds the validation piece of https://github.com/stoplightio/spectral/issues/1837

Thinking ahead I rewrote some $ids to have less work when tackling https://github.com/stoplightio/spectral/issues/1835

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

PR #1838, opened 2021-09-19
feat(ruleset-migrator): support scoped aliases

P0lip:
For https://github.com/stoplightio/spectral/issues/1837

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

PR #1840, opened 2021-09-20
feat(core): support scoped aliases

### LINKED ISSUE #1837: Change Aliases to accept targets for multiple formats and description
**User story.**
As a spectral user, I can specify targets for multiple formats and add a description. This would allow users to use the same alias for multiple formats i.e. OAS3, AsyncAPI e.t.c

Aliases today accept a single target which results in duplication for the various supported formats. 

Change aliases to a structure like: 

```jsonc
{
  "aliases": {
    // key is name - just guessing based on rest of spectral specification convention
    "schema": {
      "description": "Foo",
      "targets": [
        {
          "formats": ["oas2"],
          "given": "$.definitions.*"
        },
        {
          "formats": ["oas3", "asyncapi"],
          "given": "$.components.schemas.*"
        }
      ]
    },
  }
}
```

Description: Markdown
Formats: Array of formats

P0lip:
Closes #1837 

Needs https://github.com/stoplightio/spectral/pull/1838 https://github.com/stoplightio/spectral/pull/1839

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


packages/core/src/ruleset/rule/rule.ts
```diff
@@ -157,16 +167,48 @@ export class Rule implements IRule {
...
 
     return resolvedExpr;
   }
 
+  #resolveAliasForFormats(
+    name: string,
+    { targets }: RulesetScopedAliasDefinition,
+    formats: Set<Format> | null,
+  ): string {
+    const errorMessage = `Alias "${name}" is applicable to certain formats, but the format of the linted document is not matched`;
```
marbemac: How does this manifest itself to end user, and what behavior does it cause? If there's a test around this behavior lmk and that might tell me, just couldn't find it :).

If there is a rule that uses an alias that isn't applicable to the document being linted, will this cause the lint to not go through and/or error?

marbemac (commented):
Is there a separate issue to update the docs?

Not strictly related to this PR, but do we have sense of perf impact of using aliases versus not using aliases?

P0lip:
> Is there a separate issue to update the docs?

Yeah, there's some in platform-internal https://github.com/stoplightio/platform-internal/issues/8141

> Not strictly related to this PR, but do we have sense of perf impact of using aliases versus not using aliases?

That's actually a difficult question. I'd say it depends.
From the technical standpoint, there isn't any performance hit, because we simply unfold these aliases to regular JSONPath expressions, so the only cost we take is that "resolving" logic that I added in this PR and it's a totally negligible cost.
On the other hand, aliases may make it more tempting to write more complex JSONPath expressions, which, at times, may in fact result in a degredaded performance. 
It all boils to down the paths you supply Nimma with - while it covers pretty much all common cases, there are certain scenarios it handles worse.
These cases are shown [here](https://github.com/P0lip/nimma/blob/master/src/__tests__/codegen.test.mjs#L646).
Overall, such usage is very rare, and I doubt this will change with aliases, but it may potentially happen.

That's why I'm thinking about exposing some internal logic I use in that library so we can potentially use it in Spectral/Studio, and warn about inefficient/bad JSONPath expressions.
These poor JSONPath expressions are rather not intended anyway, because they are very likely to match properties you don't want to target.

packages/core/src/ruleset/rule/rule.ts
P0lip: image
given the following document
```yaml
swagger: 2.0
definitions:
  User:
    type: woops
```
and the following ruleset
```yaml
formats:
  - oas2
  - oas3
aliases:
  SchemaObject:
    targets:
      - formats:
          - oas3
        given: $.components.schemas[*]
rules:
  valid-type-schema:
    given: "#SchemaObject"
    then:
      field: type
      function: enumeration
      functionOptions:
        values:
          - string
          - boolean
          - number
          - null
          - object
          - array
```

marbemac: I can't quite tell from the screenshot - did that result in the lint failing, or is it just a log to stdout?

P0lip: @marbemac it fails, I updated the screenshot.

marbemac: Hmm - do we want that? makes it impossible to write some rules that target oas2 and some that target oas3 etc

P0lip: so what would be the expected behavior then if an alias cannot be resolved?
If you want to run a rule for both specs, you need to add a proper target.
For instance, the following would be considered valid:
```yaml
formats:
  - oas2
  - oas3
aliases:
  SchemaObject:
    targets:
      - formats:
          - oas2
        given: $.definitions[*]
      - formats:
          - oas3
        given: $.components.schemas[*]
rules:
  valid-type-schema:
    given: "#SchemaObject"
    then:
      field: type
      function: enumeration
      functionOptions:
        values:
          - string
          - boolean
          - number
          - null
          - object
          - array
```

marbemac: I was mulling the idea of just ignoring the rule (not set on this, just trying to think through what the best most balanced default behavior is). For example what would happen if the original example was run again, but with this adjustment - would it pass?

```yaml
formats:
  - oas2
  - oas3
rules:
  valid-type-schema:
    given: "$.components.schemas[*]"
    then:
      field: type
      function: enumeration
      functionOptions:
        values:
          - string
          - boolean
          - number
          - null
          - object
          - array
```

P0lip: > I was mulling the idea of just ignoring the rule (not set on this, just trying to think through what the best most balanced default behavior is).

Ah, ok, I see what you mean.

> For example what would happen if the original example was run again, but with this adjustment - would it pass?

it would pass, just wouldn't match anything.

That said, the case with aliases is a bit different because we do need to inform the user about potentially broken alias.
Right now we do it by throwing an error and refusing to lint the file, but I'm open to ideas.

P0lip: @mnaumanali94 whatcha think ^?

marbemac: > we do need to inform the user about potentially broken alias.

But we don't inform the user about a potentially broken json path? When you really boil it down isn't an alias just a pointer to a json path for the most part - feels weird that by simply changing a rule to use an alias with the same json path it used to have results in a different outcome when running lint 🤔.

P0lip: > But we don't inform the user about a potentially broken json path

We don't, but it's more apparent than it is with aliases.

I agree with the "pointer to a json path" bit.

I mean, I'm certainly not against making it more lenient - just need to evaluate how this would feel.
Let's make Nauman decide.
Getting rid of that throw would be very easy, so let's not take the implementation factor into the account when making the choice.
Both options are easy to implement.

P0lip: After some thinking, I actually believe that the case we deal with here might closer to to an empty given:

```
given: []
```

mnaumanali94: I think letting the file lint is the right thing to do. We should have some sort of warning though that a particular rule wasn't run as folks could assume that the file was linted against that rule. Maybe that as a message/info is the right way to go about it?

mnaumanali94: That could be a separate thing though. Doesn't need to happen in this PR

P0lip: Fair enough.
I'd still argue we should throw, but let it be. It doesn't change much from the perspective of code.

marbemac (approved):

PR #1845, opened 2021-09-22
feat(functions): revamp Ajv integration

### LINKED ISSUE #1885: memory leak in schema function
**Describe the bug**

`schema` function has a memory leak.
It seems like the validator instances are never garbage collected.

**To Reproduce**

Lint document a few times.

**Expected behavior**
There's no memory leak.

**Screenshots**
image

**Additional context**
This is not particularly painful in the context of CLI, but it's the other way around when linting is done in the same isolate for a number of times.

 - Library version: 6.x.x

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1885

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

A list of changes:
- we try to reuse the validator fn if we generated one for the same schema - currently we always create a new one unless a scheme has $id. `documentInventory` is stable enough to serve us. It's lifespan should be equal to a single `run` execution, so basically a single lint process.
- meta draft has been removed in favor of https://www.npmjs.com/package/ajv-draft-04
- bumped some related deps


domagojk (approved):
LGTM

PR #1850, opened 2021-09-27
docs: Replace inconsistent JSON examples with YAML in Custom Rulesets

shrink:
The rulesets documentation examples are YAML, except for the overrides and aliases. I was initially confused by this and thought there was some difference in the choice of JSON vs. YAML for these examples, but after testing, I couldn't find any reason to use JSON over YAML -- I think it was just the personal preference of the original author.

- [x] Replaced JSON examples with YAML 
- [x] Fixed a mistake describing "JSON Pointer" as "JSON Path" (related to #1702)
- [x] Moved **Aliases** and **Overrides** sections _above_ the **Alternative JS Ruleset Format** section

- [ ] Yes
- [x] No


philsturgeon (dismissed):

P0lip (approved):
Thanks for the PR!

PR #1864, opened 2021-10-04
fix(rulesets): support matrix parameter style for openapi3

### LINKED ISSUE #1863: matrix parameter type not supported with openapi3
**Describe the bug**
openapi 3 (or swagger2) supports matrix parameter style, which looks like this:
`/foo/bar;key=value/baz`
in this case, the query path should look like this: `/foo/bar{;key}/baz`.
however its currently not supported, and gives the following error message: 
`Parameter "key" must be used in path "/foo/bar{;key}/baz".`
It most probably comes from https://github.com/stoplightio/spectral/blob/develop/packages/rulesets/src/oas/functions/oasPathParam.ts#L5 , where the regex does not matches `;`. Note to mention, that it may also contains asterix (`*`) and question mark (`?`) by the specification.

Also it seems it doesnt work even without the `;` , when there are more than one parameters are given.

**To Reproduce**

1. Given this OpenAPI/AsyncAPI document '...'
```json
{
  "openapi": "3.0.0",
  "info": { "title": "matrix failure", "version": "1.0"},
  "servers": [{"url": "http://localhost:3000"}],
  "paths": {
    "/foo{;key}": {
      "parameters": [
        {
          "name": "key",
          "in": "path",
          "schema": {
            "type": "string"
          },
          "required": true,
          "style": "matrix"
        }
      ],
      "get": { "responses": { "200": { "description": "OK"} } } },
    "/foo{key}{another}": {
      "parameters": [
        {
          "name": "key",
          "in": "path",
          "schema": {
            "type": "string"
          },
          "required": true,
          "style": "matrix"
        },
        {
          "name": "another",
          "in": "path",
          "schema": {
            "type": "string"
          },
          "required": true,
          "style": "matrix"
        }
      ],
      "get": { "responses": { "200": { "description": "OK"} } } }
	}
}

```
2. Run this CLI command:
```bash
$ cat .spectral.json 
{\n\t"extends": ["spectral:oas", "spectral:asyncapi"]\n}
$ spectral lint example.json
[...]
 8:28    error  path-params                 Parameter "key" must be used in path "/foo{;key}".             paths./foo{;key}.parameters[0]
  30:9    error  path-params            Parameter "another" must be used in path "/foo{key}{another}".                            paths./foo{key}{another}.parameters[1]
 40:13    error  path-params            Operation must define parameter "{keyanother}" as expected by path "/foo{key}{another}".  paths./foo{key}{another}.get

```

**Expected behavior**
Expected to support matrix parameters :)

 - Library version: `6.0.0`
 - OS: osx

**Additional context**

related specs: https://swagger.io/docs/specification/serialization/#query

lintaba:
Fixes #1863 .

support for `/api{;foo}{;?bar*}` styled paths (matrix parameters)
related specs: https://swagger.io/docs/specification/serialization/#query

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [X] Yes
- [ ] No

openapi's path parameters like /api/{foo}{bar} now parsed as two separated parameter, instead of foobar. not sure if it breaks anyone's workflow, but not supposed to do so, since this pattern was already not working as intended.


philsturgeon:
Thank you @lintaba! Do you think you could add some tests that prove this new syntax works? Are you familiar with Jest or would you like a hand finding the appropriate tests to copy/tweak?

packages/rulesets/src/oas/functions/oasPathParam.ts
```diff
@@ -2,7 +2,7 @@ import type { JsonPath, Segment } from '@stoplight/types';
 import type { IFunction, IFunctionResult } from '@stoplight/spectral-core';
 import { isObject } from './utils/isObject';
 
-const pathRegex = /(\{[a-zA-Z0-9_-]+\})+/g;
+const pathRegex = /(\{;?\??[a-zA-Z0-9_-]+\*?\})/g;
```
P0lip, line 5: I believe we should be a bit more specific, and apply this regular expression only when `style=matrix`.
What do you think?

We can keep 2 regular expressions - that's ok.

lintaba: regarding to the docs (https://swagger.io/docs/specification/serialization/#path), it can always end with `*`  when `explode=true`, and can begin with ``, `.`, `;` depending on style=`simple`, `label`, `matrix`. also in my understanding it can begin with `?` or `&` for style= `form` / `pipeDelimited` / `spaceDelimited`. and moreover, can begin with `+` for `allowReserved=true`. 

So following the whole specs would need a lot of changes and edge case checks. Question here if its more okay-ish to get errors on good path's, or to not get errors on bad path's.

P0lip: hmm, true, this would indeed make the fn a bit more complicated than it is now.
@philsturgeon what do you think? I'm indifferent, it's more about how lenient we want to be, and what the accepted number of false positives is.

P0lip (commented):
Besides that one concern I shared, this looks good. Thanks for tackling this!

We could perhaps one more test case that actually verifies the "unused" scenario, but otherwise great job.

lintaba:
> Besides that one concern I shared, this looks good. Thanks for tackling this!
> 
> We could perhaps one more test case that actually verifies the "unused" scenario, but otherwise great job.

could you please give some context/example about what do you mean by "unused" scenario? 
If you meant the optional part, that is for the actual parameter during request, which is out of this validator's scope. When the parameter is listed in the url, it should be presented in the parameters. Which is already validated. If its optional, that can only mean that there is a flag on the parameters. so ofc it could be extended with a `ensureAllDefinedPathParamsAreMatchingPahts` validation. Which may be some future milestone ¯\_(ツ)_/¯

P0lip:
> could you please give some context/example about what do you mean by "unused" scenario?

basically the error scenario, right now we only have test cases for valid examples, and would be nice to include an invalid one too.

P0lip (approved):

PR #1904, opened 2021-10-18
feat(functions): export functionOptions

billiegoose:
Refactoring to make it easier to import the functionOptions schemas from `@stoplight/spectral-functions`. The function options are currently exposed on each individual function, but since the functions aren't very portable (relying on certain shims and specific bundler features that are not widely supported) that has caused issues for Stoplight developers working on the Spectral UX. By building and exporting a simple JSON object with all the functionOptions we can stay in sync with Spectral rather than maintain separate hard-coded values.

- [ ] Yes
- [x] No

**Screenshots**

Here's a (silly) example demonstrating the functionOptions being used in Storybook.

https://user-images.githubusercontent.com/587740/137547894-9f66a4af-f9f7-410a-98b3-c65cd2a11997.mov

### DIFF
--- a/packages/core/src/ruleset/rulesetFunction.ts
+++ b/packages/core/src/ruleset/rulesetFunction.ts
@@ -73,6 +73,8 @@ type SchemaDefinition = Schema | boolean;
 
 const DEFAULT_OPTIONS_VALIDATOR = (o: unknown): boolean => o === null;
 
+export type CustomFunctionOptionsSchema = Schema | null;
+
 export function createRulesetFunction<I extends unknown, O extends unknown>(
   {
     input,
--- a/packages/core/src/ruleset/rulesetFunction.ts
+++ b/packages/core/src/ruleset/rulesetFunction.ts
@@ -73,8 +73,6 @@ type SchemaDefinition = Schema | boolean;
 
 const DEFAULT_OPTIONS_VALIDATOR = (o: unknown): boolean => o === null;
 
-export type CustomFunctionOptionsSchema = Schema | null;
-
 export function createRulesetFunction<I extends unknown, O extends unknown>(
   {
     input,
--- a/packages/functions/src/optionSchemas.ts
+++ b/packages/functions/src/optionSchemas.ts
@@ -1,5 +1,5 @@
 import type { CustomFunctionOptionsSchema } from '@stoplight/spectral-core';
-import { CasingType } from '../types';
+import { CasingType } from './types';
 
 export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
   alphabetical: {
--- a/packages/functions/src/optionSchemas.ts
+++ b/packages/functions/src/optionSchemas.ts
@@ -137,8 +137,28 @@ export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
     },
   },
   truthy: null,
-  undefined: {
-    type: 'null',
+  undefined: null,
+  schema: {
+    additionalProperties: false,
+    properties: {
+      schema: {
+        type: 'object',
+      },
+      dialect: {
+        enum: ['auto', 'draft4', 'draft6', 'draft7', 'draft2019-09', 'draft2020-12'],
+        default: 'auto',
+      },
+      allErrors: {
+        type: 'boolean',
+        default: false,
+      },
+      prepareResults: true,
+    },
+    required: ['schema'],
+    type: 'object',
+    errorMessage: {
+      type: '"schema" function has invalid options specified. Example valid options: { "schema": { /* any JSON Schema can be defined here */ } , { "schema": { "type": "object" }, "dialect": "auto" }',
+    },
   },
   unreferencedReusableObject: {
     type: 'object',
--- a/packages/functions/src/optionSchemas.ts
+++ b/packages/functions/src/optionSchemas.ts
@@ -1,5 +1,7 @@
-import type { CustomFunctionOptionsSchema } from '@stoplight/spectral-core';
+import type { createRulesetFunction } from '@stoplight/spectral-core';
 import { CasingType } from './types';
+
+type CustomFunctionOptionsSchema = Parameters<typeof createRulesetFunction>[0]['input'];
 
 export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
   alphabetical: {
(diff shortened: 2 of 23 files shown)


P0lip:
@wmhilton make sure to commit all the files, it seems like you forgot to commit a few and tests fail.

> specific bundler features that are not widely supported

ha? package exports is [supported](https://webpack.js.org/guides/package-exports/#support) by all major bundlers

billiegoose:
> make sure to commit all the files, it seems like you forgot to commit a few and tests fail.

oh thank god. I was like "why the HELL are these linting errors saying it's of type `any`" I'm still on my first ☕ 



> ha? package exports is [supported](https://webpack.js.org/guides/package-exports/#support) by all major bundlers

We aren't all on the bleeding edge. Some of us still use Node 12 and Webpack 4!

billiegoose:
OK it should be good now :D

packages/functions/src/schema/optionSchemas.ts
```diff
@@ -0,0 +1,178 @@
+import type { CustomFunctionOptionsSchema } from '@stoplight/spectral-core';
```
P0lip: Let's move this file up or somethig.
`schema` = `schema` function, the core function we expose, so this feels like an odd spot for it.

packages/core/src/ruleset/rulesetFunction.ts
```diff
@@ -73,6 +73,8 @@ type SchemaDefinition = Schema | boolean;
 
 const DEFAULT_OPTIONS_VALIDATOR = (o: unknown): boolean => o === null;
 
+export type CustomFunctionOptionsSchema = Schema | null;
```
P0lip: do we really need this type? Feels somewhat redundant 🤷‍♂️

packages/functions/src/types.ts
```diff
@@ -0,0 +1,18 @@
+export enum CasingType {
```
P0lip, line 1: can we move this back to `casing` function?

P0lip (commented):
Overall OK.
I'd remove `CustomFunctionOptionsSchema` as it doesn't seem to bring much of a value.
Do keep in mind that this may break anytime, as Spectral will inevitably become fully ESM, so nested exports may break as they are never considered as a part of the public API.

packages/core/src/ruleset/rulesetFunction.ts
billiegoose: I needed `Schema` to be exported.

P0lip:
> We aren't all on the bleeding edge. Some of us still use Node 12 and Webpack 4!

FWIW Node.js 12 does support it just fine.
Well, Webpack v4 is becoming barely usable in the year of ESM, so we just gotta try to migrate, as it causes a true pain in the case of some of our packages (not Spectral tho). 😬

packages/functions/src/types.ts
billiegoose: problem is it's as a value later, and `casing.ts` pulls in things from `@stoplight/spectral-core`, so I can't easily import it in `optionSchemas.ts` without it blowing up from a small JS file into a monster with dependencies.

packages/core/src/ruleset/rulesetFunction.ts
billiegoose: not sure how you expect users to write custom functions when the type isn't exported...

packages/functions/src/types.ts
billiegoose: The general pattern we've adopted at Stoplight is to put as many pure types into a separate `types.ts` file so that the types can be re-used without dragging around implementations. I'm not sure why Spectral doesn't do this.

billiegoose:
> Do keep in mind that this may break anytime, as Spectral will inevitably become fully ESM, so nested exports may break as they are never considered as a part of the public API.

That's fine. It's still better than maintaining it separately in platform-internal IMO.

billiegoose:
> I'd remove `CustomFunctionOptionsSchema` as it doesn't seem to bring much of a value.

Currently `Schema | null` isn't exported. So it only works by type inference of the function arguments.

Would you like me to change
```ts
export const optionSchemas: Record<string, CustomFunctionOptionsSchema>
```
to

```ts
export const optionSchemas: Record<string, Parameters<typeof createRulesetFunction>[0]>
```

:trollface:

P0lip:
> export const optionSchemas: Record<string, Parameters<typeof createRulesetFunction>[0]>

yeah, that's OK.

billiegoose:
lol it ended up being 
```tsx
export const optionSchemas: Record<string, Parameters<typeof createRulesetFunction>[0]['input']>;
```

but it works

packages/functions/src/types.ts
P0lip: > The general pattern we've adopted at Stoplight is to put as many pure types into a separate types.ts

Never heard of it, to be honest. 😆 
We do that in platform-internal, but it's quite project-dependant overall.

packages/functions/src/optionSchemas.ts
```diff
@@ -0,0 +1,180 @@
+import type { createRulesetFunction } from '@stoplight/spectral-core';
+import { CasingType } from './types';
+
+type CustomFunctionOptionsSchema = Parameters<typeof createRulesetFunction>[0]['input'];
+
+export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
```
P0lip, line 6: packages/functions/src/schema/index.ts you need this one too

packages/functions/src/types.ts
P0lip: but yeah, I see now. It's alright.
TBH, I'll probably incorporate `json-schema-to-typescript` at some point, so we don't need to maintain these `option` types anymore.

billiegoose: > `json-schema-to-typescript`

👀 ooooh

P0lip: `json-schema-to-typescript` - we actually use it already in Spectral for ruleset-migrator. It works pretty well!

P0lip (dismissed):

P0lip (approved):

packages/functions/src/optionSchemas.ts
billiegoose: oooh I didn't even notice that one ha.

pamgoodrich:
@wmhilton : Is this something that customers would be interested in or is it something that impacts Stoplight devs only?

PR #1914, opened 2021-10-24
fix(rulesets): variable not supported in server url within OAS 3.1

### LINKED ISSUE #1911: variable not supported in server url within openapi 3.1.0
**Describe the bug**
When a variable is used in a server url within an openapi 3.1.0 document as follow:
```
openapi: 3.1.0
...
servers:
- url: https://api.example.com/{version}
  description: |
    API location
  variables:
    version:
      default: v1

```
, the following error is reported by spectral :

```
spectral lint specs/openapi.yaml


/home/me/udd/public-profiles-api/specs/openapi.yaml
 103:8  error  oas3-schema  "url" property must match format "uri".  servers[0].url

✖ 1 problem (1 error, 0 warnings, 0 infos, 0 hints)
```

All is fine if the variable is removed or if the openapi version is set to 3.0.3 .

**To Reproduce**
see above

**Expected behavior**

Spectral should not report an error in this case

 - Library version: @stoplight/spectral-cli 6.1.0
 - OS: Linux

P0lip:
Fixes #1911 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

I treated this issue as an opportunity to pull the latest JSON Schema definition for OAS 3.1.
That being said, a few manual changes have been applied:
- `$dynamicRef` -> `$ref` cause Ajv doesn't like it. The former definition has the same transformation applied.
- fixed a typo in `descriptions` - should be `description`
- the `url` property in schema had still an invalid type - they changed `uri` to `uri-reference`, but it's `uri-template`.

### DIFF
--- a/packages/rulesets/src/oas/schemas/3.1.json
+++ b/packages/rulesets/src/oas/schemas/3.1.json
@@ -1,5 +1,5 @@
 {
-  "$id": "https://spec.openapis.org/oas/3.1/schema/2021-04-15",
+  "$id": "https://spec.openapis.org/oas/3.1/schema/2021-09-28",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "type": "object",
   "properties": {
@@ -11,7 +11,8 @@
       "$ref": "#/$defs/info"
     },
     "jsonSchemaDialect": {
-      "$ref": "#/$defs/uri",
+      "type": "string",
+      "format": "uri",
       "default": "https://spec.openapis.org/oas/3.1/dialect/base"
     },
     "servers": {
@@ -119,7 +120,8 @@
           "type": "string"
         },
         "url": {
-          "$ref": "#/$defs/uri"
+          "type": "string",
+          "format": "uri"
         }
       },
       "required": ["name"],
@@ -138,7 +140,8 @@
       "type": "object",
       "properties": {
         "url": {
-          "$ref": "#/$defs/uri"
+          "type": "string",
+          "format": "uri-template"
         },
         "description": {
           "type": "string"
@@ -167,7 +170,7 @@
         "default": {
           "type": "string"
         },
-        "descriptions": {
+        "description": {
           "type": "string"
         }
       },
@@ -292,6 +295,7 @@
     },
     "path-item-or-reference": {
       "if": {
+        "type": "object",
         "required": ["$ref"]
       },
       "then": {
@@ -367,7 +371,8 @@
           "type": "string"
         },
         "url": {
-          "$ref": "#/$defs/uri"
+          "type": "string",
+          "format": "uri"
         }
       },
       "required": ["url"],
@@ -460,6 +465,9 @@
               },
               "then": {
                 "properties": {
+                  "name": {
+                    "pattern": "[^/#?]+$"
+                  },
                   "style": {
                     "default": "simple",
                     "enum": ["matrix", "label", "simple"]
@@ -484,7 +492,7 @@
                 "properties": {
                   "style": {
                     "default": "simple",
-                    "enum": ["simple"]
+                    "const": "simple"
                   }
                 }
               }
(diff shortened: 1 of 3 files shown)


packages/rulesets/src/oas/schemas/3.1.json
```diff
@@ -138,7 +140,8 @@
       "type": "object",
       "properties": {
         "url": {
-          "$ref": "#/$defs/uri"
+          "type": "string",
+          "format": "uri-template"
```
P0lip, line 144: the original schema has `uri-reference` here.

packages/rulesets/src/oas/schemas/3.1.json
```diff
@@ -167,7 +170,7 @@
         "default": {
           "type": "string"
         },
-        "descriptions": {
+        "description": {
```
P0lip, line 173: typo, it's `description` -> https://spec.openapis.org/oas/v3.1.0#server-object

packages/rulesets/src/oas/schemas/3.1.json
```diff
@@ -840,45 +856,42 @@
           "default": false,
           "type": "boolean"
         },
-        "allowEmptyValue": {
```
marbemac, line 843: 🤔 where did this go? And why didn't header have schema or content before?

packages/rulesets/src/oas/schemas/3.1.json
```diff
@@ -840,45 +856,42 @@
...
             "style": {
               "default": "simple",
-              "enum": ["simple"]
+              "const": "simple"
             },
             "explode": {
               "default": false,
               "type": "boolean"
-            },
-            "allowReserved": {
```
marbemac, line 859: Ditto - seems like a substantial change?

packages/rulesets/src/oas/schemas/3.1.json
```diff
@@ -909,7 +922,8 @@
       "type": "object",
       "properties": {
         "$ref": {
-          "$ref": "#/$defs/uri"
+          "type": "string",
+          "format": "uri-reference"
```
marbemac, line 926: Interesting - why did some `"#/$defs/uri"` become `uri` (see towards top of file), and some `uri-reference`?

packages/rulesets/src/oas/schemas/3.1.json
```diff
@@ -1003,7 +1017,8 @@
                 "const": "http"
               },
               "scheme": {
-                "const": "bearer"
+                "type": "string",
+                "pattern": "^[Bb][Ee][Aa][Rr][Ee][Rr]$"
```
marbemac, line 1021: ha

marbemac (approved):
Generally looks OK - more just questions for my own edification on the changes to the schema.

packages/rulesets/src/oas/schemas/3.1.json
P0lip: haha yeah, it's case insensitive 😆

packages/rulesets/src/oas/schemas/3.1.json
P0lip: `uri-reference` means it can be a relative URI, like `../foo.json`, while `uri` format enforces a well-formed absolute URI with the scheme and so on.

packages/rulesets/src/oas/schemas/3.1.json
P0lip: this is for [Header Object](https://spec.openapis.org/oas/v3.1.0#header-object), and `allowEmptyValue` is not a valid trait since it only applies to query parameters and Header Object has to follow "header"

> All traits that are affected by the location MUST be applicable to a location of header (for example, style).

https://spec.openapis.org/oas/v3.1.0#parameter-object

image

packages/rulesets/src/oas/schemas/3.1.json
P0lip: same as the other - it's a header object and `allowReserved` is applicable only to query parameters
image

Issue #1961, opened 2021-11-17
Document missing schema functions


pamgoodrich:
These schema function options are missing from the doc:
    
    
- **Location: schema function:** dialect iconXXX
    
    Recommended text: @Nauman Ali  to provide. Pam will add to Spectral doc.

This option is internal and should not be added to the doc:
    
- **Location: schema function:** prepareResults iconXXX
    
    Recommended text: @Nauman Ali to provide. Pam will add to Spectral doc.

They need to go here: 
https://meta.stoplight.io/docs/spectral/ZG9jOjExNg-core-functions#schema

Look for text here:
https://www.notion.so/stoplight/Shared-Style-Guide-Help-Text-4d29982fb0214befa7b77a4673f92d0c

P0lip:
> Location: schema function: prepareResults iconXXX

@mnaumanali94 @pamgoodrich this is for an internal, don't document it, we shouldn't expose it in the UI either

pamgoodrich:
@P0lip Thank you. It is exposed in shared style guides now. I'll add a note to the bug list that it should not be exposed. @mnaumanali94

pamgoodrich:
@P0lip : If you can add some info about the dialect option to this issue, I can give help text to the UI team and add info to the Spectral doc.

P0lip:
`draft` - the JSON Schema draft used by function. Valid values are: 'auto', 'draft4', 'draft6', 'draft7', 'draft2019-09', 'draft2020-12'.


---

On a side note - is there any reason we'd keep them in some doc?
We should keep them in the schema in the Spectral codebase itself. This way the values are not hardcoded and users can annotate their own functions.

mnaumanali94:
There are a couple of things here:

Spectral documentation:
That's manually generated so we should document them there for sure. 

Help text in Studio for new editor:
We can pick that up from JSON schema.

P0lip:
> That's manually generated so we should document them there for sure.

we could and probably should use json schemas we have for that. This way we don't need to update docs manually. 

Our existing documentation on functions doesn't have anything that couldn't be expressed in JSON Schema

mnaumanali94:
How would we do that in meta currently? 

Once we have docs for Style guides, we could potentially generate it. Currently I think manual is the only way to go.

P0lip:
> How would we do that in meta currently?

I mean, we'd have a script somewhere in this repo that would generate the markdown file based on the schemas we have.
Super simple to implement

[...]

PR #1962, opened 2021-11-17
chore(functions): annotate options

P0lip:
Related to https://github.com/stoplightio/spectral/issues/1961#issuecomment-971873647
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


packages/functions/src/optionSchemas.ts
```diff
@@ -9,6 +9,7 @@ export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
     properties: {
       keyedBy: {
         type: 'string',
+        description: 'key to sort an object by',
```
mpodlasin: Is there a specific reason why everything here starts from a small letter? 🤔 

If this is something we want to reflect in UI later, it would be better to have it in a "sentence" format, ie starting with big letter and ending with a dot?

docs/reference/functions.md
```diff
@@ -62,6 +62,15 @@ The value should be `false`, `""`, `0`, `null` or `undefined`. Basically anythin
 
 Count the length of a string an or array, the number of properties in an object, or a numeric value, and define minimum and/or maximum values.
 
+<!-- title: functionOptions -->
```
mpodlasin, line 65: Beautiful, Mister.

packages/functions/src/optionSchemas.ts
P0lip: Could be yeah. I mostly copied it from docs and didn't change anything.

mpodlasin (approved):

PR #1982, opened 2021-12-02
feat(core): allow to override severity in custom functions

albertored:
When writing specific custom functions I found it useful to be able to override the severity defined in the rule and to return different severities based on the type of the error found. This PR addresses exactly this.

I haven't modified the documentation because I'm not sure it is something you want to add but If you find it useful I will do it

**Checklist**

- [x]  Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


P0lip:
The code looks brilliant.
However, I'm not fully convinced whether we'd like to have it in Spectral.
It's certainly a great addition for more powerful users, yet it may also confusion as `severity` set in the ruleset wouldn't be respected at times. I'll let @mnaumanali94 decide here.
At the very least we should add a note somewhere in documentation warning about the side effects of that override.

albertored:
Yes, I agree it is something not so common and that can be confusing but I find it very useful when defining complex custom functions. I understand if you think it's not something that should be added to spectral, at the moment I'm using the [patch-package](https://www.npmjs.com/package/patch-package) package to manually apply the change to spectral and I can live with it.

Maybe this problem is a syntom of a more deep one. If we define a _custom function_ as a function that performs some validation on a _general_ piece of the specification I am using it in the wrong way. My custom functions are better defined as _custom rules_ because they validate a _specific_ piece of the specification and in this context it makes sense that they can return different severities while they should not accept a severity from the ruleset file. I don't know if my thought was clear enough, but nevermind, it goes beyond the scope of this PR

mnaumanali94:
@albertored Hey, thank you for using spectral! I understand the need but like your rightly noted, custom functions aren't built to be used as rules per se and more for particular manipulations that spectral doesn't allow. If you can share your ruleset, i'd love to see your usecases and how we can improve spectral itself.

PR #1988, opened 2021-12-06
feat(ruleset-bundler): add ruleset loader

### LINKED ISSUE #1956: Expose Javascript function to load custom ruleset from file
**User story.**
As a **developer consuming spectral-core from javascript**, I can **load custom ruselet from yml files**, so that I can use them programatically on a microservice.

I'm trying to use spectral-core 6 from javascript code and I'm not able to programatically load rulesets from yml files.
Loading files as yml files doesn't work as it doesn't resolve functions (like truthy, falsy...) and all I get are string properties...

Digging in spectral-core source code there is a function for reading files is included in spectral-cli but that function is not exported as a type.d.ts (or is it?).

https://github.com/stoplightio/spectral/blob/develop/packages/cli/src/services/linter/utils/getRuleset.ts#L36

Trying to import the function like this doesn't work:
```ts
import { getRuleset } from '@stoplight/spectral-cli/dist/services/linter/utils';
```
(I guess I can not import from _dist_ folder just like that)

I would like spectral-core to expose `getRuleset` function for other applications to use it programatically
or even update `Spectral.setRuleset` signature so it can receive rules as string urls

```ts
Spectral.setRuleset(ruleset: RulesetDefinition | Ruleset | string): void
```

```ts
const fileUrl = '<local or http file url>';
spectral.setRuleset(fileUrl);
spectral.run(..);
```

**Additional context**

P0lip:
Closes #1956

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


.circleci/config.yml
```diff
@@ -72,7 +72,7 @@ commands:
       - run: yarn pretest
       - run:
           name: Run node tests
-          command: yarn test.jest --coverage --maxWorkers=<< parameters.max-workers >>
```
marbemac: This removal of `--coverage` intentional? (ok if it is, just double checking)

packages/ruleset-bundler/README.md
```diff
@@ -13,3 +11,24 @@
 - treeshake - whether to enable tree shaking. False by default.
 
 **Bolded** options are required.
+
+## Loading YAML/JSON Ruleset
+
+It is assumed you have [@stoplight/spectral-ruleset-migrator](https://www.npmjs.com/package/@stoplight/spectral-ruleset-migrator) installed.
```
marbemac: Any reason to not have this as a dependency if it's always required? Env issue, dedupe issue, something else?

packages/ruleset-bundler/package.json
```diff
@@ -18,6 +18,11 @@
     ".": {
       "default": "./dist/index.js"
     },
+    "./with-loader": {
+      "default": "./dist/loader/node.js",
+      "node": "./dist/loader/node.js",
+      "browser": "./dist/loader/browser.js"
```
marbemac, line 24: 👌

packages/ruleset-bundler/src/loader/common/bundle.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ErrorWithCause } from 'pony-cause';
+
+import { stdin } from '../../plugins/stdin';
+import { isBasicRuleset, isErrorWithCode } from './utils';
+import { bundleRuleset, BundleOptions, IO } from '../../index';
+
+export async function bundle(rulesetFile: string, bundleOptions: BundleOptions, { fs }: IO): Promise<string> {
+  try {
+    if (isBasicRuleset(rulesetFile)) {
+      const migratedRuleset = await migrateRuleset(rulesetFile, {
```
marbemac, line 12: Ah I think I see - `migrateRuleset` naming confused me. More like a ruleset transformer in this case, to transform from plain object format to esm format.

marbemac (commented):
excellent work, mostly just a couple of questions

.circleci/config.yml
P0lip: Yeah, it's already gone in develop branch

packages/ruleset-bundler/README.md
P0lip: > Any reason to not have this as a dependency if it's always required? Env issue, dedupe issue, something else?

hmmm, yeah, we could have it installed, why not.

PR #1992, opened 2021-12-09
fix(functions): schema enum option

mallachari:
Required by https://github.com/stoplightio/platform-internal/issues/9295

Adds type property to dialect enum option for schema function

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/functions/src/optionSchemas.ts
+++ b/packages/functions/src/optionSchemas.ts
@@ -157,6 +157,7 @@ export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
         description: 'Any valid JSON Schema document.',
       },
       dialect: {
+        type: 'string',
         enum: ['auto', 'draft4', 'draft6', 'draft7', 'draft2019-09', 'draft2020-12'],
         default: 'auto',
         description: 'The JSON Schema draft used by function.',


P0lip (changes requested):
This should be fixed elsewhere.
Enum usually isn't used with type keyword, so fixing the piece of code here won't fix other potential usages of enum.

mallachari:
There is only one other usage of enum within that file for casing function and it has type property set to string. Couldn't all properties for function options have type assigned?

P0lip:
> There is only one other usage of enum within that file for casing function and it has type property set to string. Couldn't all properties for function options have type assigned?

but these options can be specified by users as well, they are not specific to `spectral-functions` packages.
Each custom function may specify its own too.

I'm just not sure why using `type` helps tho.

Using type with enum is kind of like you did the following in TS

```ts
type Abc = string & ('foo' | 'bar');
```
where `string` is the `type: string`, and `'foo' | 'bar'` is `enum: [foo, bar]`

We should just render a dropdown anytime there's an enum.

mallachari:
Alright, alright, I feel convinced :D

PR #1998, opened 2021-12-13
chore(core): export schema definition

mallachari:
Needed for https://github.com/stoplightio/platform-internal/issues/9297

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/ruleset/index.ts
+++ b/packages/core/src/ruleset/index.ts
@@ -1,6 +1,6 @@
 export { assertValidRuleset, RulesetValidationError } from './validation';
 export { getDiagnosticSeverity } from './utils';
-export { createRulesetFunction, SchemaDefinition } from './rulesetFunction';
+export { createRulesetFunction, SchemaDefinition as RulesetFunctionSchemaDefinition } from './rulesetFunction';
 export { Format } from './format';
 export { RulesetDefinition, RuleDefinition, ParserOptions, HumanReadableDiagnosticSeverity } from './types';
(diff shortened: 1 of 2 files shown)


packages/core/src/ruleset/index.ts
```diff
@@ -1,6 +1,6 @@
 export { assertValidRuleset, RulesetValidationError } from './validation';
 export { getDiagnosticSeverity } from './utils';
-export { createRulesetFunction } from './rulesetFunction';
+export { createRulesetFunction, SchemaDefinition } from './rulesetFunction';
```
P0lip:
```suggestion
export { createRulesetFunction, SchemaDefinition as RulesetFunctionSchemaDefinition } from './rulesetFunction';
```

P0lip (commented):
:+1:
Let's just re-export it under a different name to avoid confusion.

mallachari:
Actually the changes you made yesterday for studio `RuleFunctionOptions` solve the problem so we might not need that type for now. Unless you think that it may be better to use it instead or might be useful in future.

P0lip (approved):

P0lip:
> Actually the changes you made yesterday for studio RuleFunctionOptions solve the problem so we might not need that type for now. Unless you think that it may be better to use it instead or might be useful in future.

Yeah, was about to say we're probably good in terms of Studio, but it won't hurt to have that exported.

PR #2011, opened 2021-12-23
feat(cli): generate multiple outputs to different files in a single c…

### LINKED ISSUE #1970: Generate multiple output formats with one cli call
**User story.**
As a spectral user, I would like to be able to generate lint reports in different formats with a single cli call without the need or lining the same spec multiple times (it can be time-consuming for huge specs).

An option is to change `--output` and `--format` cli flags to accept a comma separated list of values so that a user can decide to generate more than one report and save it into different files. In particular

```bash
# as today, stylish report on stdout 
spectral lint --ruleset ruleset.yml openapi.json

# as today, json report on stdout 
spectral lint --ruleset ruleset.yml -f json openapi.json

# json report on file report.json, stylish report on file report.txt
spectral lint --ruleset ruleset.yml -f json,stylish -o report.json,report.txt openapi.json

# json report on file report.json, stylish report on stdout
spectral lint --ruleset ruleset.yml -f json,stylish -o report.json,stdout openapi.json

# TBD error or only json report on stdout
spectral lint --ruleset ruleset.yml -f json,stylish openapi.json
```

I can create a PR if you think is worth adding the feature

albertored:
Fixes #1970 .

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


P0lip (commented):
Hey!
My apologies for such a delayed review on this one.
I'm thinking we could perhaps do `output.html`, `output.junit` and so on?
Might cause less friction. What do you think?

fgreinacher:
I think we could simplify this even more by keeping the existing options with some extra syntax support:

- `--format json,junit,stylish` to generate multiple formats.
- `--output json:results.json,junit:results.xml` to specify file names
   - when no filename specified for a formatter, it'll write to stdout (like now)
   - fail if there are multiple formatters and only a single filename (e.g. `--format json,junit --output results.txt`)
   - fail if there is a filename is specified for a formatter that is not active (e.g. `--format json --output junit:results.xml`)

WDYT @P0lip @albertored?

P0lip:
@fgreinacher 
That would work too, I like it.
One could also consider the following:

```
--format json --format junit
```

However, I'm not certain about the syntax for the output if we went with the above convention.
We'd probably need something as follows

```
--output:json <file> --output:junit <file>
```

albertored:
@fgreinacher I also like it, it is similar to the one I originally suggested in the issue but more clear.
Should we wait for some other feedback or can I implement the needed changes?

fgreinacher:
> @fgreinacher I also like it, it is similar to the one I originally suggested in the issue but more clear.
> Should we wait for some other feedback or can I implement the needed changes?

I think @P0lip should have the last call on this 😀

albertored:
> I think @P0lip should have the last call on this grinning

Yes, sure, I forgot to tag him

P0lip:
@albertored go for it if you have spare time! Thanks a lot for putting all the effort into it.
I like that functionality, so as soon as code is ready, we could have it merged in and released

albertored:
@P0lip  @fgreinacher done, let me know what do you think, the logic for handling those options is quite complex.

I also have a lint error that I'm not able to fix...
```
/packages/cli/src/commands/lint.ts
   99:17  error    Unsafe assignment of an `any` value  @typescript-eslint/no-unsafe-assignment
```

packages/cli/src/commands/__tests__/lint.test.ts
```diff
@@ -152,11 +156,34 @@ describe('lint', () => {
...
+    (formatOutput as jest.Mock).mockClear();
+    (formatOutput as jest.Mock).mockReturnValue('<formatted output>');
+
+    await run(`lint --format html,json --output json:foo.json ./__fixtures__/empty-oas2-document.json`);
+    await new Promise(resolve => void process.nextTick(resolve));
+    expect(writeOutput).toBeCalledTimes(2);
+    expect(writeOutput).nthCalledWith(1, '<formatted output>', undefined);
+    expect(writeOutput).nthCalledWith(2, '<formatted output>', 'foo.json');
+  });
+
```
fgreinacher, line 180: Should we test error cases here? For example:

- `--format json,junit --output results.txt` (output needs to include format)
- `--format json --output junit:results.xml` (unselected format in output)

test-harness/scenarios/strict-options.scenario
```diff
@@ -20,8 +20,8 @@ Options:
       --version                  Show version number  [boolean]
       --help                     Show help  [boolean]
   -e, --encoding                 text encoding to use  [string] [choices: "utf8", "ascii", "utf-8", "utf16le", "ucs2", "ucs-2", "base64", "latin1"] [default: "utf8"]
-  -f, --format                   formatter to use for outputting results  [string] [choices: "json", "stylish", "junit", "html", "text", "teamcity", "pretty"] [default: "stylish"]
```
fgreinacher, line 23: I don't really like that we lost the `choices` in the help text.

Wondering whether it would be easier (implementation- and usage-wise) to allow multiple `--format` and `--output` parameters instead of the providing multiple values within one parameter, e.g. `--format stylish --format junit --output junit:junit.xml` instead of of `--format stylish,junit --output junit:junit.xml`.

packages/cli/src/commands/__tests__/lint.test.ts
albertored: Yes, I was waiting to have a feedback on the implementation first

test-harness/scenarios/strict-options.scenario
albertored: Initially, I was looking at having `--format json junit` but `yargs` [does not behave well](https://github.com/yargs/yargs/blob/main/docs/api.md#arraykey) with array arguments together with positional ones.

Repeating the `--format` flag is [completely doable](https://github.com/yargs/yargs/blob/main/docs/tricks.md#arrays) instead

fgreinacher: We could also consider https://github.com/yargs/yargs/blob/main/docs/tricks.md#objects to make the implementation even easier (e.g. `--output.junit junit.xml`)

P0lip: yup, it feels like using objects might be a good idea for outputs.

I'm trying to explore how invasive changing the type of `format` option would be.
It feels like we'd need to consider it a breaking change if we changed the type of the option, as `yargs` would need to be instrumented with `--` to stop the parsing, or, alternatively, pass `format` after the positional arguments.

P0lip: thoughts https://github.com/stoplightio/spectral/pull/2037?

P0lip:
Closing this one as we've already merged #2037.
Once again a huge thanks to all of you.

PR #2014, opened 2021-12-29
fix(ruleset-migrator): order of functions property should not matter

P0lip:
Right now, if you place `functions` somewhere below `rules`, you'll see `void 0` inserted.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

PR #2015, opened 2021-12-30
feat(ruleset-migrator): cleaner output

P0lip:
2 changes here:
- import members are sorted
- for CJS, shorthand object properties are used when possible 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No.


domagojk:
Sorting imports looks good, but shouldn't this fix the issue with `functions` defined after `rules`?

It seems it is not working now. When I change the properties order like this:

```ts
it('should not apply to custom functions', async () => {
      await fs.promises.writeFile(
        path.join(cwd, 'custom-npm-provider-custom-functions.json'),
        JSON.stringify({
          rules: {
            rule: {
              then: {
                given: '$',
                function: 'customFunction',
              },
            },
          },
          functions: ['customFunction'], // <----------  moved to the bottom
        }),
      );
```

I get this error

Screenshot 2022-01-03 at 16 33 13

P0lip:
@domagojk we have #2014 for it. This is just a complementary PR to reduce the noise in #2014

domagojk (approved):
Ah, ok, it is just an honest one then 😄 👍

PR #2016, opened 2022-01-04
feat(core): support multiple JSONPath expressions for aliases

P0lip:
Related to https://github.com/stoplightio/platform-internal/issues/8141

Needed by https://github.com/stoplightio/spectral/pull/2018
Example usage https://github.com/stoplightio/spectral/blob/a2019f14f7db93d31effd795882c30d060a196fe/packages/rulesets/src/oas/index.ts#L38

I believe that once this one is in, we can consider the aliases frozen and won't need to make any further changes.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


docs/guides/4-custom-rulesets.md
```diff
@@ -322,13 +324,69 @@ For now the JSON, YAML, and JS, are all being maintained, and there are no curre
 
 ## Aliases
```
P0lip, line 325: @mnaumanali94 what do you think?

docs/guides/4-custom-rulesets.md
```diff
@@ -322,13 +324,70 @@ For now the JSON, YAML, and JS, are all being maintained, and there are no curre
...
 
 Aliases can be defined in an array of key-value pairs at the root level of the ruleset.
-
-```
-aliases: {
-  {Name}: '{JSONPath}'
+This is the JSON Schema definition describing aliases.
+It's a superset of `given`, with the notable difference being the possibility to distinguish between different formats.
+
+```json
```
marbemac: I'm not sure this will really be all that helpful for folks reading the docs, and might just be confusing? I'm pretty well versed with json schema, and even so it's not easy to imagine what this schema is trying to describe. The examples are much more helpful 🤷‍♂️

docs/guides/4-custom-rulesets.md
```diff
@@ -343,7 +402,51 @@ aliases:
...
+```yaml
+aliases:
+  PathItemObject: $.paths[*]
+  OperationObject: "#PathItem[get,put,post,delete,options,head,patch,trace]"
+  ParameterObject:
+    description: an optional property describing the purpose of the alias
+    targets:
+      - formats:
+          - oas2
+        given:
```
marbemac, line 376: Lol, we're ending up w similar sort of issue we always get frustrated with OpenAPI for doing... different ways to describe the same thing all over the place - e.g. an `alias` can be a string, or this object w description/targets/etc - `targets.given` can be a string or an array of strings, etc, etc 😛. No action item here, just think we should do whatever we can in the future to minimize these sorts of variants since they make tooling more difficult to build.

docs/guides/4-custom-rulesets.md
```diff
@@ -343,7 +402,51 @@ aliases:
...
+          - $.parameters[*]
+      - formats:
+          - oas3
+        given:
+          - "#PathItemObject.parameters[*]"
+          - "#OperationObject.parameters[*]"
+          - $.components.parameters[*]
+```
+
+Rulesets can then reference aliases in the [given](#given) keyword, either in full: `"given": "#Paths"`, or use it as a prefix for further JSONPath syntax, like dot notation: `"given": "#InfoContact.name"`.
```
marbemac: Very neat - perhaps adjust this to use an example from above? E.g. `"given": "#PathItemObject"` and `"given": "#ParameterObject.name"` or something? just easier to connect the dots

packages/core/src/meta/ruleset.schema.json
```diff
@@ -172,46 +172,44 @@
...
-                "items": {
-                  "type": "object",
-                  "properties": {
-                    "formats": {
-                      "$ref": "shared#formats"
-                    },
-                    "given": {
-                      "$ref": "alias-value"
-                    }
+        "if": {
```
marbemac, line 175: You fancy huh? 💃

packages/core/src/ruleset/rule.ts
```diff
@@ -28,7 +30,7 @@ export interface IRule {
   recommended: boolean;
   documentationUrl: string | null;
   then: IRuleThen[];
-  given: string[];
+  given: GivenDefinition;
```
marbemac: Ah, it used to be array only? Should we keep it that way in that case? Re the `different ways to describe the same thing` comment above.

docs/guides/4-custom-rulesets.md
P0lip: True, I might just delete it. 
Do you think the examples here are sufficient or should I add a few additional ones?

docs/guides/4-custom-rulesets.md
P0lip: Yeaaaaah. I'd just do arrays tbh, but this is now compliant with `given`

packages/core/src/ruleset/rule.ts
P0lip: ah shoot, this is the internal representation (that's after all the normalization) we use in spectral, so this indeed only an array. I should revert this particular change

packages/core/src/meta/ruleset.schema.json
P0lip: this is _much_ better for validation.
`oneOf` wouldn't generate any meaningful error messages, while this one does.

it's kind of like "if array, validate it this way, otherwise apply some different logic", so it yields better results.
Otherwise one would have "must match oneOf", which is more difficult to understand.

docs/guides/4-custom-rulesets.md
marbemac: I think delete and good for now

marbemac (approved):
Generally looks good - however as discussed, definitely think that we should not expose/advertise the baked in aliases until we have time to really think them through.

DiogoDoreto:
This PR introduced a breaking change to the `aliases` property. I started seeing errors like the one below after updating:

```
Error at #/aliases/PathItem: must be array
```

I'll update my code accordingly, but I'm adding this comment here to help anyone searching for this error in the future.

P0lip:
@DiogoDoreto true. I felt a bit uncomfortable releasing this change as non-breaking, as I knew this could affect a few users using it. In fact, the initial iteration of this PR was backward compatible, but having a few internal conversations we decided to make that soft breaking change.
However, rest assured it's a one-off situation, as we do try to strictly follow semantic versioning at all times.

PR #2018, opened 2022-01-06
feat(rulesets): use aliases in ruleset

### LINKED ISSUE #1972: oas3-valid-schema-example: Issue with properties called "default"?
**Describe the bug**

oas3-valid-schema-example: I have an OpenAPI file with an example that contains a property `default`. It seems that's the reason validation somehow applies strange rules for it.

I'm getting two errors:

>  1713:28  error  oas3-valid-schema-example  schema is invalid: data/type must be equal to one of the allowed values, data/type must be array, data/type must match a schema in anyOf  paths./udf_runtimes.get.responses[200].content.application/json.example.PHP.default
> 1726:28  error  oas3-valid-schema-example  schema is invalid: data/type must be equal to one of the allowed values, data/type must be array, data/type must match a schema in anyOf  paths./udf_runtimes.get.responses[200].content.application/json.example.R.default

I don't fully understand why that is. The default schema specifies it should be "string" and the values in the examples are strings.

**To Reproduce**

1. Given this OpenAPI/AsyncAPI document: https://github.com/Open-EO/openeo-api/blob/spectral6/openapi.yaml
2. and the config file: https://github.com/Open-EO/openeo-api/blob/spectral6/.spectral.yml  (this doesn't give an error, you need to remove the "except" rules first)
2. Run this CLI command: spectral lint openapi.yaml
3. See the error given above

**Expected behavior**
No error message, but instead a successful validation.
This worked before in 5.x, but got broken once I migrated from 5 to 6.1.

 - Library version: 6.1.0
 - OS: Windows 10

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/1972

Needs https://github.com/stoplightio/spectral/pull/2016

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/rulesets/src/oas/index.ts
+++ b/packages/rulesets/src/oas/index.ts
@@ -33,7 +33,60 @@ const ruleset = {
   formats: [oas2, oas3, oas3_0, oas3_1],
   aliases: {
     PathItem: ['$.paths[*]'],
+    ResponsesObject: {
+      targets: [
+        { formats: [oas2], given: ['$.responses', '#OperationObject.responses'] },
+        { formats: [oas3], given: ['$.components.responses', '#OperationObject.responses'] },
+      ],
+    },
+    ResponseObject: '#ResponsesObject[*]', // [(@ && !@.$ref)]
+    ParametersDefinitionsObject: {
+      targets: [
+        { formats: [oas2], given: ['$.parameters'] },
+        { formats: [oas3], given: ['$.components.parameters'] },
+      ],
+    },
+    ParametersObject: {
+      targets: [
+        { formats: [oas2], given: ['#PathItem.parameters', '#OperationObject.parameters'] },
+        {
+          formats: [oas3],
+          given: ['#PathItem.parameters', '#OperationObject.parameters'],
+        },
+      ],
+    },
+    ParameterObject: {
+      targets: [
+        { formats: [oas2], given: ['#ParametersDefinitionsObject[*]', '#ParametersObject[*]'] }, // ?(@ && !@.$ref)
+        {
+          formats: [oas3],
+          given: ['#ParametersDefinitionsObject[*]', '#ParametersObject[*]'], // [?(@ && !@.$ref)]
+        },
+      ],
+    },
+    EncodingObject: '#MediaTypeObject.encoding',
+    MediaTypeObject: ['#RequestBodyObject.content', '#ParameterObject.content', '#ResponseObject.content'],
+    HeadersObject: {
+      targets: [
+        { formats: [oas2], given: ['#ResponseObject.headers'] },
+        {
+          formats: [oas3],
+          given: ['$.components.headers', '#EncodingObject.headers', '#ResponseObject.headers'],
+        },
+      ],
+    },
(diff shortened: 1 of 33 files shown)


packages/rulesets/src/oas/index.ts
```diff
@@ -33,7 +33,60 @@ const ruleset = {
...
+        { formats: [oas2], given: ['#ResponseObject.headers[*]'] },
+        {
+          formats: [oas3],
+          given: ['#HeadersObject[*]'], // ?(@ && !$.ref)
+        },
+      ],
+    },
+    RequestBodyObject: ['#OperationObject.requestBody', '$.components.requestBodies[*]'],
     OperationObject: '#PathItem[get,put,post,delete,options,head,patch,trace]',
+    SchemaObject: ['#MediaTypeObject.schema', '#ParameterObject.schema', '$.components.schemas[*]'],
```
hudlow: Isn't this a little misleading? I mean, isn't there a possibility of a lot of schema objects embedded in properties (and embedded in properties of properties, and so on) that won't be captured here?

packages/rulesets/src/oas/index.ts
```diff
@@ -671,12 +710,7 @@ const ruleset = {
       formats: [oas3],
       recommended: true,
       type: 'validation',
-      given: [
-        "$.components.schemas..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
-        "$..content..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
-        "$..headers..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
-        "$..parameters..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]",
-      ],
+      given: ['#SchemaObject', '#SchemaObject..[?(@ && @.example)]'],
```
hudlow: Won't this capture `components.schema.MySchema.properties` if there's a property called `example`? Or maybe I'm missing something?

packages/rulesets/src/oas/index.ts
P0lip: Yeah, it's not final yet. `SchemaObject` will have to catch all schemas, I've got it somewhere but not committed yet

packages/rulesets/src/oas/index.ts
P0lip: It will, although the current impl does it as well.
In any case, before I mark that PR as ready to review, I'll need to figure out a way to do it. It's not that trivial using  JSONPath expressions.,so I'll probably need to think of something else

hudlow: Apologies, I totally missed that this was a draft!

PR #2022, opened 2022-01-17
fix(ruleset-migrator): resolve npm packages correctly

### LINKED ISSUE #2019: `spectral lint` when using NPM-provided rulesets does not resolve rulesets
**Describe the bug**

When running `spectral lint`, via `npx`, I cannot provide an NPM-based dependency for 

**To Reproduce**

Given the following `.spectral.yaml`:

```yaml
extends: 
  - spectral:oas
  - spectral-aws-apigateway-ruleset
```

And the following `openapi.yml`:

```yaml
openapi: 3.1.0
info:
  title: ''
  version: ''
  description: ''
paths: {}
```

Then running:

```
./node_modules/.bin/spectral lint openapi.yml
# or
npx @stoplight/spectral-cli lint openapi.yml
```

**Expected behavior**

The rules from https://www.npmjs.com/package/spectral-aws-apigateway-ruleset are downloaded and executed

**Actual Behaviour**

```
ENOENT: no such file or directory, open '/home/jamie/workspaces/cddo/federated-api-model/schemas/spectral-aws-apigateway-ruleset'
```

**Screenshots**

N/A

 - Spectral CLI: 6.1.0
 - OS: Linux

**Additional context**

N/A

P0lip:
Fixes #2019

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
+++ b/packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
@@ -1,0 +1,23 @@
+import { isPackageImport } from '../isPackageImport';
+
+describe('isPackageImport util', () => {
+  it.each([
+    'nimma/legacy',
+    'nimma',
+    'lodash',
+    'lodash/get',
+    'lodash/get.js',
+    '@stoplight/path',
+    '@stoplight/spectral-core',
+    '@stoplight/spectral-core/dist/file.js',
+  ])('given valid %s package import, should return true', input => {
+    expect(isPackageImport(input)).toBe(true);
+  });
+
+  it.each(['', '/nimma/legacy', 'path', 'https://cdn.skypack.dev/@stoplight/spectral-core'])(
+    'given invalid %s import, should return false',
+    input => {
+      expect(isPackageImport(input)).toBe(false);
+    },
+  );
+});
--- a/packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
+++ b/packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
@@ -14,7 +14,7 @@ describe('isPackageImport util', () => {
     expect(isPackageImport(input)).toBe(true);
   });
 
-  it.each(['', '/nimma/legacy', 'path', 'https://cdn.skypack.dev/@stoplight/spectral-core'])(
+  it.each(['', 'spectral:oas', '/nimma/legacy', 'path', 'https://cdn.skypack.dev/@stoplight/spectral-core'])(
     'given invalid %s import, should return false',
     input => {
       expect(isPackageImport(input)).toBe(false);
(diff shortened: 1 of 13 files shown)


packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    'lodash/get',
+    'lodash/get.js',
+    '@stoplight/path',
+    '@stoplight/spectral-core',
+    '@stoplight/spectral-core/dist/file.js',
+  ])('given valid %s package import, should return true', input => {
+    expect(isPackageImport(input)).toBe(true);
+  });
+
+  it.each(['', '/nimma/legacy', 'path', 'https://cdn.skypack.dev/@stoplight/spectral-core'])(
```
jamietanna: Is it worth also adding `spectral:oas` as an example here?

packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { isPackageImport } from '../isPackageImport';
+
+describe('isPackageImport util', () => {
+  it.each([
```
jamietanna, line 4: Can we also have a test case here for a versioned package? the docs specify that we should be able to pin versions

packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
P0lip: could add it, though in reality `isPackageImport` is unlikely to ever receive `stoplight:oas` as it's being replaced with `@stoplight/spectral-rulesets` here https://github.com/stoplightio/spectral/blob/develop/packages/ruleset-migrator/src/transformers/extends.ts#L21

packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
P0lip: This is not how pinning works though.
Overall there are 2 ways to pin a version:
- if you are a CLI, the version of the package you have installed will be preferred. This means the version itself is specified in `package.json` and no in the ruleset.
- regardless of the type of a user, one can always try to use services like https://unpkg.com/, but their have their own way of doing that and this wouldn't be considered as a package import, since it's just an external ruleset we fetch

jamietanna: If it's not possible [as per the current docs](https://github.com/stoplightio/spectral/blob/develop/docs/guides/7-sharing-rulesets.md?plain=1#L119-L124) can we amend them please?

Reading them, I'd expect to be able to i.e. just run the below without a `package.json` to specify the version of the pinned dependencies:

```sh
npx @stoplight/spectral-cli lint v1alpha/openapi.yml
```

domagojk (approved):
LGTM

packages/ruleset-migrator/src/utils/__tests__/isPackageImport.spec.ts
P0lip: > If it's not possible as per the current docs can we amend them please?

that bit is indeed outdated.
Going to update it.

PR #2028, opened 2022-01-19
fix(cli): correctly handle special characters in JUnit failure details

### LINKED ISSUE #2027: JUnit formatter incorrectly escapes failure details
**Describe the bug**

The JUnit formatter runs failure details through `xmlEscape` although these are contained within a `CDATA` section:

https://github.com/stoplightio/spectral/blob/develop/packages/cli/src/formatters/junit.ts#L51-L55

This [is unnecessary](https://stackoverflow.com/a/223782) and can cause visualization problems when rendering the results in a post-processing tool.

**To Reproduce**

Run spectral with JUnit formatter against a spec with errors:

```xml
<testcase time="0" name="org.spectral.json:api-ContentType-Error" classname="path/to/spec">
    <failure message="SHOULD have a content type 'application/vnd.api+json'.">
        <![CDATA[
line 126, col 19, SHOULD have a content type &apos;application/vnd.api+json&apos;. (json:api-ContentType-Error) at path             #/paths/~1info/get/responses/200/content
        ]]>
    </failure>
</testcase>
```

**Expected behavior**

```xml
<testcase time="0" name="org.spectral.json:api-ContentType-Error" classname="path/to/spec">
    <failure message="SHOULD have a content type 'application/vnd.api+json'.">
        <![CDATA[
line 126, col 19, SHOULD have a content type 'application/vnd.api+json'. (json:api-ContentType-Error) at path             #/paths/~1info/get/responses/200/content
        ]]>
    </failure>
</testcase>
```

**Prososal**

Do what's suggested in the [SO thread](https://stackoverflow.com/a/36331725) and drop `xmlEscape` in favor of something like `replace("]]>", "]]]]><![CDATA[>")`.

fgreinacher:
Those details are contained in XML CDATA section and should therefore not be XML-escaped.

Fixes #2027

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

🛠 with ❤ at [Siemens](https://opensource.siemens.io)

### DIFF
--- a/packages/cli/src/formatters/junit.ts
+++ b/packages/cli/src/formatters/junit.ts
@@ -29,6 +29,13 @@ import { printPath, PrintStyle } from '@stoplight/spectral-runtime';
 import { Formatter } from './types';
 import { groupBySource, xmlEscape } from './utils';
 
+/**
+ * Prepares the given text for inclusion in an XML CDATA section.
+ */
+function prepareForCdata(text: string): string {
+  return text.replace(']]>', ']]]]><![CDATA[>');
+}
+
 export const junit: Formatter = (results, { failSeverity }) => {
   let output = '';
 
@@ -50,8 +57,8 @@ export const junit: Formatter = (results, { failSeverity }) => {
         output += `<failure message="${xmlEscape(result.message)}">`;
         output += '<![CDATA[';
         output += `line ${result.range.start.line + 1}, col ${result.range.start.character + 1}, `;
-        output += `${xmlEscape(result.message)} (${result.code}) `;
-        output += `at path ${xmlEscape(printPath(result.path, PrintStyle.EscapedPointer))}`;
+        output += `${prepareForCdata(result.message)} (${result.code}) `;
+        output += `at path ${prepareForCdata(printPath(result.path, PrintStyle.EscapedPointer))}`;
         output += ']]>';
         output += `</failure>`;
         output += '</testcase>\n';
--- a/packages/cli/src/formatters/__tests__/__fixtures__/errors-with-special-xml-strings.json
+++ b/packages/cli/src/formatters/__tests__/__fixtures__/errors-with-special-xml-strings.json
@@ -1,0 +1,46 @@
+[
+  {
+    "code": "special-xml-strings",
+    "message": "start ' \" < > end",
+    "path": [
+      "root",
+      "'",
+      "\"",
+      "leaf"
+    ],
+    "severity": 0,
+    "source": "",
+    "range": {
+      "start": {
+        "line": 0,
+        "character": 0
+      },
+      "end": {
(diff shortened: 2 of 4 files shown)


packages/cli/src/formatters/__tests__/__fixtures__/oas3-schema-errors.json
```diff
@@ -51,7 +51,7 @@
   },
   {
     "code": "oas3-schema",
-    "message": "should have required property '$ref'",
+    "message": "should have required property '$ref' and handle special strings like \", <, > or ]]> well",
```
P0lip: this is a serialized output of `spectral.run`, thus it might be better to leave it as is and simply add another test case to `packages/cli/src/formatters/__tests__/junit.test.ts`

packages/cli/src/formatters/junit.ts
```diff
@@ -50,8 +57,8 @@ export const junit: Formatter = (results, { failSeverity }) => {
         output += `<failure message="${xmlEscape(result.message)}">`;
```
P0lip, line 60: do we need `xmlEscape` here as well then?
if not, could just remove the call together with the function

fgreinacher: Here we need it, because it will be included in the XML attribute (not in CDATA).

packages/cli/src/formatters/__tests__/__fixtures__/oas3-schema-errors.json
fgreinacher: Makes sense, done!

fgreinacher:
Harness tests adapted in https://github.com/stoplightio/spectral/pull/2028/commits/6807c2d038d18a660f658e0f9af109c87060920d 😀

P0lip (dismissed):
Thank you!

P0lip (approved):

fgreinacher:
@P0lip Do you already have a plan for a new release? We are keen to try this out - in the best case together with #2011 and #1970 - without ugly workarounds 😀

P0lip:
@fgreinacher we have a daily release cronjob that runs at around 2pm CET, so your changes should be released within a few hours.

fgreinacher:
Awesome, thanks @P0lip. I was a bit confused because there has not been a [GitHub release](https://github.com/stoplightio/spectral/releases) for a long time. But I guess you just don't create releases anymore after switching to a monorepo setup, so all good!

P0lip:
There'll be comments posted on your issues once the code is out.
We create Github Releases only for the CLI package, and indeed there wasn't any release since October or so.
`npm` picks up the latest version of each Spectral package anyway, so it's mostly Docker images & Binaries that are lagging behind.

PR #2029, opened 2022-01-19
fix(cli): add path to JUnit testcase name

fgreinacher:
This helps downstream tools disambiguate individual test failures better.

There is no formal spec for the JUnit XML format but JUnit itself does it in a very similar way for dynamic/parametrized tests (see https://github.com/junit-team/junit5/pull/1236).

Fixes #2025.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

🛠 with ❤ at [Siemens](https://opensource.siemens.io)

PR #2032, opened 2022-01-20
fix: adapt harness tests

fgreinacher:
Adapts the harness tests broken in #2029.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


fgreinacher:
@P0lip Is it possible that the CircleCI pipeline does not/cannot run on forks?

P0lip:
It should be run. That's how it always used to work, but for some reason it appears to be broken now.
I amended your previous commit though, so we should be good.

P0lip:
it's executed on my commits, so hm the author of the PR doesn't seem to matter.

PR #2033, opened 2022-01-21
feat(core): aliases take only array-ish given

P0lip:
As requested by @mnaumanali94, aliases are arrays-only now.
This is in line with that @marbemac suggested back in the previous PR.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

### DIFF
--- a/docs/guides/4-custom-rulesets.md
+++ b/docs/guides/4-custom-rulesets.md
@@ -31,7 +31,7 @@ It has a specific syntax known as [JSONPath](https://goessner.net/articles/JsonP
 Both of them support all the main JSONPath functionality and a little bit more, but this syntax may differ slightly from other JSONPath implementations.
 
 Your `given` value can be a string containing any valid JSONPath expression, or an array of expressions to apply a rule to multiple parts of a document.
-You can also consume your [aliases][#aliases] here if you have some defined.
+You can also consume your (aliases)[#aliases] here if you have some defined.
 
 Use the [JSONPath Online Evaluator](http://jsonpath.com/) to determine what `given` path you want.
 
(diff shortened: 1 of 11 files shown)


docs/guides/4-custom-rulesets.md
```diff
@@ -351,10 +356,12 @@ aliases:
     targets:
```
marbemac, line 356: I can't quite tell in the diff here, but is there an extra space (targets is indented more than description)?

marbemac (approved):
👍 nice - and always have the option to introduce non-array syntax in the future if there's high demand / use case for it.

Just one nit re the docs, otherwise looks good

mnaumanali94 (approved):
Beautiful. Hopefully we can get given here too eventually.

docs/guides/4-custom-rulesets.md
P0lip: it looks good, verified it manually now haha
We should probably use JSON as default and have some automated tool that generates YAML examples based on it.
This way we would have 2 examples at all times, which would be actually useful.

P0lip:
Despite it being a de jure breaking change, de facto it is not such since the actual usage is almost non-existent.
However, it's a one-off exception. I'd really love to follow the semantic versioning very strictly as always used to do.

billiegoose:
> Despite it being a de jure breaking change, de facto it is not such since the actual usage is almost non-existent.
However, it's a one-off exception. I'd really love to follow the semantic versioning very strictly as always used to do.

a month later this change made it to Studio and SHOCKINGLY it broke styleguides! (of course it did 🙃) @mnaumanali94 making breaking changes and violating semver is a terrible idea. @P0lip trust your gut next time.

PR #2035, opened 2022-01-23
fix(rulesets): add scopes to required in oauth2 security schemes

mkistler:
This PR fixes an error in the oas2 schema. The "scopes" property is required for all "oauth2" security schemes [[ref](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#:~:text=oauth2-,Required.,-The%20available%20scopes)], but was not specified in the "required" attribute of the "oauth2" security scheme sub-schemas.  This PR corrects that.

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


pavelkornev:
I would ask to update schema files both for OAS2 and OAS3 to the latest ones:

- OAS2 - https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v2.0/schema.json
- OAS3 - https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.json

There are few other differences as well.

**Edit:** BTW, in this PR scopes are only added to OAS2, but they're also missing in the current OAS3 version too.

mkistler:
I amended my commit to add scopes to the oauth2 flows in the oas3 schema.

pavelkornev:
I looked again in the OAS2 schema file and noticed that in the canonical schema `scopes` is not a required field and this change will make Spectral OAS2 schema incompatible with the [official OAS2 schema](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v2.0/schema.json). This requirement for `scopes` is a breaking change and will break validation for many.

IMO, for OAS2 the requirement should be created in https://github.com/OAI/OpenAPI-Specification project first.

@P0lip
I have also noticed that [Spectral OAS3](https://github.com/stoplightio/spectral/blob/develop/packages/rulesets/src/oas/schemas/3.0.json#L3) is refencing to JSON Schema Draft-07 where [the official schema](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.json#L3) is still pointing to Draft-04. Is there a reason for that?

In general, I think Spectral should avoid their own OAS schemas implementation. Why not derive schemas from this official package as Spectral dependency? https://www.npmjs.com/package/@apidevtools/openapi-schemas — this way scopes in SecurityDefinition (topic for this PR) would land automatically in Spectral.

mikekistler:
@pavelkornev I wholeheartedly support you in wanting OpenAPI to fix their buggy schema.  But please don't hold this PR hostage to the whims and resource limitations of some external team.

P0lip:
> I have also noticed that Spectral OAS3 is refencing to JSON Schema Draft-07 where the official schema is still pointing to Draft-04. Is there a reason for that?

Yeah, the reason here is Ajv. Ajv doesn't support Draft 4 out of the box, so we'd need to use [ajv-draft-04](https://github.com/ajv-validator/ajv-draft-04) or use the costly migration process (this could be that at build time at the expense of having a more complicated pipeline, so it's not that much better for us either).
Back when I made the change the above `ajv-draft-04` didn't exist, so we were left with the latter option.

> In general, I think Spectral should avoid their own OAS schemas implementation. Why not derive schemas from this official package as Spectral dependency? https://www.npmjs.com/package/@apidevtools/openapi-schemas — this way scopes in SecurityDefinition (topic for this PR) would land automatically in Spectral.

Having our own OAS schemas is, alas, somewhat needed.
There's some information in here https://github.com/stoplightio/spectral/pull/1914, but long story short:
- Ajv doesn't like `$dynamicRef` which is used in OAS 3.1 JSON Schema
- there are a few nit-picks I had to manually address. We should port them back to OAS 3.1 JSON Schema, but I didn't have time to do so.

P0lip (approved):

PR #2044, opened 2022-01-31
fix(cli): output to stdout not working with multiple output formatters

### LINKED ISSUE #2042: Output to stdout not working with multiple output formatters
**Describe the bug**

spectral lint fails with error message `The number of outputs must match the number of formats` when on the command line there are two formatters specified but only one output file

**To Reproduce**
1. Given an arbitrary OpenAPI document `my-api.yaml` and any ruleset `.spectral.yaml`
2. Run this CLI command: `spectral lint --format stylish --format junit --output.junit junit-results.xml my-api.yaml`
3. See error  `[usage message omitted] The number of outputs must match the number of formats`

**Expected behavior**

junit formatted output is written to junit-results.xml while stylish formatted output is written to stdout.

Expecting this behavior since the usage message says
` -o, --output                   where to output results, can be a single file name, multiple "output.<format>" or missing to print to stdout`

**Environment**
 - spectral version: 6.2.0
 - OS: Microsoft Windows 10 Enterprise Version 10.0.19042 Build 19042

**Additional context**
No error when formatter count matches output file count, e.g. with
`spectral lint --format stylish --format junit --output.stylish stylish-results --output.junit junit-results.xml my-api.yml`

P0lip:
Fixes #2042 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -54,11 +54,11 @@
     "strip-ansi": "6.0",
     "text-table": "0.2",
     "tslib": "^2.3.0",
-    "yargs": "17.2.1"
+    "yargs": "17.3.1"
   },
   "devDependencies": {
     "@types/xml2js": "^0.4.9",
-    "@types/yargs": "^17.0.5",
+    "@types/yargs": "^17.0.8",
     "copyfiles": "^2.4.1",
     "jest-when": "^3.4.2",
     "nock": "^13.1.3",
--- a/packages/cli/src/commands/__tests__/lint.test.ts
+++ b/packages/cli/src/commands/__tests__/lint.test.ts
@@ -12,9 +12,13 @@ jest.mock('../../services/linter');
 
 function run(command: string) {
   const parser = yargs.command(lintCommand).help();
-  return new Promise(done => {
+  return new Promise((resolve, reject) => {
     parser.parse(command, (err: Error, argv: unknown, output: string) => {
-      done(output);
+      if (err) {
+        reject(`${err.message}\n${output}`);
+      } else {
+        resolve(output);
+      }
     });
   });
 }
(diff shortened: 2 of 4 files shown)


packages/cli/package.json
```diff
@@ -54,11 +54,11 @@
     "strip-ansi": "6.0",
     "text-table": "0.2",
     "tslib": "^2.3.0",
-    "yargs": "17.2.1"
+    "yargs": "17.3.1"
```
marbemac, line 57: Was this a regression with yargs or something?

packages/cli/src/commands/__tests__/lint.test.ts
```diff
@@ -204,7 +207,7 @@ describe('lint', () => {
   });
 
   it('shows help if unknown format is passed', () => {
```
marbemac, line 209: Perhaps pop a test in here to verify something like `spectral lint --format stylish --format junit --output.junit junit-results.xml my-api.yaml` doesn't regress in future?

marbemac (changes requested):

packages/cli/package.json
P0lip: nope, it wasn't, I just updated it while I was at it.

packages/cli/src/commands/__tests__/lint.test.ts
P0lip: we have it already haha :D
See https://github.com/stoplightio/spectral/issues/2042#issuecomment-1026193383

When I made the following change https://github.com/stoplightio/spectral/pull/2044/files#diff-5c1bab693e4315f93ddc3bc536502fe23b536bbed6214ca5a093571db6389f24R14-R21, the test started to fail.

P0lip:
there are no actual changes, I just tested this new beauty 🤓 
image

marbemac (approved):
💯

PR #2046, opened 2022-02-03
feat(rulesets): validate API security in oas-operation-security-defined

mkistler:
This PR adds checking in the function for oas[23]-operation-security-defined to also check the API-level security requirement to ensure it also only references defined security schemes.

I admit that it is stretching the intent of these rules a bit, to check the API-level definition in a rule that by its name would only work on the operation-level security requirement. But it seemed like overkill to define an entirely new rule to perform the same checking as this one, just at the global level.  And in some sense the API level security _is_ the operation-level security for any operation that does not override it, so it didn't seem too unreasonable to include these checks here.

UPDATE: I've updated this PR to add checking for _all_ the fields in a security requirements object -- not just the first one.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [X] No


packages/rulesets/src/oas/functions/oasOpSecurityDefined.ts
```diff
@@ -50,6 +53,31 @@ export default createRulesetFunction<{ paths: Record<string, unknown> }, Options
     const schemes = _get(targetVal, schemesPath);
     const allDefs = isObject(schemes) ? Object.keys(schemes) : [];
 
+    // Check global security requirements
+
+    {
```
P0lip: do we need this block? It doesn't seem like `security` is redeclared within the current scope. It is a later on at https://github.com/stoplightio/spectral/blob/e792e63337288f09b2a33b46cb0b99a256b34f3e/packages/rulesets/src/oas/functions/oasOpSecurityDefined.ts#L84, yet this is another block scope, so we should be good.

mkistler: Amended PR to remove unnecessary block.

P0lip (approved):
LGTM! Thanks a lot.

PR #2048, opened 2022-02-03
feat(ruleset-bundler): a number of minor enhancements

P0lip:
Related to https://github.com/stoplightio/platform-internal/pull/7403

This is an accumulated set of changes I've kept track of after the initial release of the package and the actual integrations attempts in Studio and Spectral CLI.
I suggest reviewing this PR commit by commit because each change is somewhat different.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ruleset-bundler/src/plugins/commonjs.ts
+++ b/packages/ruleset-bundler/src/plugins/commonjs.ts
@@ -1,0 +1,1 @@
+export { default as commonjs } from '@rollup/plugin-commonjs';
--- a/packages/ruleset-bundler/src/utils/dedupeRollupPlugins.ts
+++ b/packages/ruleset-bundler/src/utils/dedupeRollupPlugins.ts
@@ -1,0 +1,12 @@
+// this function makes sure we can only have one plugin with the same name
+// the last plugin definition has a precedence
+import type { Plugin } from 'rollup';
+
+export function dedupeRollupPlugins(plugins: Plugin[]): Plugin[] {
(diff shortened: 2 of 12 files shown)


packages/ruleset-bundler/src/plugins/commonjs.ts
```diff
@@ -0,0 +1 @@
+export { default as commonjs } from '@rollup/plugin-commonjs';
```
marbemac, line 1: How is consumer expected to import this - deep import through `/dist` or something? (or is not meant for general consumption)

packages/ruleset-bundler/src/utils/dedupeRollupPlugins.ts
```diff
@@ -0,0 +1,12 @@
+// this function makes sure we can only have one plugin with the same name
+// the last plugin definition has a precedence
+import type { Plugin } from 'rollup';
+
+export function dedupeRollupPlugins(plugins: Plugin[]): Plugin[] {
```
marbemac, line 5: 👌 very nice

packages/ruleset-bundler/src/plugins/commonjs.ts
P0lip: we have package exports, so can can import this as follows
```ts
import { commonjs } from '@stoplight/spectral-ruleset-bundler/plugins/commonjs';
```

marbemac (dismissed):
👏

PR #2050, opened 2022-02-04
fix(rulesets): operation-tags should fail on empty array

eazy-g:
Because an empty array is truthy in JS, the `operation-tags` rule needs to explicitly check on number of items in the array

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No


P0lip (approved):
Good call, thanks!

PR #2052, opened 2022-02-07
fix(core): consider `message` when de-duplicating results

### LINKED ISSUE #2038: Default "compute fingerprint" function removes unique validations
**Describe the bug**

The [`defaultComputeResultFingerprint`](https://github.com/stoplightio/spectral/blob/develop/packages/core/src/utils/prepareResults.ts#L11) function can produce duplicate hashes for different validation instances, causing one or more instances of unique rules to go unreported. This happens when validations come from the same function (code) and deal with the same artifact in the API definition (path). The other fields used (range, source) aren't specific enough to always identify unique instances.

**To Reproduce**

The [`operation-parameters`](https://github.com/stoplightio/spectral/blob/develop/packages/rulesets/src/oas/functions/oasOpParams.ts) rule in the OAS ruleset is affected by this issue. Given the following API definition:

<details> 
<summary> Sample OpenAPI Document </summary>

```yaml
swagger: "2.0"
info:
  description: "Sample API definition"
  version: "1.0.0"
  title: "Example Document"
  contact:
    email: "support@example.com"
host: "https://example.com/api"
basePath: "/v1"
tags:
- name: "things"
  description: "operations pertaining to things"
paths:
  /thing:
    post:
      tags:
      - "things"
      summary: "Create a new thing"
      description: "Create a new thing"
      operationId: "create_thing"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - in: "body"
        name: "body"
        description: "Thing string"
        schema:
          type: string
      - in: "body"
        name: "body"
        description: "Duplicate thing string"
        schema:
          type: string
      responses:
        400:
          description: "Bad request"
```

</details>

Run `spectral lint test.yaml` and the output will include the following validation:
```
31:9  warning  operation-parameters        A parameter in this operation already exposes the same combination of "name" and "in" values.  paths./thing.post.parameters[1]
```

However, this will be the only validation from the `operation-parameters` rule. In my own tool, which utilizes Spectral, I provided a custom de-duplication strategy and two, separate validations are printed:
```
  Message :   A parameter in this operation already exposes the same combination of "name" and "in" values.
  Rule    :   operation-parameters
  Path    :   paths./thing.post.parameters.1
  Line    :   31

  Message :   Operation must not have more than a single instance of the "in:body" parameter.
  Rule    :   operation-parameters
  Path    :   paths./thing.post.parameters.1
  Line    :   31
```

(As an aside, [the documented method](https://meta.stoplight.io/docs/spectral/ZG9jOjI1MTg3-spectral-in-java-script#using-a-custom-de-duplication-strategy) for providing a custom de-duplication function appears to be incorrect. Passing a config object with a `computeFingerprint` property doesn't seem to be supported, looking at the source code. I had to directly set `_computeFingerprint` in my own code.)

**Expected behavior**

I expect all unique rules to be printed, even if they are similar and operate on the same OpenAPI artifact. As such, I would expect the `message` field of the rule to be taken into account when de-duplicating. This has come up multiple times in my own custom rules/functions.

I realize this behavior may be intentional, since the validations operating on the same OpenAPI artifact could be thought of as redundant. That said, I feel that the user should see everything that is wrong with an artifact at once to avoid fixing something, only to see a new rule come up that could have been shown the first time. If you want to maintain existing behavior, just let me know and I'll continue using my custom function. I wanted to bring it up because it seems like a bug to me and it affects the current Spectral ruleset.

If you're open to this behavior changing, I'm happy to open a PR that addresses the issue.

 - Library version: 1.9.0
 - OS: MacOS Big Sur

dpopp07:
Before this change, the `code` and `path` of the results were the primary factors
considered in de-duplication. Results that came from the same rule and same OpenAPI
artifact but had different messages were de-duplicated and not all unique results
were returned.

This change updates the logic to take `message` into account so that all unique
results are displayed to the user.

I added a new test to confirm the behavior. I also updated an existing test - one of the tests resulted in multiple results returned for the same OpenAPI artifact with different error messages. I looked into addressing the extra rules but in this case, they are a little bit redundant and so I updated the tests to make the additional results expected. Let me know how you feel about this as an approach.

I also removed the outdated docs about custom de-duplication logic.

Fixes #2038

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/rulesets/src/oas/__tests__/duplicated-entry-in-enum.test.ts
+++ b/packages/rulesets/src/oas/__tests__/duplicated-entry-in-enum.test.ts
@@ -42,16 +42,6 @@ testRule('duplicated-entry-in-enum', [
         path: ['definitions', 'Test', 'enum'],
         severity: DiagnosticSeverity.Warning,
       },
-      {
-        message: '"enum" property must not be valid',
-        path: [ 'definitions', 'Test', 'enum' ],
-        severity: DiagnosticSeverity.Warning,
-      },
-      {
-        message: '"enum" property must match exactly one schema in oneOf',
-        path: [ 'definitions', 'Test', 'enum' ],
-        severity: DiagnosticSeverity.Warning,
-      }
     ],
   },
 
@@ -113,16 +103,6 @@ testRule('duplicated-entry-in-enum', [
         path: ['components', 'schemas', 'Test', 'enum'],
         severity: DiagnosticSeverity.Warning,
       },
-      {
-        message: '"enum" property must not be valid',
-        path: [ 'components', 'schemas', 'Test', 'enum' ],
-        severity: DiagnosticSeverity.Warning,
-      },
-      {
-        message: '"enum" property must match exactly one schema in oneOf',
-        path: [ 'components', 'schemas', 'Test', 'enum' ],
-        severity: DiagnosticSeverity.Warning,
-      }
     ],
   },
 ]);
--- a/packages/rulesets/src/oas/__tests__/duplicated-entry-in-enum.test.ts
+++ b/packages/rulesets/src/oas/__tests__/duplicated-entry-in-enum.test.ts
@@ -66,6 +66,16 @@ testRule('duplicated-entry-in-enum', [
         path: ['definitions', 'Test', 'enum'],
         severity: DiagnosticSeverity.Warning,
       },
+      {
+        message: '"enum" property must not be valid',
+        path: [ 'definitions', 'Test', 'enum' ],
+        severity: DiagnosticSeverity.Warning,
+      },
+      {
+        message: '"enum" property must match exactly one schema in oneOf',
+        path: [ 'definitions', 'Test', 'enum' ],
+        severity: DiagnosticSeverity.Warning,
+      }
     ],
   },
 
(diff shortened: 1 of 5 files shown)


P0lip:
Hey!
I just wanted to let you know I noted this PR and it looks good to me.
I want to apply one change to duplicated-entry-in-enum rule, so as soon as I get that sorted out and merged, we'll be able to proceed with your PR.

mkistler:
This is a welcome change! I think many users will benefit from this. Great work @dpopp07!

dpopp07:
@P0lip Sounds great, thank you for the update!

@mkistler Thanks Mike! I hope all is well

packages/rulesets/src/oas/__tests__/duplicated-entry-in-enum.test.ts
```diff
@@ -42,6 +42,16 @@ testRule('duplicated-entry-in-enum', [
...
+      {
+        message: '"enum" property must not be valid',
+        path: [ 'definitions', 'Test', 'enum' ],
+        severity: DiagnosticSeverity.Warning,
+      },
+      {
+        message: '"enum" property must match exactly one schema in oneOf',
+        path: [ 'definitions', 'Test', 'enum' ],
+        severity: DiagnosticSeverity.Warning,
+      }
```
P0lip:
```suggestion
```

packages/rulesets/src/oas/__tests__/duplicated-entry-in-enum.test.ts
```diff
@@ -103,6 +113,16 @@ testRule('duplicated-entry-in-enum', [
...
+      {
+        message: '"enum" property must not be valid',
+        path: [ 'components', 'schemas', 'Test', 'enum' ],
+        severity: DiagnosticSeverity.Warning,
+      },
+      {
+        message: '"enum" property must match exactly one schema in oneOf',
+        path: [ 'components', 'schemas', 'Test', 'enum' ],
+        severity: DiagnosticSeverity.Warning,
+      }
```
P0lip:
```suggestion
```

P0lip (approved):
Thank you!

PR #2068, opened 2022-02-19
docs(rulesets): improve docs for override of particular elements

### LINKED ISSUE #2036: Errors and omissions in documentation of overrides
**Describe the bug**

The [documentation of the Overrides feature](https://meta.stoplight.io/docs/spectral/ZG9jOjI1MTg5-custom-rulesets#overrides) incorrectly states that it supports
> Override rulesets to apply on particular JSONPath's

This is incorrect because Overrides use _JSON Pointers_ and not JSON Paths to identify the elements in the file where overrides should be applied.

It is unfortunate that JSON Pointers are used instead of JSON Paths, since most of the other externals of Spectral use JSON Paths, but there's probably no chance of changing that now.  But all the more reason to document it clearly.  And that should include some pointers to resources on JSON pointers (e.g. how to properly escape slashes).

**To Reproduce**

Look at the documentation.

**Expected behavior**

The documentation is correct and complete (i.e. guides users to success).

**Additional context**

I am willing to submit a PR to fix the doc problems if the Spectral developers will confirm that my diagnosis is accurate.

mkistler:
Fixes #2036.

**Checklist**

- [ ] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/docs/guides/4-custom-rulesets.md
+++ b/docs/guides/4-custom-rulesets.md
@@ -402,38 +402,53 @@ Rulesets can then reference aliases in the [given](#given) keyword, either in fu
 
 Previously Spectral supported exceptions, which were limited in their ability to target particular rules on specific files or parts of files, or changing parts of a rule. Overrides is the much more powerful version of exceptions, with the ability to customize ruleset usage for different files and projects without having to duplicate any rules.
 
-Overrides can be used to:
+Overrides can be used to apply rulesets on:
 
-- Override rulesets to apply on particular files/folders `files: ['schemas/**/*.draft7.json']`
-- Override rulesets to apply on particular JSONPath's `files: ['**#/components/schemas/Item']`
-- Override rulesets to apply on particular formats `formats: [jsonSchemaDraft7]`
+- Particular formats `formats: [jsonSchemaDraft7]`
+- Particular files/folders `files: ['schemas/**/*.draft7.json']`
+- Particular elements of files `files: ['**#/components/schemas/Item']`
 - Override particular rules
 
 **Example**
 
 ```yaml
 overrides:
-  - files:
-      - schemas/**/*.draft7.json
-    formats:
-      - json-schema-draft7
-    rules:
-      valid-number-validation:
-        given:
-          - $..exclusiveMinimum
-          - $..exclusiveMaximum
-        then:
-          function: schema
-          functionOptions:
-            type: number
+  formats:
+    - json-schema-draft7
+  files:
+    - schemas/**/*.draft7.json
+  rules:
+    valid-number-validation:
+      given:
+        - $..exclusiveMinimum
+        - $..exclusiveMaximum
+      then:
+        function: schema
+        functionOptions:
+          type: number
 ```
 
-One can also combine a glob for a filepath with a JSONPath after the anchor, i.e.:
+To apply an override to particular elements of files, combine a glob for a filepath
+with a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) after the anchor, i.e.:
 
 ```yaml
 overrides:
   - files:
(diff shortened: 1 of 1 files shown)


docs/guides/4-custom-rulesets.md
```diff
@@ -402,38 +402,51 @@ Rulesets can then reference aliases in the [given](#given) keyword, either in fu
...
+        - $..exclusiveMinimum
+        - $..exclusiveMaximum
+      then:
+        function: schema
+        functionOptions:
+          type: number
+```
+
+To apply an override to particular elements of files, combine a glob for a filepath
+with a [JSON Pointer](https://github.com/janl/node-jsonpointer) after the anchor, i.e.:
```
mkistler: I chose to point to the npm package used in Spectral for handling JSON pointers, but other options are to point to [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901) or a nice tutorial on JSON pointer (I couldn't find one).

docs/guides/4-custom-rulesets.md
```diff
@@ -402,38 +402,51 @@ Rulesets can then reference aliases in the [given](#given) keyword, either in fu
 
 Previously Spectral supported exceptions, which were limited in their ability to target particular rules on specific files or parts of files, or changing parts of a rule. Overrides is the much more powerful version of exceptions, with the ability to customize ruleset usage for different files and projects without having to duplicate any rules.
 
-Overrides can be used to:
+Overrides can be used to apply rulesets on:
 
-- Override rulesets to apply on particular files/folders `files: ['schemas/**/*.draft7.json']`
-- Override rulesets to apply on particular JSONPath's `files: ['**#/components/schemas/Item']`
-- Override rulesets to apply on particular formats `formats: [jsonSchemaDraft7]`
+- Particular formats `formats: [jsonSchemaDraft7]`
```
mkistler, line 407: I moved formats above files/folders so that the options would be ordered from "general" -> "specific".

docs/guides/4-custom-rulesets.md
```diff
@@ -402,38 +402,51 @@ Rulesets can then reference aliases in the [given](#given) keyword, either in fu
...
-          function: schema
-          functionOptions:
-            type: number
+      some-inherited-rule: "off"
 ```
 
-One can also combine a glob for a filepath with a JSONPath after the anchor, i.e.:
+JSON Pointers have a different syntax than JSON Paths used in the `given` component of a rule.
+In JSON Pointers, path components are prefixed with a "/" and then concatenated to form the pointer.
+Since "/" has a special meaning in JSON pointer, it must be encoded as "~1" when it appears in a component, and "~" must be encoded as "~0".
```
mkistler, line 444: It would have been nice to point to an online JSON pointer evaluator, as we have for JSON Paths, but I could not find a good one.

docs/guides/4-custom-rulesets.md
P0lip: Yeah, that works I think, although we don't use that package directly anywhere IIRC.

docs/guides/4-custom-rulesets.md
P0lip: https://www.jsonquerytool.com/ has `json pointer` to choose from
image

docs/guides/4-custom-rulesets.md
mkistler: In that case I think linking to the RFC is probably better.

mkistler:
Pushed amended commit to address PR comments.

docs/guides/4-custom-rulesets.md
P0lip: Yeah, I like it. They should have some examples there too, so perhaps we link it to that section then?

P0lip (approved):
Thank you!

PR #2079, opened 2022-03-06
fix(functions): reset RegExp.lastIndex to zero when using cached RegExp objects

pavelkornev:
We have noticed that when `g` marker is used the Spectral becomes non-deterministic due to RegExp cache in `pattern` function. So, in our case, validator returns different results on every second run (please see added tests).

More details on RegExp.lastIndex:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex
https://stackoverflow.com/a/13587801/280989

**Quick example:**
```js
const rg = new RegExp("^[\\w]+$", "gi");
rg.exec("123"); // ['123', index: 0, input: '123', groups: undefined]
rg.exec("123"); // null
rg.exec("123"); // ['123', index: 0, input: '123', groups: undefined]
```

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/functions/src/__tests__/pattern.test.ts
+++ b/packages/functions/src/__tests__/pattern.test.ts
@@ -18,6 +18,12 @@ describe('Core Functions / Pattern', () => {
       expect(await runPattern('aBc', { match: '/[abc]+/im' })).toEqual([]);
     });
 
+    it('should return same results when given a global (g) marker (pattern cache usecase)', async () => {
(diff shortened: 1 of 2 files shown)


P0lip (approved):
Nice find!
Thanks a lot.

P0lip:
I pushed a new commit so that CI starts. It got stuck for some reason.

PR #2067, opened 2022-02-19
feat(rulesets): support 2.1.0, 2.2.0, 2.3.0 AsyncAPI versions

magicmatatjahu:
Fixes #1759.

**Checklist**

- [X] Tests added / updated
- [ ] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [X] No

There're not a breaking changes but are some backward compatible named exports.

**Additional context**

This is my first contribution to this project (btw. awesome project!), so if I forgot something please let me know. I don't know if I should describe somewhere in the documentation the support for the latest versions of AsyncAPI, if so please let me know where.

Changes:
- remove JSON Schema for AsyncAPI 2.0.0 from source code - use `@asyncapi/specs` package where there are always up-to-date JSON Schemas.
- adding additional formats like `aas2_0`, `aas2_1` etc.
- adding/updating tests for new formats.
- adding `asyncApi2DocumentSchema` rule identical as for OpenAPI to validate the document using the appropriate JSON Schema version - I know about duplicated code so please let me know if I should create something like `common.ts` file with reusable functions for AsyncAPI and OpenAPI in the `rulesets` package.

NOTE:

I cannot run all tests locally, because I have several failed tests related to the `spectral-functions` package like:

```
  ● Linter service › when loading specification files from web › outputs warnings

    expect(received).resolves.toEqual()

    Received promise rejected instead of resolved
    Rejected to value: [Error: Cannot resolve module '@stoplight/spectral-functions' from paths ['/Users/-/Documents/Github/spectral/packages/cli/src/services/__tests__/__fixtures__', '/Users/-/Documents/Github/spectral/packages/cli/src/services/linter/utils'] from /Users/-/Documents/Github/spectral/packages/cli/src/services/__tests__/__fixtures__/spectral.js]

      299 |       });
      300 |
    > 301 |       return expect(run(`lint http://foo.local/openapi`)).resolves.toEqual([
          |              ^
      302 |         {
      303 |           code: 'info-matches-stoplight',
      304 |           message: 'Info must contain Stoplight',

      at expect (node_modules/expect/build/index.js:178:15)
      at Object.<anonymous> (packages/cli/src/services/__tests__/linter.test.ts:301:14)
``` 

Is a known issue, or perhaps it's related to my environment? 🤷🏼

PR #2082, opened 2022-03-07
docs: add all AsyncAPI formats

jonaslagoni:
After https://github.com/stoplightio/spectral/pull/2067 I think this change was missed 🙂

Technically I think `asyncapi2` [is still supported](https://github.com/stoplightio/spectral/blob/48395273d28fc58ab450277681956b1b50f3f8ee/packages/formats/src/asyncapi.ts#L19) as is `asyncApi2` (which is not present in the docs) 🤔

I decided to remove it for the docs, not sure if that is desirable?  

##### Checklist

- [ ] Tests added / updated
- [X] Docs added / updated

##### Does this PR introduce a breaking change?

- [X] Yes
- [X] No

See above.

### DIFF
--- a/docs/getting-started/3-rulesets.md
+++ b/docs/getting-started/3-rulesets.md
@@ -75,7 +75,11 @@ The `extends` keyword can be combined with extra rules in order to extend and ov
 
 Formats are an optional way to specify which API description formats a rule, or ruleset, is applicable to. Currently Spectral supports these formats:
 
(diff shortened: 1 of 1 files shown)


P0lip (approved):
Thanks!

Re `asyncapi2` yeah, we gotta unify the naming somehow.

I guess it's probably best to use asyncapi in favor of aas, but that's something for the next major release.

PR #2084, opened 2022-03-08
chore(repo): add CODEOWNERS

P0lip:


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+* @stoplightio/team-plaid
+/packages/rulesets/src/asyncapi  <add-asyncapi-team>
```
P0lip: @marbemac could you make sure to create some team and add https://github.com/magicmatatjahu?
@jonaslagoni would you like to be included as well?

magicmatatjahu: Sorry for such a delay! Thanks a lot! I feel honored :) If it would be possible @smoya himself was also interested.

jonaslagoni: @P0lip it would be an honor 🙂

smoya: >  If it would be possible @smoya himself was also interested.

It would be amazing! Very much appreciated if you do @P0lip !

marbemac (approved):

.github/CODEOWNERS
P0lip: The invites have been sent. LMK in case you haven't received one.

PR #2097, opened 2022-03-17
feat(rulesets): add unused components server rule

jonaslagoni:
Partly solves https://github.com/stoplightio/spectral/issues/1103

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No

### DIFF
--- a/docs/reference/asyncapi-rules.md
+++ b/docs/reference/asyncapi-rules.md
@@ -369,3 +369,15 @@ that acts as a library (a container storing reusable objects, leveraged by other
 specifications that reference those objects).
 
 **Recommended:** Yes
+
+### asyncapi-unused-components-server
+
+Potential unused reusable `server` entry has been detected.
+
+<!-- theme: warning -->
+
+_Warning:_ This rule may identify false positives when linting a specification
+that acts as a library (a container storing reusable objects, leveraged by other
+specifications that reference those objects).
+
+**Recommended:** Yes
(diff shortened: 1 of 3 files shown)


P0lip:
@jonaslagoni assuming you've received an invite and accepted it (which seems like you do according to what GH settings page tells me), perhaps create a branch in the Spectral repo itself?
You're welcome to create branches directly in the repo.
This should help with CI checks and our CircleCI setup doesn't like forks particularly much.

jonaslagoni:
> You're welcome to create branches directly in the repo.
> This should help with CI checks and our CircleCI setup doesn't like forks particularly much.

Haha, alright, I'll try to remember. It's just the standard we apply under AsyncAPI and the one that creates less clutter in upstream 😄 

I was wondering why the jobs was not picked up by the CI 😅

P0lip:
> Haha, alright, I'll try to remember. It's just the standard we apply under AsyncAPI and the one that creates less clutter in upstream 😄

Well, you can use the fork if that's how you approach it. I just wanted to let you know you can create branches here too, but it's not compulsory by any means, so you can continue to use the fork if that's how you do things.

> I was wondering why the jobs was not picked up by the CI 😅

It should work just fine and it usually does, but for some reason, it got stuck this time around.
If that continues to happen, I'll spend some time investigating that.
I know we had a similar issue in https://github.com/stoplightio/prism for certain contributors, so it's not an isolated case.

P0lip (approved):
It seems like I'll need to merge this one cause of the CI.
LMK when you're good with the change, and I'll do it.
it's 👌 from my side

jonaslagoni:
> It seems like I'll need to merge this one cause of the CI.
> LMK when you're good with the change, and I'll do it.
> it's 👌 from my side

Fine by me yea 👍

jonaslagoni:
I think we should block this until https://github.com/stoplightio/spectral/issues/1103#issuecomment-1071788756 is resolved, otherwise it would be a breaking-change to change it 🙂

Issue #2100, opened 2022-03-21
Support all custom validation for AsyncAPI 2.x.x


magicmatatjahu:
**Chore summary**

This issue contains list of rules that we need introduce in the AsyncAPI ruleset to fullfil all custom validation for 2.x.x versions.

> Note: This is a collective task, which serves for cooperation and informing the community what was done and what is left to do.

> Note: You can also write your own proposals/features/PRs in the comments, I will update then the main list.

**Tasks**

- [x] https://github.com/stoplightio/spectral/issues/1102
  - https://github.com/stoplightio/spectral/pull/2101
- [ ] https://github.com/stoplightio/spectral/issues/1103
- [x] https://github.com/stoplightio/spectral/issues/1732
- [x] https://github.com/stoplightio/spectral/issues/2102
  - https://github.com/stoplightio/spectral/pull/2121
- [x] https://github.com/stoplightio/spectral/issues/2103
  - https://github.com/stoplightio/spectral/pull/2104
- [x] https://github.com/stoplightio/spectral/issues/2123
  - https://github.com/stoplightio/spectral/pull/2122
- [x] https://github.com/stoplightio/spectral/issues/2124
  - https://github.com/stoplightio/spectral/pull/2122
- [x] https://github.com/stoplightio/spectral/issues/2223
  - https://github.com/stoplightio/spectral/pull/2224
- [ ] https://github.com/stoplightio/spectral/issues/2261
  - https://github.com/stoplightio/spectral/pull/2262
- https://github.com/stoplightio/spectral/pull/2281
- https://github.com/stoplightio/spectral/pull/2282
- https://github.com/stoplightio/spectral/pull/2284
- ...

**Additional context**

As the AsyncAPI team will take an active part in the development of the official ruleset for AsyncAPI we should support the validation of the overall specification.

derberg:
we had a release where uniqueness of `messageId` was introduced

PR #2104, opened 2022-03-22
feat(rulesets): add rules for validation uniqueness of tag names

### LINKED ISSUE #2103: Add rule to check uniqueness of `tags.*.name`
**User story.**
As a user, I want to know if the tags defined in my AsyncAPI document (in any place) have a unique names.

The following AsyncAPI document is invalid and Spectral should throw errors due to non-uniqueness of `tags.*.name`.

```yaml
asyncapi: '2.3.0'
...

tags:
  - name: someTag
    ...
  - name: someTag
    ...
```

**Additional context**
Part of #2100

magicmatatjahu:
Fixes https://github.com/stoplightio/spectral/issues/2103

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No

**Additional context**

Part of #2100 

Add rule for validation uniqueness of tag names:

**Bad Example**

```yaml
tags:
  - name: "Badger"
  - name: "Badger"
```

**Good Example**

```yaml
tags:
  - name: "Aardvark"
  - name: "Badger"
```


packages/rulesets/src/asyncapi/functions/asyncApi2UniquenessTags.ts
```diff
@@ -0,0 +1,33 @@
...
+  const tagNames = tags.map(item => item.name);
+  return tagNames.reduce((acc, item, idx, arr) => {
+    if (arr.indexOf(item) !== idx && acc.indexOf(item) < 0) {
+      acc.push(item);
+    }
+    return acc;
+  }, [] as string[]);
+}
+
+export default createRulesetFunction<Array<{ name: string }>, null>(
```
marbemac: Interesting - so as far as I can tell, this function is equally applicable / useful in the openapi context (and this same rule, with different `given`, would be valuable in the openapi context). @magicmatatjahu @P0lip what do you think re sharing common functions across default rulesets? I know that we have `@stoplight/spectral-functions`, but the functions in there are very generic, while tags uniqueness is a bit more specific but generic across openapi and asyncapi 🤔.

P0lip: I agree, it's something that could be potentially shared.
We could create a directory under rulesets package for shared utils.

packages/rulesets/src/asyncapi/functions/asyncApi2UniquenessTags.ts
```diff
@@ -0,0 +1,33 @@
...
+  {
+    input: null,
+    options: null,
+  },
+  function asyncApi2UniquenessTags(targetVal, _, ctx) {
+    if (!targetVal || targetVal.length === 0) return [];
```
P0lip: one can avoid that if you add a schema for the input

packages/rulesets/src/asyncapi/functions/asyncApi2UniquenessTags.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (arr.indexOf(item) !== idx && acc.indexOf(item) < 0) {
+      acc.push(item);
+    }
+    return acc;
+  }, [] as string[]);
+}
+
+export default createRulesetFunction<Array<{ name: string }>, null>(
+  {
+    input: null,
```
P0lip:
```suggestion
    input: { 
      type: 'array',
      items: {
        type: 'object',
        properties: {
          name: {
            type: 'string',
          },
        },
      },
    },
```

packages/rulesets/src/asyncapi/functions/asyncApi2UniquenessTags.ts
```diff
@@ -0,0 +1,33 @@
...
+  },
+  function asyncApi2UniquenessTags(targetVal, _, ctx) {
+    if (!targetVal || targetVal.length === 0) return [];
+
+    const duplicatedTags = getDuplicateTagNames(targetVal);
+    if (!duplicatedTags || duplicatedTags.length === 0) return [];
+
+    return [
+      {
+        message: `Tags contains duplicate tag names: ${duplicatedTags.join(', ')}.`,
```
P0lip: wouldn't you like to report a duplicate for each particular tag?
Say, you have 10 tags, and only one of them is a dupe. In such a case if you pointed at a particular tag, the error would be slightly more useful.

magicmatatjahu: Yeah, good idea. I changed error's message and path to something like that:

```
          message: `"tags" object contains duplicate tag name "someTag".`,
          path: [...ctx.path, '1', 'name'],
```

What do you think? Is it ok for you, or we should only point to the index of tag, not to the tag's name?

packages/rulesets/src/asyncapi/functions/asyncApi2UniquenessTags.ts
magicmatatjahu: I created `shared` folder in the `rulesets` packages with `uniquenessTags` rule. I also create `openapi-tags-uniqueness` rule with this same behaviour as for AsyncAPI. If you would like to introduce that rule in separate PR, please let me know, I will remove that.

docs/reference/openapi-rules.md
```diff
@@ -250,6 +250,28 @@ tags:
   - name: "Badger"
 ```
 
+### asyncapi-tags-uniqueness
+
+OpenAPI object should have non-duplicated tag names (identifiers).
```
P0lip:
```suggestion
OpenAPI object must not have duplicated tag names (identifiers).
```

docs/reference/asyncapi-rules.md
```diff
@@ -341,6 +341,28 @@ tags:
 
 **Recommended:** No
 
+### asyncapi-tags-uniqueness
+
+Tags have duplicate names (identifiers).
```
P0lip:
```suggestion
Tags must not have duplicate names (identifiers).
```

packages/rulesets/src/oas/index.ts
```diff
@@ -212,6 +213,17 @@ const ruleset = {
         },
       },
     },
+    'openapi-tags-uniqueness': {
+      description: 'Each tags must have a unique names.',
```
P0lip:
```suggestion
      description: 'Each tag must have a unique name.',
```

packages/rulesets/src/asyncapi/functions/asyncApi2DocumentSchema.ts
```diff
@@ -91,7 +91,7 @@ export default createRulesetFunction<unknown, null>(
     input: null,
     options: null,
   },
-  function oasDocumentSchema(targetVal, _, context) {
+  function asyncApi2DocumentSchema(targetVal, _, context) {
```
P0lip, line 97: good catch 😄

packages/rulesets/src/asyncapi/index.ts
```diff
@@ -358,6 +359,40 @@ export default {
         },
       },
     },
+    'asyncapi-tags-uniqueness': {
+      description: 'Each tags must have a unique names.',
```
P0lip:
```suggestion
      description: 'Each tag must have a unique name.',
```

packages/rulesets/src/shared/functions/uniquenessTags.ts
```diff
@@ -0,0 +1,61 @@
...
+        ...duplicatedTags.map(duplicatedIndex => ({
+          message: `"tags" object contains duplicate tag name "${duplicatedTag}".`,
+          path: [...ctx.path, duplicatedIndex, 'name'],
+        })),
+      );
+    });
+
+    return results;
+  },
+);
```
P0lip: How about this?
```suggestion
import { createRulesetFunction, IFunctionResult } from '@stoplight/spectral-core';

type Tags = Array<{ name: string }>;

function getDuplicateTags(tags: Tags): number[] {
  return tags
    .map(item => item.name)
    .reduce<number[]>((acc, item, i, arr) => {
      if (arr.indexOf(item) !== i) {
        acc.push(i);
      }

      return acc;
    }, []);
}

export default createRulesetFunction<Tags, null>(
  {
    input: {
      type: 'array',
      items: {
        type: 'object',
        properties: {
          name: {
            type: 'string',
          },
        },
      },
    },
    options: null,
  },
  function uniquenessTags(targetVal, _, ctx) {
    const duplicatedTags = getDuplicateTags(targetVal);
    if (duplicatedTags.length === 0) return;

    const results: IFunctionResult[] = [];

    for (const duplicatedIndex of duplicatedTags) {
      const duplicatedTag = targetVal[duplicatedIndex].name;

      results.push({
        message: `"tags" object contains duplicate tag name "${duplicatedTag}".`,
        path: [...ctx.path, duplicatedIndex, 'name'],
      });
    }

    return results;
  },
);
```

P0lip (commented):
Looking good overall.

packages/rulesets/src/shared/functions/uniquenessTags.ts
magicmatatjahu: @P0lip Yes, it is much easier, thank you very much! Suggestion applied.

magicmatatjahu:
@P0lip Sorry for such a delay! All suggestions have been applied. Let me know if everything is ok and if it is ok, you can merge when you want :) Unfortunately I had to do a mixed reset + force push (it was the easiest way for me) because the commit linting was showing me errors when I merged from the Github UI `develop` branch to my PR.

packages/rulesets/src/shared/functions/uniquenessTags.ts
```diff
@@ -0,0 +1,49 @@
...
+  {
+    input: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          name: {
+            type: 'string',
+          },
+        },
```
P0lip, line 28:
```suggestion
        properties: {
          name: {
            type: 'string',
          },
        },
        required: ['name'],
```

P0lip (dismissed):
Looking good 🚀

P0lip (approved):

PR #2151, opened 2022-05-05
feat(core): relax formats validation

P0lip:
If a new format is added to stoplight/spectral-formats, the ruleset meta schema would need to be updated accordingly.
By relaxing the imposed constraints, that whole process is a tad simpler.
We'd inform users about invalid formats, albeit the error is likely to be a bit vaguer, as we wouldn't print the list of all available formats.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


packages/core/src/ruleset/meta/json-extensions.json
```diff
@@ -36,22 +36,7 @@
     },
     "Format": {
       "$anchor": "format",
-      "enum": [
```
P0lip, line 39: @mnaumanali94 you good with that change?

mnaumanali94 (approved):

mnaumanali94 (approved):

PR #2164, opened 2022-05-23
feat(core): improve alias validation

### LINKED ISSUE #2154: Move alias validation from runtime to static
**Chore summary**
Currently if you use an alias in `given` that isn't defined, spectral detects the issue while running `lint` against an openapi file but not while validating the ruleset. 

**Tasks**
- [ ] Validation should fail if an alias is being used that isn't defined

P0lip:
Closes #2154 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Changes**

- [extracted](https://github.com/stoplightio/spectral/pull/2164/files#diff-2ac12b44fc3e0b0c43b42e64947c022ccc9543625ad3b6a6efcba574219f8068L154) `resolveAlias` method out of the `Rule` class and moved it to a [separate file](https://github.com/stoplightio/spectral/blob/209fdd558d687c4c82114a1fd68bbba6287f84ab/packages/core/src/ruleset/alias.ts) so it can be used elsewhere
  - the function itself remained pretty much unchanged
 
- `x-spectral-runtime` keyword takes new option, namely `alias` -> https://github.com/stoplightio/spectral/blob/209fdd558d687c4c82114a1fd68bbba6287f84ab/packages/core/src/ruleset/validation/ajv.ts#L62
  - a side change was the adaption of the same validation approach for functions, so that it uses a similar mechanism, [here](https://github.com/stoplightio/spectral/pull/2164/files#diff-29b1db27ac801e0dec894b898b15a315e649cfd54e5280cc45aaf5a800ce63bfR52-R61) 
  
- adjusted [packages/core/src/ruleset/meta/shared.json](https://github.com/stoplightio/spectral/blob/209fdd558d687c4c82114a1fd68bbba6287f84ab/packages/core/src/ruleset/meta/shared.json) to make use of new option ^
  - almost all changes in tests are related to the schema changes I've made


docs/guides/4-custom-rulesets.md
```diff
@@ -396,6 +396,48 @@ aliases:
...
+        then:
+          field: description
+          function: pattern
+          functionOptions:
+            match: Stoplight
+  - files:
+      - "**/*.json"
+    aliases:
+      Value:
+        - "$..value"
```
marbemac, line 442: Might be worth adding a note re override behavior. E.g. - if an alias called `Stoplight` were defined here, would it take precedence over the `Stoplight` alias defined in the root?

packages/core/src/ruleset/meta/shared.json
```diff
@@ -50,9 +50,26 @@
     },
     "PathExpression": {
       "$id": "path-expression",
-      "type": "string",
-      "pattern": "^[$#]",
-      "errorMessage": "must be a valid JSON Path expression or a reference to the existing Alias optionally paired with a JSON Path expression subset"
+      "if": {
+        "type": "string"
+      },
+      "then": {
```
marbemac, line 56: This conditional logic stuff in json schema is cool and terrifying at the same time.

marbemac (commented):
Let's carve out 20 min to walk through this one on a call, I'd like to understand a few bits better.

chohmann:
I don't know much about Spectral... not sure if I'm the best person to review this or not.

P0lip:
No worries, you can totally skip it.
I believe you were added because you're a member of`stoplight/documentation` and this group is now a code owner of the `docs/` directory.

chohmann:
ahhh ok that makes sense. Thanks @P0lip !

mmiask:
I'm having trouble understanding the changes. @P0lip  - what could speed up the review is some kind of changes summary, describing the main change and how other files relate to that. Would you mind putting such summary in PR description? Or if it is better for you, discuss it on a call and do live PR review?

mmiask (approved):
Read through this PR with @P0lip on a call. Awesome work, looks great 👍

Issue #2107, opened 2022-03-24
fails to run base on js format ruleset 


jianyexi:
**Describe the bug**

**To Reproduce**

1. Given this ruleset file
```
> cat myruleset.js 
import { oas2, oas3 } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-formats/+esm";
import { falsy } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-functions@1.6.1/+esm";
import { oas } from "https://cdn.jsdelivr.net/npm/@stoplight/spectral-rulesets/+esm";
const ruleset = {
    extends: [
        oas
    ],
    rules: {
        "info-contact": "off",
        "no-$ref-siblings": "off",
        "additional-properties-and-properties": {
            "description": "Don't specify additionalProperties as a sibling of properties.",
            "severity": "warn",
            "formats": [oas2, oas3],
            "given": "$..[?(@object() && @.type === 'object' && @.properties)]",
            "then": {
                "field": "additionalProperties",
                "function": falsy
            }
        },
    }
};
export default ruleset;
```
2. run 
```
> spectral lint -r myruleset.js  swagger.json
```
3. See error
```
Error running Spectral!
Error #1: 'isObject' is not exported by https://cdn.jsdelivr.net/npm/lodash@4.17.21/+esm, imported by https://cdn.jsdelivr.net/npm/@stoplight/json@3.17.2/+esm
          at net/npm/@stoplight/…
          at error                 …hared/rollup.js:160  base = Object.assig…
          at error                 …red/rollup.js:12436  return error(props);
          at traceVariable         …red/rollup.js:12797  return this.error(e…
          at findVariable          …red/rollup.js:11419  const variable = th…
```

**Expected behavior**
return the linting result

jianyexi:
add another case:
1. Given this remote ruleset file
```
> cat    https://foo/myruleset.js 
var spectralFormats = require('@stoplight/spectral-formats');
var spectralFunctions = require('@stoplight/spectral-functions');
const ruleset = {
    extends: [
    ],
    rules: {
        "info-contact": "off",
        "no-$ref-siblings": "off",
        "additional-properties-and-properties": {
            "description": "Don't specify additionalProperties as a sibling of properties.",
            "severity": "warn",
            "formats": [spectralFormats .oas2, spectralFormats .oas3],
            "given": "$..[?(@object() && @.type === 'object' && @.properties)]",
            "then": {
                "field": "additionalProperties",
[...]

PR #2174, opened 2022-06-03
fix(ruleset-bundler): never externalize builtins

P0lip:
Fixes the 2nd case in #2107 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

This is somewhat similar to https://github.com/stoplightio/spectral/pull/2109 with that difference we alter the default resolution only if the builtins plugin is registered. Otherwise, externalization is not modified.


packages/ruleset-bundler/src/__tests__/index.jest.test.ts
```diff
@@ -97,4 +98,50 @@ var spectral = {
 
 export { spectral as default };`);
   });
+
+  it('given node target, should support commonjs for remote ruleset with builtin modules', async () => {
```
P0lip, line 102: taken from https://github.com/stoplightio/spectral/pull/2109

mnaumanali94 (approved):

PR #2212, opened 2022-07-22
Fix unused percent encoded

mkistler:
This PR fixes the used component rules to handle percent-encoded refs to components.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


P0lip (approved):
Nice catch

PR #2243, opened 2022-08-15
fix(rulesets): __importDefault undefined

### LINKED ISSUE #2240: Issue importing oas ruleset from @stoplight/spectral-rulesets
**Describe the bug**
When trying to use the oas ruleset exported by the @stoplight/spectral-rulesets package in a backstage.io custom plugin an error occurs that does not let me render my frontend page or use the ruleset.

The error that is displayed on the webpage is: 
`ERROR 501: Lazy loaded chunk failed to load, try to reload the page: __importDefault is not defined`

Here is how the packages are being imported:
```
import { Spectral, Document } from "@stoplight/spectral-core";
import { oas } from "@stoplight/spectral-rulesets";
```
Ive also tried importing it using "require" but get the same error when trying that way.

Here is how the packages are being used:
```    
console.log(oas);
const myRuleset = {
    extends: [oas],
    rules: {
        "no-empty-description": {
            given: "$..description",
            message: "Description must not be empty",
            then: {
                function: truthy,
            },
        },
    },
}
useEffect(() => {
    spectral.setRuleset(myRuleset);
    console.log(spectral.ruleset);
    spectral.run(myDocument).then(setLintIssues);
}, []);
```

**Environment:**
Node: v1.15.1
Typescript: 4.6.4
Spectral-core: 1.13.0
Spectral-functions: 1.7.0
Spectral-parsers: 1.0.1
Spectral-ruleset-bundler: 1.3.1
Spectral-rulesets: 1.11.1
OS: MacOS Monterey
Browser: Chrome 103

P0lip:
Fixes https://github.com/stoplightio/spectral/issues/2240

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/rulesets/src/shared/functions/index.ts
+++ b/packages/rulesets/src/shared/functions/index.ts
@@ -1,1 +1,3 @@
-export { default as uniquenessTags } from './uniquenessTags';
+import { default as uniquenessTags } from './uniquenessTags';
+
+export { uniquenessTags };
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -44,7 +44,7 @@
     "declaration": true,
     "removeComments": true,
     "skipLibCheck": true,
-    "noEmitHelpers": true,
+    "noEmitHelpers": false,
     "importHelpers": true
   }
 }


packages/rulesets/src/shared/functions/index.ts
```diff
@@ -1 +1,3 @@
-export { default as uniquenessTags } from './uniquenessTags';
+import { default as uniquenessTags } from './uniquenessTags';
+
+export { uniquenessTags };
```
P0lip, line 3: this results in a slightly smaller output, as no `__importDefault` function is declared

tsconfig.json
```diff
@@ -44,7 +44,7 @@
     "declaration": true,
     "removeComments": true,
     "skipLibCheck": true,
-    "noEmitHelpers": true,
+    "noEmitHelpers": false,
```
P0lip, line 47: but in case someone does want to use `export { default as xyz }`, we're now covered and the function will be output.

PR #2262, opened 2022-08-29
feat(rulesets): validate unresolved AsyncAPI document

### LINKED ISSUE #2261: Add a rule to validate unresolved AsyncAPI document
**Chore summary**

As a user I want to know if I defined `$ref` fields in proper places in my AsyncAPI document

The following AsyncAPI document is invalid and Spectral should throw errors due to invalid defined references (in impossible places):

```yaml
asyncapi: '2.0.0'
channels: 
  someChannel:
    publish:
      $ref: '#/components/x-operations/someOperation'
components:
  'x-operations':
    someOperation: ...
```

We cannot use `$ref` in operations.

**Additional context**
Part of #2100

magicmatatjahu:
Fixes #2261.

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No

**Screenshots**

The following AsyncAPI document is invalid and Spectral should throw errors due to invalid defined references (in impossible places):

```yaml
asyncapi: '2.0.0'
channels: 
  someChannel:
    publish:
      $ref: '#/components/x-operations/someOperation'
components:
  'x-operations':
    someOperation: ...
```

and Spectral will return errors:

```yaml
[
  {
    message: 'Referencing here is not allowed',
    path: ['channels', 'someChannel', 'publish', '$ref'],
    severity: DiagnosticSeverity.Error,
  }
]
```

### DIFF
--- a/packages/functions/src/optionSchemas.ts
+++ b/packages/functions/src/optionSchemas.ts
@@ -172,6 +172,7 @@ export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
         type: 'object',
         description:
           'Assigns a unique id (by reference to the JS object) to a schema. It is used to optimize the creation of a validation function for a given schema; a given function will be stored by given id and retrieved between execution of validation.',
+        'x-internal': true,
       },
       prepareResults: {
         'x-internal': true,
--- a/packages/functions/src/optionSchemas.ts
+++ b/packages/functions/src/optionSchemas.ts
@@ -168,6 +168,11 @@ export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
         default: false,
         description: 'Returns all errors when true; otherwise only returns the first error.',
       },
+      uniqueId: {
+        type: 'object',
+        description:
+          'Assigns a unique id (by reference to the JS object) to a schema. It is used to optimize the creation of a validation function for a given schema; a given function will be stored by given id and retrieved between execution of validation.',
+      },
       prepareResults: {
         'x-internal': true,
       },
--- a/packages/rulesets/src/asyncapi/functions/asyncApi2DocumentSchema.ts
+++ b/packages/rulesets/src/asyncapi/functions/asyncApi2DocumentSchema.ts
@@ -104,7 +104,7 @@ function getSchema(formats: Set<Format>): Record<string, unknown> | void {
   }
 }
 
-// For optimizing the retrieving/creation of AJV's validation funcition for a given AsyncAPI version.
+// For optimizing the retrieving/creation of AJV's validation function for a given AsyncAPI version.
 // Currently each validation run creates a separate `documentInventory` (needed by `schemaFn`), which serves as an identifier for the weakMap's element of available AJV's validation functions.
 // This variable will always be the same for each validation run, regardless of the number of runs or Spectral instances.
 const CONST_DOCUMENT_INVENTORY: RulesetFunctionContext['documentInventory'] =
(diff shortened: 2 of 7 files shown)


packages/rulesets/src/asyncapi/functions/asyncApi2DocumentSchema.ts
```diff
@@ -89,24 +101,51 @@ function getSchema(formats: Set<Format>): Record<string, unknown> | void {
   }
 }
 
-export default createRulesetFunction<unknown, null>(
+// For optimizing the retrieving/creation of AJV's validation funcition for a given AsyncAPI version.
+// Currently each validation run creates a separate `documentInventory` (needed by `schemaFn`), which serves as an identifier for the weakMap's element of available AJV's validation functions.
+// This variable will always be the same for each validation run, regardless of the number of runs or Spectral instances.
+const CONST_DOCUMENT_INVENTORY: RulesetFunctionContext['documentInventory'] =
+  {} as RulesetFunctionContext['documentInventory'];
```
magicmatatjahu: There I have big problem. Due to fact that `schema` function from `@stoplight/spectral-functions` uses the context's `documentInventory` reference to save and reuse AJV's validation function for given schema I made that above trick to optimize validation. Should we add option to the `schema` function to allow above logic in a simpler way by some `id`? 

cc @P0lip

P0lip: Agreed, that'd be a nice addition

jonaslagoni:
@magicmatatjahu is this ready to be reviewed or do you still need to do some work regarding https://github.com/stoplightio/spectral/pull/2262#discussion_r957380757?

magicmatatjahu:
@jonaslagoni Yeah, when I will find time, I will create PR with `uniqueId` for schema function.

packages/functions/src/optionSchemas.ts
```diff
@@ -168,6 +168,11 @@ export const optionSchemas: Record<string, CustomFunctionOptionsSchema> = {
         default: false,
         description: 'Returns all errors when true; otherwise only returns the first error.',
       },
+      uniqueId: {
+        type: 'object',
+        description:
+          'Assigns a unique id (by reference to the JS object) to a schema. It is used to optimize the creation of a validation function for a given schema; a given function will be stored by given id and retrieved between execution of validation.',
+      },
```
magicmatatjahu: Due to fact that `instances` in the `schema` function is a WeakMap's instance, we cannot use `string` type only `object` or `array`, which are references in JS. WDYT about it? I don't think so that we should change `instances` to the Map's instance (to support strings), because then we can end up with stackoverflow.

cc @P0lip

P0lip: Fair enough. Just make sure to add `x-internal` to indicate the option isn't meant to be used in rulesets.
```suggestion
      uniqueId: {
        type: 'object',
        description:
          'Assigns a unique id (by reference to the JS object) to a schema. It is used to optimize the creation of a validation function for a given schema; a given function will be stored by given id and retrieved between execution of validation.',
        'x-internal': true,
      },
```

magicmatatjahu: Thanks! Pushed.

jonaslagoni:
Any update on this one @magicmatatjahu? Or do you want to close it?

mnaumanali94:
Closing as stale. Feel free to open if needed. @jonaslagoni

PR #2281, opened 2022-09-16
feat(rulesets): add rule to check if messageId is defined

magicmatatjahu:
**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No

**Additional context**

- change severity of `asyncapi-operation-operationId` rule from error to warning
- add `asyncapi-message-messageId` rule, with similar functionality like `asyncapi-operation-operationId`
- add unit tests
- update docs
- add more formats (needed by new rule)
- merge traits before checking ids.

cc @jonaslagoni @smoya

### DIFF
--- a/docs/reference/asyncapi-rules.md
+++ b/docs/reference/asyncapi-rules.md
@@ -261,7 +261,7 @@ channels:
 
 ### asyncapi-message-messageId
 
-Each Message should have a "messageId" field defined. It is recommended, but still optional. Tools can use it to define function names, class method names and even URL passwords in documentation systems.
+Each Message should have a "messageId" field defined. It is recommended, but still optional. Tools can use it to define function names, class method names and even URL hashes in documentation systems.
 
 **Recommended:** Yes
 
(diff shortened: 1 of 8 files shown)


magicmatatjahu:
I have to update that PR to handle also cases with traits. For `messageId` and `operationId`.

smoya (commented):
💯 👍 

I +1 to the change. I'm concerned about the possible breaking change this could become. 

If any user relies on the `operationId` for a critical feature, this change will possibly make their apps to fail. 
I believe this should be considered a breaking change, but let's see what other maintainers say.

P0lip:
Yeah, I'd say it's a breaking change indeed.
As such, we could release `spectral-rulesets` 2.x.x if you're okay with.
I don't have any issue with it.

magicmatatjahu:
@smoya @P0lip I'm not a fan to release version 2.0.0 rulesets with such a case. As there will be much bigger changes then we will change error to warning, now I reverted my changes regarding `operationId`. Please check again, thanks!

docs/reference/asyncapi-rules.md
```diff
@@ -297,7 +303,7 @@ channels:
 
 ### asyncapi-operation-operationId
 
-This operation ID is essentially a reference for the operation. Tools may use it for defining function names, class method names, and even URL hashes in documentation systems.
+Each Operation must have an "operationId" field defined. Tools can use it to define function names, class method names and even URL hashes in documentation systems.
```
philsturgeon, line 306:
```suggestion
Each Operation must have an "operationId" field defined. Tools may use it to define function names, class method names, and even URL hashes in documentation systems.
```

philsturgeon (approved):

PR #2284, opened 2022-09-19
fix(rulesets): handle empty payload and headers in AsyncAPI message's examples validation

magicmatatjahu:
**Checklist**

- [X] Tests added / updated
- [ ] Docs added / updated - not needed

- [ ] Yes
- [X] No

The current rule for AsyncAPI message's examples not support use cases where payloads or headers may be empty. Also it does not merge traits in first row and then validate. That PR fix these two "bugs".

### DIFF
--- a/packages/rulesets/src/asyncapi/functions/utils/mergeTraits.ts
+++ b/packages/rulesets/src/asyncapi/functions/utils/mergeTraits.ts
@@ -1,0 +1,36 @@
+import { isPlainObject } from '@stoplight/json';
+
+type HaveTraits = { traits?: any[] } & Record<string, any>;
+
+export function mergeTraits<T extends HaveTraits>(data: T): T {
+  if (Array.isArray(data.traits)) {
+    data = { ...data }; // shallow copy
+    for (const trait of data.traits as T[]) {
+      for (const key in trait) {
+        data[key] = merge(data[key], trait[key]);
+      }
+    }
+  }
+  return data;
+}
+
+function merge<T>(origin: unknown, patch: unknown): T {
+  // If the patch is not an object, it replaces the origin.
+  if (!isPlainObject(patch)) {
+    return patch as T;
+  }
+
+  const result = !isPlainObject(origin)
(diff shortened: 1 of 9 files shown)


packages/rulesets/src/asyncapi/functions/utils/mergeTraits.ts
```diff
@@ -0,0 +1,36 @@
+import { isPlainObject } from '@stoplight/json';
+
+type HaveTraits = { traits?: any[] } & Record<string, any>;
+
+export function mergeTraits<T extends HaveTraits>(data: T): T {
```
jonaslagoni, line 12: I would probably add some more docs here and specifically stating that this is JSON Merge Patch.

https://www.rfc-editor.org/rfc/rfc7386

jonaslagoni (changes requested):
Just one comment in terms of traits, as this makes it easier for developers that stumble upon the code to follow it.

Otherwise it looks good 👍

packages/rulesets/src/asyncapi/functions/utils/mergeTraits.ts
magicmatatjahu: Done, thanks! :)

jonaslagoni (dismissed):
👍

jonaslagoni (dismissed):
🐆

magicmatatjahu:
@P0lip Could you review and merge PR? Thanks!

yarn.lock
```diff
@@ -2444,6 +2444,25 @@ __metadata:
   languageName: node
```
P0lip: could you revert this `yarn.lock` change?

P0lip (commented):
Other than the yarn.lock, everything looks good to me

P0lip (approved):

yarn.lock
magicmatatjahu: @P0lip I reverted, but it was very strange, because without execution of `yarn` tests were failing. Let's see if it passes this time.

PR #2306, opened 2022-10-11
refactor(rulesets): remove unused rules style param

philsturgeon:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] Yes
- [ ] No

We have discussed removing the style parameter becasue it has not proved very useful compared to simply using Error and Warning, or splitting rules into multiple rulesets for different purposes or strictnesses. 

I had thought this would be a breaking change, but the more I look at it, the less I can see it being used in any way for anything. Perhaps we use it in Studio but as far as Spectral is concerned I am not sure this is breaking. Rulesets will still validate because JSON Schema doesn't care about extra properties.

**Screenshots**

**Additional context**

Add any other context about the pull request here. Remove this section if there is no additional context.

### DIFF
--- a/packages/core/src/ruleset/meta/rule.schema.json
+++ b/packages/core/src/ruleset/meta/rule.schema.json
@@ -72,6 +72,11 @@
         "else": {
           "$ref": "#/$defs/Then"
         }
+      },
+      "type": {
+        "enum": ["style", "validation"],
+        "type": "string",
+        "errorMessage": "allowed types are \"style\" and \"validation\""
       }
     },
     "required": ["given", "then"],
--- a/packages/core/src/ruleset/meta/rule.schema.json
+++ b/packages/core/src/ruleset/meta/rule.schema.json
@@ -72,11 +72,6 @@
         "else": {
           "$ref": "#/$defs/Then"
         }
-      },
-      "type": {
-        "enum": ["style", "validation"],
-        "type": "string",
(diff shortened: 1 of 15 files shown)


packages/core/src/ruleset/meta/rule.schema.json
```diff
@@ -72,11 +72,6 @@
         "else": {
           "$ref": "#/$defs/Then"
         }
-      },
-      "type": {
```
P0lip: I'd say keep it, otherwise it's a breaking change if there are folks that using it - the ruleset validation will probably mark it as a non-recognized (additional) property due to https://github.com/stoplightio/spectral/blob/813704d3f57a392a7083542b5df453d789a19e0b/packages/core/src/ruleset/meta/rule.schema.json#L78

P0lip (commented):
I'm okay with removing the type. I'd just keep it in the rule.schema.json to prevent a breaking change for users.

packages/core/src/ruleset/meta/rule.schema.json
philsturgeon: Good catch, thank you.

philsturgeon:
@P0lip I've deprecated it in the schema instead.

Any idea what is going on with these failing tests? Don't get that error locally.

P0lip:
I'll take a look at that test failure in a bit.

P0lip (approved):

PR #2326, opened 2022-11-02
fix(core): improve deep ruleset inheritance

### LINKED ISSUE #2211: Switching rules "off" doesn't work when using chain of inherited rulesets
If there is an hierarchy of rulesets (let’s say 5 inherited rulesets one from each other), then in the top ruleset in the `extends` field I switch off  all the rules, Spectral disables only rules from the direct parent, but not the rule of the other ancestors (parent-parent...).

**Example:**
```
const ruleset3 = {
    extends: [ruleset2],
    rules: { … }
}
```
```
const ruleset4 = {
    extends: [ruleset3],
    rules: { … }
}
```
```
const ruleset5 = {
   extends: [[ruleset4, “off”]],
    rules: { … }
}
```

It disables only rules from the `ruleset4` (direct parent), but not from the `ruleset3` and further ancestors.

**To Reproduce**

1. Checkout the repository - https://github.com/pavelkornev/spectral-example/
2. Run `npm start`
3. There should be no validation errors as they are all switched off, but there will be an error coming from the rule defined in `ruleset1`.
4. See https://github.com/pavelkornev/spectral-example/blob/master/src/index.ts#L14-L52

**Expected behavior**

When inheriting ruleset, it's expected that no matter how many ancestors it has, if I switch off all inherited rules, they must be switched off.

P0lip:
Fixes #2211 

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/src/ruleset/__tests__/__fixtures__/severity/off-proxy2.ts
+++ b/packages/core/src/ruleset/__tests__/__fixtures__/severity/off-proxy2.ts
@@ -1,0 +1,26 @@
+import { DiagnosticSeverity } from '@stoplight/types';
+import { truthy } from '@stoplight/spectral-functions';
+
+const ruleset1 = {
+  rules: {
+    'custom-info-description': {
+      message: 'API Description is missing',
+      severity: DiagnosticSeverity.Error,
+      given: '$.info',
+      then: {
+        field: 'description',
+        function: truthy,
+      },
+    },
+  },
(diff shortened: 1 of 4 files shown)


pavelkornev:
Can this be merged already?

mnaumanali94 (approved):

PR #2345, opened 2022-11-21
feat(core): support end-user extensions in the rule definitions

pavelkornev:
This change introduces a user extensions for rule definitions.

**Problem**
As a user of Spectral, we want to keep additional metadata close to the rule definitions. E.g. when this rule is introduced, who is responsible, etc.

**Solution**
As it's not possible to foresee what kind of metadata users might need, therefore introducing additional fields with a concrete specificity doesn't scale and will pollute the Rule interface.

The main idea behind this solution is to give Spectral users a way to define custom extensions similar to `x-` attributes in OpenAPI documents. In this change, the new property `extensions` is introduced in the rule type which can be used by users to define any user's payload inside.

Alternative idea for the property name: `metadata`.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional information**
The new field is not supported by the `print` utility. As the type of the extensions properties is unknown, it might be challenging to print it via `treeify`. Since this utility is only used for testing purposes, it seems to be okay to ignore the new property in this utility function. The similar situation is with `toJSON()`.


packages/core/src/ruleset/rule.ts
```diff
@@ -24,9 +24,10 @@ export interface IRule {
   documentationUrl: string | null;
   then: IRuleThen[];
   given: string[];
+  extensions: Record<string, unknown> | null;
 }
 
-type RuleJson = Omit<IRule, 'then'> & {
+type RuleJson = Omit<IRule, 'then' | 'extensions'> & {
```
P0lip:
```suggestion
type RuleJson = Omit<IRule, 'then'> & {
```
Let's have it included :)

P0lip (commented):
Looks good to me!
One minor request - could you please add `extensions` to the `toJSON` method? https://github.com/stoplightio/spectral/blob/362e74f5b371507a5e3124e6c74afc0061ca071f/packages/core/src/ruleset/rule.ts#L172

```diff
       resolved: this.resolved,
       formats: this.formats,
+      extensions: this.extensions,
       then: this.then.map(then => ({
 ```

packages/core/src/ruleset/rule.ts
pavelkornev: > Looks good to me! One minor request - could you please add `extensions` to
> 
> https://github.com/stoplightio/spectral/blob/362e74f5b371507a5e3124e6c74afc0061ca071f/packages/core/src/ruleset/rule.ts#L172
> 
> ?
> ```diff
>        resolved: this.resolved,
>        formats: this.formats,
> +      extensions: this.extensions,
>        then: this.then.map(then => ({
> ```

As extensions type is `Record<string, unknown>` we don't know what comes as a value here, it's `unknown`... it might be not a valid "JSON" in this sense.

Alternatively, we can allow only primitives as values in this object to make sure it's a JSON compatible object.

P0lip: > it's unknown... it might be not a valid "JSON" in this sense

Well, it's going to be JSON for JSON rulesets, but yeah, could be a non-serializable value in the case of JS rulesets.

> Alternatively, we can allow only primitives as values in this object to make sure it's a JSON compatible object.

objects and arrays are both JSON compatible objects too.
I believe we can leave it as is - JSON.stringify will just drop any invalid values.

pavelkornev: done

pavelkornev:
@P0lip should we adjust also the documentation? 
https://github.com/stoplightio/spectral/blob/develop/docs/guides/4a-rules.md

P0lip (approved):

P0lip:
> @P0lip should we adjust also the documentation?
> https://github.com/stoplightio/spectral/blob/develop/docs/guides/4a-rules.md

Up to you, I can do it later on. :)

pavelkornev:
@P0lip please go ahead and merge as i don't have the write access to the `develop` branch and cannot merge myself.

PR #2353, opened 2022-12-01
feat(core): scoring data customizable

PagoNxt-Trade:
**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [X] No

**Additional context**

This PR add a command line parameter to add a config file for results configuration which adds a scoring calculated and qualified with this results and a limit for passing


pablonxt:
From PagoNxt Trade, we are very excited to share this functionality with the community!

P0lip:
Hey!
Thanks a lot for the PR. I had time off recently, and now trying to catch up with everything. I'll make sure to have a pass on it this week.

packages/cli/src/commands/lint.ts
```diff
@@ -127,6 +135,11 @@ const lintCommand: CommandModule = {
           description: 'path/URL to a ruleset file',
           type: 'string',
         },
+        'scoring-config': {
+          alias: 's',
```
P0lip: tbh this option is rather quite specific, therefore I'm not sure whether an alias for it makes that much sense.
Would just keep it without an alias for now.

packages/cli/src/formatters/utils/getScoring.ts
```diff
@@ -0,0 +1,68 @@
...
+import fs from 'fs';
+
+export const getScoringConfig = (scoringFile?: string): ScoringConfig | undefined => {
+  if (scoringFile === void 0) {
+    return undefined;
+  } else if (!path.isAbsolute(scoringFile)) {
+    scoringFile = path.join(process.cwd(), scoringFile);
+  }
+
+  const scoringConfig: ScoringConfig = JSON.parse(fs.readFileSync(scoringFile, 'utf-8')) as ScoringConfig;
```
P0lip:
```suggestion
  const scoringConfig: ScoringConfig = JSON.parse(fs.promises.readFile(scoringFile, 'utf8'));
```

I'd also say we should impose some validation.
We could define a JSON Schema for the format and use Ajv for that purpose.
What do you think?

docs/guides/2-cli.md
```diff
@@ -60,6 +61,92 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
 - [OpenAPI ruleset](../reference/openapi-rules.md)
 - [AsyncAPI ruleset](../reference/asyncapi-rules.md)
 
+## Scoring the API
```
P0lip, line 64: @pamgoodrich do you want to have a quick look at the docs here?

packages/cli/src/formatters/utils/uniqueErrors.ts
```diff
@@ -0,0 +1,16 @@
...
+      !filteredResults.some(
+        (element: IRuleResult) => element.code === result.code && element.message === result.message,
+      )
+    ) {
+      filteredResults.push(result);
+    }
+  });
+
+  return filteredResults;
+};
```
P0lip, line 16:
```suggestion
function hash(result: IRuleResult): string {
  return `code:${result.code};message:${result.message}`
}

export const getUniqueErrors = (results: IRuleResult[]): IRuleResult[] => {
  const filteredResults = new Map<string, IRuleResult>();
  for (const result of results) {
    if (
      !filteredResults.some(
        (element: IRuleResult) => element.code === result.code && element.message === result.message,
      )
    ) {
      filteredResults.push(result);
    }
  });

  return filteredResults;
};
```

packages/cli/src/formatters/json.ts
```diff
@@ -1,6 +1,21 @@
...
+
+const version = process.env.npm_package_version;
+
+export const json: Formatter = (results: ISpectralDiagnostic[], options: FormatterOptions) => {
+  let spectralVersion = '';
+  let groupedResults;
+  let scoringText = '';
+  if (options.scoringConfig !== void 0) {
+    if (options.scoringConfig.customScoring !== undefined) {
+      spectralVersion = `${options.scoringConfig.customScoring} ${version as string}`;
```
P0lip: I don't think including a version makes much sense.
The result does not directly depend on the version of the CLI client, but on the ruleset you use, so seems like we could just not include it?

packages/cli/src/formatters/pretty.ts
```diff
@@ -24,22 +24,40 @@
...
   return ` ${range.start.line + 1}:${range.start.character + 1}`;
 }
 
-export const pretty: Formatter = results => {
+export const pretty: Formatter = (results: ISpectralDiagnostic[], options: FormatterOptions) => {
   const cliui = require('cliui');
   let output = '\n';
+  if (options.scoringConfig !== void 0) {
+    if (options.scoringConfig.customScoring !== void 0) {
+      output += `${options.scoringConfig.customScoring}${version as string}\n`;
```
P0lip: Ditto ^

packages/cli/src/commands/lint.ts
```diff
@@ -197,20 +211,30 @@ const lintCommand: CommandModule = {
...
+        process.exit(
+          scoringThresholdNotEnough(results, scoringConfigData) ? 1 : severeEnoughToFail(results, failSeverity) ? 1 : 0,
+        );
       } else if (config.quiet !== true) {
         const isErrorSeverity = getDiagnosticSeverity(failSeverity) === DiagnosticSeverity.Error;
         process.stdout.write(
           `No results with a severity of '${failSeverity}' ${isErrorSeverity ? '' : 'or higher '}found!\n`,
         );
+        if (scoringConfig !== void 0) {
+          process.stdout.write(`SCORING: (100%)\nPASSED!`);
```
P0lip, line 235: maybe we just skip it if no scoring was provided?

docs/guides/2-cli.md
```diff
@@ -60,6 +61,92 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+    }
+```
+
+Where:
+
+- scoringSubtract : An object with a key/value pair objects for every result level we want to subtract percentage, with the percentage to subtract from number of results on every result type
+- scoringLetter : An object with key/value pairs with scoring letter and scoring percentage, that the result must be greater , for this letter
+- threshold : A number with minimum percentage value to provide valid the file we are checking
+- warningsSubtract : A boolean to setup if accumulate the result types to less the scoring percentage or stop counting on most critical result types
+- uniqueErrors : A boolean to setup a count with unique errors or with all of them
```
P0lip: We should define what a unique error would be.
Something like "An error is considered unique if its code and message have not been seen yet" or something similar.

packages/cli/src/formatters/utils/uniqueErrors.ts
```diff
@@ -0,0 +1,16 @@
+import { IRuleResult } from '@stoplight/spectral-core';
+
+export const uniqueErrors = (results: IRuleResult[]): IRuleResult[] => {
```
P0lip:
```suggestion
export const getUniqueErrors = (results: IRuleResult[]): IRuleResult[] => {
```

packages/cli/src/formatters/pretty.ts
```diff
@@ -83,6 +114,15 @@ export const pretty: Formatter = results => {
   output += ui.toString();
   output += chalk[summaryColor].bold(`${uniqueIssues.length} Unique Issue(s)\n`);
   output += chalk[summaryColor].bold(`\u2716${summaryText !== null ? ` ${summaryText}` : ''}\n`);
+  if (options.scoringConfig !== void 0) {
+    output += chalk[scoringColor].bold(`\u2716${scoringText !== null ? ` ${scoringText}` : ''}\n`);
+    const scoring = +(scoringText !== null ? scoringText.replace('%', '').split(/[()]+/)[1] : 0);
+    if (scoring >= options.scoringConfig.threshold) {
+      output += chalk['green'].bold(`\u2716 PASSED!\n`);
+    } else {
+      output += chalk['red'].bold(`\u2716 NOT PASSED!\n`);
```
P0lip:
```suggestion
      output += chalk['red'].bold(`\u2716 FAILED!\n`);
```

packages/cli/src/formatters/stylish.ts
```diff
@@ -92,6 +124,15 @@ export const stylish: Formatter = results => {
   }
 
   output += chalk[summaryColor].bold(`\u2716 ${summaryText}\n`);
+  if (options.scoringConfig !== void 0) {
+    output += chalk[scoringColor].bold(`\u2716${scoringText !== null ? ` ${scoringText}` : ''}\n`);
+    const scoring = +(scoringText !== null ? scoringText.replace('%', '').split(/[()]+/)[1] : 0);
+    if (scoring >= options.scoringConfig.threshold) {
+      output += chalk['green'].bold(`\u2716 PASSED!\n`);
+    } else {
+      output += chalk['red'].bold(`\u2716 NOT PASSED!\n`);
```
P0lip:
```suggestion
      output += chalk['red'].bold(`\u2716 FAILED!\n`);
```

P0lip (commented):
Looking pretty neat!

What do you think @mnaumanali94?

packages/cli/src/commands/lint.ts
PagoNxt-Trade: This is going to happen when 'results.length > 0' and we have scoring, then, the scoring is '100%'
Or I'm not understanding your comment

packages/cli/src/formatters/utils/getScoring.ts
PagoNxt-Trade: We've got your suggestion
For the validation ... We guess that could be a duplicated validation, with type defined on _cli/src/formatters/types.ts_ should be enough to validate input file, but, if anybody thinks like you about this, we could add this

packages/cli/src/commands/lint.ts
P0lip: ah yeah, I saw it wrong. Yup, you're right. No action needed here

kevinswiber:
@PagoNxt-Trade This is interesting! I have some concerns about locking in a scoring mechanism, as flexibility to model different use cases and opinions is a significant component to retrieving value out of the system.

It might be helpful to get the perspective of some Apigee folks who are also working on a scoring approach via https://github.com/apigee/registry.  I believe, in their case, they're allowing folks to provide their own custom [CEL](https://github.com/google/cel-spec) formulas to determine the score.

Any input @timburks, @shrutiparabgoogle, @theganyo?

pablonxt:
Hi,

I am the API Development Product Owner at PagoNxt Trade, developers are mostly on holidays till 9th Jan.

@kevinswiber you approach sounds good, but we would need to research on it. Anyhow I like to say that the scoring is somehow flexible as people can add their scoring configuration adapted to their needs. It mostly subtracting something when you find errors. At least is a first approach to scoring.

Somethings we would like to add in a near future are this:

**More severity levels**. Possibility to have more severity levels like **critical**, **warning+**, **warning-**. Adding **critical** at least, would be useful.

**scoring and rules by category**. Adding a new field category to the rules. This will enable to do scoring by category and a general scoring. Example of categories: Definition, Documentation, Security

Seeing the scoring in Stoplight Studio would be also a must.

pablonxt:
Hi @P0lip can you review the latest changes and if you agree resolve the discussions? Most of them are addressed . Thanks

P0lip (commented):
The code looks pretty good to me.
I know @mnaumanali94 wanted to take a look at the functionality itself, so let's wait for his take.

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
 - [OpenAPI ruleset](../reference/openapi-rules.md)
 - [AsyncAPI ruleset](../reference/asyncapi-rules.md)
 
+## Scoring the API
+
+Scoring an API definition is a way to understand in a high level, how compliant is the API definition with the rulesets provided. This helps teams to understand the quality of the APIs regarding the definition.
```
pamgoodrich: Scoring an API definition is a way to understand in a high level, how compliant is the API definition with the rulesets provided.

Suggest

Scoring an API definition is a way to understand at a high level how compliant the API definition is with the rulesets provided.

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+
+The scoring is produced in two different metrics:
+
+- A number scoring. Who cames as substracting from 100% from any error or warning
```
pamgoodrich: Suggest: 

A number scoring: Calculated by subtracting any error or warning from 100%.

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+- A letter, who groups numeric scorings in letters from A (better) to any
```
pamgoodrich: Suggest:

A letter scoring, which groups numeric scoring in letters from A to Z, with A being the best score.

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+        },
+      "threshold":50,
+      "warningsSubtract": true,
+      "uniqueErrors": false
+    }
+```
+
+Where:
+
+- scoringSubtract : An object with a key/value pair objects for every result level we want to subtract percentage, with the percentage to subtract from number of results on every result type
```
pamgoodrich: ...a key/value pair objects..." 

I think you need to remove "a" or make "objects" singular.

So:

An object with key/value pair objects for...

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+- scoringLetter : An object with key/value pairs with scoring letter and scoring percentage, that the result must be greater , for this letter
```
pamgoodrich: Remove space after greater and before the comma

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+- threshold : A number with minimum percentage value to provide valid the file we are checking
+- warningsSubtract : A boolean to setup if accumulate the result types to less the scoring percentage or stop counting on most critical result types
```
pamgoodrich: Suggest:

- warningsSubtract : A boolean to accumulate the result types to less than the scoring percentage or to stop counting on most critical result types

docs/guides/2-cli.md
```diff
@@ -60,6 +61,86 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+- uniqueErrors : A boolean to setup a count with unique errors or with all of them. An error is considered unique if its code and message have not been seen yet
```
pamgoodrich: Suggest:

A boolean to count unique errors or all errors.

pamgoodrich (commented):
Thank you for writing the documentation for this feature. I made a few suggestions for clarity and grammatical correctness.

pablonxt:
@pamgoodrich all your suggestions have been applied. Thanks!

shrutiparabgoogle:
> @PagoNxt-Trade This is interesting! I have some concerns about locking in a scoring mechanism, as flexibility to model different use cases and opinions is a significant component to retrieving value out of the system.
> 
> It might be helpful to get the perspective of some Apigee folks who are also working on a scoring approach via https://github.com/apigee/registry. I believe, in their case, they're allowing folks to provide their own custom [CEL](https://github.com/google/cel-spec) formulas to determine the score.
> 
> Any input @timburks, @shrutiparabgoogle, @theganyo?

This is interesting! Thanks for the mention.

I can see that the scoring approach mentioned here is making some assumptions:
- that the score is always a percentage
- the formula used to calculate the score is also opinionated (subtract from 100)

The approach that we have implemented in https://github.com/apigee/registry is flexible with the goal to accommodate scores for anything and not only lint results. Hence, we made use of CEL expressions so that users can define their own formula for calculating the score. 

While that kind of flexibility was one of the goals for us, I can see that it can be an overkill for lint specific scoring. Couple of thoughts on the current approach:
- Will scores always be percentages or is there a need to support a more flexible range? It might also help to add some validation while accepting a scoring config, so that users don't accidentally generate invalid scores (like negative percentages -2%?). I might have missed it if it’s already in place.
- What is the purpose of letter scoring? (trying to understand the use case here). IIUC, it is mapping the generated percentage to a letter score, what difference does looking at a letter provide me vs looking at a percentage? Perhaps these are two different ways of conveying the same information and there can be flexibility in whether users want a lettered score or a percentage score?
- The use of threshold and warningsSubtract flags is not very clear to me from the documentation, explicitly mentioning the formula in the docs might help gain clarity in the assumptions made.
- If the user doesn’t provide a scoringConfig, is there a default config they can get started with?

pablonxt:
Hi @shrutiparabgoogle 

Thanks for your input. Here you have our answers.

Usually scoring is done starting with 100% and substracting a value depending on the issue severity (error, warning,...) this is a common practice for scoring. We think is a complete overkill to go for CELs

What can be a more flexible range? We see this as previous sentence (a percentage)

About the possible negative, we will run some tests to validate that it can't happen

The purpose of letter scoring is to show a nicer scoring than a raw number. Letter scoring comes as anything you like A+, A, B, C ... See qualys ssl tests results here: https://www.ssllabs.com/ssltest/

The number or percentage is required by the algorithm to be able to substract values

We can make letter scoring optional if you don't provide the configuration of it. So results will be always percentaje and optionally letter if you provide the letter scoring configuration.

>The use of threshold and warningsSubtract flags is not very clear to me from the documentation, explicitly mentioning the formula in the docs might help gain clarity in the assumptions made.

threshold : A number with minimum percentage value to provide valid the file we are checking. Any scoring below this thresold will mark the API as a failure in the scoring.

warningsSubtract, we are going to change its name to onlySubtractHigherSeverityLevel and adaprt the documentation for more clarity.

onlySubtractHigherSeverityLevel : A boolean to decide if only the higher severity level who appears in the results for the API to analize, are subtracted from scoring or every severity level are subtracted from scoring.

See sample:

true

API with Errors and Warnings, only Errors substract from scoring
API with Warnings, Warnings substract from scoring

false

API with Errors and Warnings, Errors and Warnings substracts from scoring
API with Warnings, Warnings substract from scoring


>If the user doesn’t provide a scoringConfig, is there a default config they can get started with?

If you don't provide scoringConfig there will be no scoring. As spectral does now.

Does this work better?

pablonxt:
Hi .. do you have any update? @P0lip @pamgoodrich @mnaumanali94

P0lip:
@mnaumanali94 ^

docs/guides/2-cli.md
```diff
@@ -60,6 +61,99 @@ Here you can build a [custom ruleset](../getting-started/3-rulesets.md), or exte
...
+      "uniqueErrors": false
+    }
+```
+
+Where:
+
+- scoringSubtract : An object with key/value pair objects for every result level we want to subtract percentage, with the percentage to subtract from number of results on every result type
+- scoringLetter : An object with key/value pairs with scoring letter and scoring percentage, that the result must be greater, for this letter
+- threshold : A number with minimum percentage value to provide valid the file we are checking. Any scoring below this thresold will mark the API as a failure in the scoring.
+- onlySubtractHigherSeverityLevel : A boolean to decide if only the higher severity level who appears in the results for the API to analize, are subtracted from scoring or every severity level are subtracted from scoring.
```
jharmn, line 125: analize-> analyze

pablonxt:
@mnaumanali94 do you have any update? We need this PR merged

kopernic-pl:
I see that this PR has gained quite an interest. @PagoNxt-Trade would you be able to bring it up to date with current code base state?

PR #2355, opened 2022-12-01
fix(rulesets): length.min said "must not be longer than"

philsturgeon:
When the rule is minimum 20, if there is less than 20 characters it will say "must not be longer than 20". This removes the not, making minimum now correctly say "must be longer than 20".

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [x] Yes
- [ ] No

Any change to a message could be considered breaking but this was broken already, so hopefully people arent programatically doing things on string responses.

### DIFF
--- a/packages/functions/src/__tests__/length.test.ts
+++ b/packages/functions/src/__tests__/length.test.ts
@@ -35,7 +35,7 @@ describe('Core Functions / Length', () => {
     async input => {
       expect(await runLength(input, { min: 4 })).toEqual([
         {
-          message: 'The document must not be longer than 4',
+          message: 'The document must be longer than 4',
           path: [],
         },
       ]);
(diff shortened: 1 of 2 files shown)


dotslashderek (dismissed):

P0lip (changes requested):
In such a case, I suppose we should update all messages.
They all use `must` keyword

philsturgeon:
@P0lip I don't think we should block this on the grounds of consistency. Currently its literally incorrect.

Note that the rule is `if ('min' in opts && value < opts.min) {`

"must not be longer than"

If minimum is 5 characters, and you provide 4 characters, it will say: 

"foo must not be longer than 5 characters"

whiiiich is the opposite of what it means. It means

"foo must be longer than 5 characters"

Can we merge to fix this?

philsturgeon:
@P0lip @mnaumanali94 this bug is holding up the Documentation ruleset, as right now it's saying "must not be longer than 20 characters" for a shorter than 20 characters description. I'll code the tests with the bug in place but somebody else will need to update that later.

philsturgeon:
@P0lip redone with minimal change to fix the bug. Can I get a thumb?

P0lip (approved):

PR #2369, opened 2022-12-21
docs(repo): added info about inherited overrides

### LINKED ISSUE #2331: "overrides" section from included ruleset is ignored
**Describe the bug**
I have a master ruleset file which includes other files via "extends".
I need to define some "overrides" rules and I want to place them in a separate file and add it via "extens".
So, when I define the "overrides" section in external file, it is not applied.
If I do the same in master file, it works as expected.

**To Reproduce**

1. Create two ruleset files where ruleset-1 includes ruleset-2 via "extends"
2. Define "overrides" section in ruleset-2

**Actual behavior**
Overrides from included file are not applied

**Expected behavior**
Overrides from included file should be applied

**Environment:**
 - spectral-cli 6.6.0

pamgoodrich:
Fixes #2331.

Added notes and examples for inherited overrides.

### DIFF
--- a/docs/guides/4d-overrides.md
+++ b/docs/guides/4d-overrides.md
@@ -128,7 +128,7 @@ No results with a severity of 'error' or higher found!
 
 #### Extended Rulesets (extends)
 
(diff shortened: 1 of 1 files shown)


P0lip (dismissed):
Thank you!

P0lip (dismissed):

PR #2403, opened 2023-02-18
fix(ruleset-bundler): defaults should be last one

dweber019:
This fixes the Webpack error
```
Module not found: Default condition should be last one
```
when importing like
```
import { bundleAndLoadRuleset } from '@stoplight/spectral-ruleset-bundler/with-loader';
```

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No


dweber019:
FYI @P0lip

P0lip (approved):

PR #2405, opened 2023-02-19
fix(ruleset-migrator): use module for require.resolve

### LINKED ISSUE #2404: Warnings critical dependency with webpack, spectral-ruleset-migrator
**Describe the bug**
I receive the following error when using Spectral like this
https://github.com/dweber019/backstage/blob/feat/api-spectral-linter/plugins/api-docs-spectral-linter/src/api/LinterClient.ts#L21

```
[0] WARNING in ../../node_modules/@stoplight/spectral-ruleset-migrator/dist/requireResolve.js 5:15-22
[0] Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
[0] 
[0] webpack compiled with 1 warning
```

I think it could be a similar issue like https://github.com/stoplightio/spectral/issues/1431

**To Reproduce**

1. Clone https://github.com/dweber019/backstage/tree/feat/api-spectral-linter
2. yarn install
3. yarn dev

**Expected behavior**
No warning

OS:   Darwin 22.3.0 - darwin/x64
node: v16.14.2
yarn: 3.2.3
Webpack: 5.74.0
Webpack config: https://github.com/dweber019/backstage/blob/feat/api-spectral-linter/packages/cli/src/lib/bundler/config.ts

dweber019:
Will fix https://github.com/stoplightio/spectral/issues/2404

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

- [ ] Yes
- [X] No

### DIFF
--- a/packages/ruleset-migrator/src/requireResolve.ts
+++ b/packages/ruleset-migrator/src/requireResolve.ts
@@ -1,6 +1,9 @@
+import { createRequire } from 'module';
+
 export default <NodeRequire['resolve'] | null>((id, opts) => {
   try {
-    return require.resolve(id, opts);
+    const req = createRequire(process.cwd());
+    return req.resolve(id, opts);
   } catch {
     return null;
   }


dweber019:
FYI @P0lip

P0lip:
Thanks for the PR!
could you please re-format the commit message so that the lint step succeeds?

packages/ruleset-migrator/src/requireResolve.ts
```diff
@@ -1,6 +1,6 @@
 export default <NodeRequire['resolve'] | null>((id, opts) => {
   try {
-    return require.resolve(id, opts);
+    return eval('require.resolve')(id, opts);
```
P0lip: could we use `module` in the same as we do it [here](https://github.com/stoplightio/spectral/blob/97134a5fb37b8a1c9cb00cccbcf68fcb7bd9a959/packages/ruleset-bundler/src/loader/node.ts#L1) and [here](https://github.com/stoplightio/spectral/blob/97134a5fb37b8a1c9cb00cccbcf68fcb7bd9a959/packages/cli/src/services/linter/utils/getRuleset.ts#L6).

You can provide `process.cwd()` as the base uri.

dweber019: I hope I did it correctly 😅

P0lip (approved):

packages/ruleset-migrator/src/requireResolve.ts
P0lip: looks good!
We have some tests that cover it so if something was off, they would fail.

P0lip:
Thank you!

PR #2460, opened 2023-04-21
feat(rulesets): add message traits array path to asyncapi headers schema type object rule

CalemRoelofsSB:
Recreation of #2402 to fix issues around commitlint

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

Currently, the `asyncapi-headers-schema-type-object` will not trigger when the `headers` object is defined explicitly in the `message.traits` array. 
For example:
```yaml
...
channels:
  'users/{userId}/loggedIn':
    publish:
      message:
        traits:
          - headers
              type: integer
```

The new path added in this PR will match on that array and perform the assertion on the `type` field of any object in that array that has a top level key of `headers`.

### DIFF
--- a/packages/rulesets/src/asyncapi/__tests__/asyncapi-headers-schema-type-object.test.ts
+++ b/packages/rulesets/src/asyncapi/__tests__/asyncapi-headers-schema-type-object.test.ts
@@ -26,6 +26,18 @@ const document = {
         message: cloneDeep(headersBearer),
       },
     },
+    'users/{userId}/loggedIn': {
+      publish: {
+        message: {
+          traits: [cloneDeep(headersBearer)],
+        },
+      },
+      subscribe: {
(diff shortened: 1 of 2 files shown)


CalemRoelofsSB:
ping @P0lip

P0lip (approved):

P0lip:
Looks good to me, although I'd wait for an approval from our AsyncAPI colleagues as they're the most familiar with it :)

jonaslagoni (approved):
LGTM 👍

CalemRoelofsSB:
@P0lip @jonaslagoni thank you for the review folks! Unfortunately, I don't have permissions to merge this branch into `develop` so if one of you could oblige, that would be most appreciated :)

PR #2475, opened 2023-05-17
fix(ruleset-bundler): remove extraneous 'external dependency' warnings

billiegoose:
Fixes https://github.com/stoplightio/platform-internal/issues/16525

**Checklist**

- [ ] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**

Before:
239012037-c8539711-0ebd-4293-baa6-04f604a4f3d5

After:
239021842-ef4fcd3b-f6db-4c56-af6a-5fbd4e408c04

**Additional context**

These warnings create noise in our DataDog logs for the `api-design` service.

### DIFF
--- a/packages/ruleset-bundler/src/index.ts
+++ b/packages/ruleset-bundler/src/index.ts
@@ -35,6 +35,14 @@ export async function bundleRuleset(
       if (e.code === 'MISSING_NAME_OPTION_FOR_IIFE_EXPORT') {
         return;
       }
+      // The Spectral packages themselves are not included in the bundle.
+      if (
+        e.code === 'UNRESOLVED_IMPORT' &&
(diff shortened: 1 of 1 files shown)


P0lip (approved):
I think it's fine for now.
Ideally, we should perhaps allow custom external . This way we could also get rid of the option we pass here 
image

FWIW, that warning might be actually valuable for some less advanced users of Spectral if they use Spectral programmatically and choose a wrong bundle target, etc.

The reason we can safely keep `@stoplight/spectral-` imports in our platform is that that ruleset is meant to be used either in Studio or CLI, so the bundling will happen again.
Otherwise, if you, say, have a web app and set node as a bundle target, the imports will be kept, but loading that code will fail.

PR #2508, opened 2023-07-13
feat(formatters): add GitHub Actions formatter

siketyan:
This pull requests add a built-in formatter for annotating errors in GitHub Actions.

Spectral already have spectral-action for running Spectral and annotating errors in the workflow. The action creates a new check entry to annotate, but sometimes the annotation is incorrectly attached to another job due to GitHub's bug. To avoid that, it would be great if the Spectral CLI support formatting errors to GitHub Actions format [^1] as a built-in formatter, as same as TeamCity formatter does.

[^1]: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**

N/A

**Additional context**

N/A

### DIFF
--- a/packages/formatters/src/github-actions.ts
+++ b/packages/formatters/src/github-actions.ts
@@ -36,7 +36,7 @@ export const githubActions: Formatter = results => {
         .map(p => p.join('='))
         .join(',');
 
-      return `::${OUTPUT_TYPES[result.severity]} ${paramsString}::${result.message}`;
+      return `::${OUTPUT_TYPES[result.severity]} ${paramsString}::${result.message.replaceAll('\n', '%0A')}`;
     })
     .join('\n');
 };
--- a/packages/formatters/src/github-actions.ts
+++ b/packages/formatters/src/github-actions.ts
@@ -36,7 +36,12 @@ export const githubActions: Formatter = results => {
         .map(p => p.join('='))
         .join(',');
 
-      return `::${OUTPUT_TYPES[result.severity]} ${paramsString}::${result.message.replaceAll('\n', '%0A')}`;
+      // As annotated messages must be one-line due to GitHub's limitation, replacing all LF to %0A here.
+      // see: https://github.com/actions/toolkit/issues/193
+      // FIXME: Use replaceAll instead after removing Node.js 14 support.
+      const message = result.message.replace(/\n/g, '%0A');
+
+      return `::${OUTPUT_TYPES[result.severity]} ${paramsString}::${message}`;
     })
     .join('\n');
 };
--- a/packages/formatters/src/github-actions.ts
+++ b/packages/formatters/src/github-actions.ts
@@ -18,11 +18,13 @@ type OutputParams = {
 };
 
 export const githubActions: Formatter = results => {
+  const path = require('path') as { relative: (from: string, to: string) => string };
+
   return results
     .map(result => {
       const params: OutputParams = {
         title: result.code.toString(),
-        file: require('path').relative(process.cwd(), result.source ?? ''),
+        file: path.relative(process.cwd(), result.source ?? ''),
         col: result.range.start.character,
         endColumn: result.range.end.character,
         line: result.range.start.line,
--- a/packages/formatters/src/github-actions.ts
+++ b/packages/formatters/src/github-actions.ts
@@ -1,3 +1,4 @@
+import { relative } from '@stoplight/path';
 import { DiagnosticSeverity, Dictionary } from '@stoplight/types';
 import { Formatter } from './types';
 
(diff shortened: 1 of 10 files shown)


packages/formatters/src/github-actions.ts
```diff
@@ -0,0 +1,41 @@
...
+  title?: string;
+  file: string;
+  col?: number;
+  endColumn?: number;
+  line?: number;
+  endLine?: number;
+};
+
+export const githubActions: Formatter = results => {
+  const path = require('path') as { relative: (from: string, to: string) => string };
```
P0lip:
```ts
import * as path form '@stoplight/path
```
could we use `@stoplight/path`?

packages/formatters/src/index.ts
```diff
@@ -4,6 +4,7 @@ export * from './junit';
 export * from './html';
 export * from './text';
 export * from './teamcity';
+export * from './github-actions';
```
P0lip: I don't think we can export it from this file. We use `process.cwd` which is specific to Node.js.
We may need to do something similar to what we did with `pretty` (at the very bottom of the file).
Could we also update README?

packages/formatters/src/github-actions.ts
siketyan: Fixed in [a18d591](https://github.com/stoplightio/spectral/pull/2508/commits/a18d5917afa10aa5f95110384cd6b413d8168c93).

packages/formatters/src/index.ts
siketyan: Fixed in [d79e4b9](https://github.com/stoplightio/spectral/pull/2508/commits/d79e4b986353bb0092644ec6eb10b5c471e21f51) and [00b5dfc](https://github.com/stoplightio/spectral/pull/2508/commits/00b5dfc8f04d4b0c398baf8856d8debeb93d6c47).

P0lip:
I'll try to take a look at it this week!

packages/formatters/src/github-actions.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const githubActions: Formatter = results => {
+  return results
+    .map(result => {
+      // GitHub Actions requires relative path for annotations, determining from working directory here
+      const file = relative(process.cwd(), result.source ?? '');
+      const params: OutputParams = {
+        title: result.code.toString(),
+        file,
+        col: result.range.start.character,
```
P0lip: doesn't Github's `endColumn` start at 1?
`IDiagnostic` that Spectral implements follows LSP where [positions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position) are 0-based, both the character and line.

```suggestion
        col: result.range.start.character + 1,
```

packages/formatters/src/github-actions.ts
```diff
@@ -0,0 +1,42 @@
...
+        endColumn: result.range.end.character,
```
P0lip: ditto ^

```suggestion
        endColumn: result.range.end.character + 1,
```

packages/formatters/src/github-actions.ts
```diff
@@ -0,0 +1,42 @@
...
+        line: result.range.start.line,
+        endLine: result.range.end.line,
+      };
+
+      const paramsString = Object.entries(params)
+        .map(p => p.join('='))
+        .join(',');
+
+      return `::${OUTPUT_TYPES[result.severity]} ${paramsString}::${result.message}`;
```
P0lip: Do new lines in `result.message` need to be escaped somehow? At times, `result.message` may contain new lines if people define it that way in their ruleset

P0lip (commented):
Other than a potential column mismatch, this looks great.
Thanks a lot for the contribution!

packages/formatters/src/github-actions.ts
siketyan: Nice catch, both `line` and `col` start at 1 in GitHub Actions [^1].
Fixed in [77cddc0](https://github.com/stoplightio/spectral/pull/2508/commits/77cddc005d1b47f58e88cd9371c01f78665a1283).

[^1]: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message

packages/formatters/src/github-actions.ts
siketyan: @P0lip Yes, we need to escape newlines. As described in https://github.com/actions/toolkit/issues/193, we can use `%0A` URL-encoded character here.
Fixed in [b829876](https://github.com/stoplightio/spectral/pull/2508/commits/b8298760dcc74c8684a7823731a2448a44f9f34d).

P0lip (approved):
LGTM. Thanks again!
I'll merge and release it today

P0lip:
Could you merge the latest develop branch? it's out of sync with the base branch

siketyan:
@P0lip merged, sorry for the delay; I was in the dream 💤

PR #2526, opened 2023-08-21
feat(rulesets): add oas3-server-variables rule

### LINKED ISSUE #2524: Support for OAS Server Variables
### **Requirements**

- [ ] Implement a Spectral rule to ensure URLs are valid format even when variables are used

- [ ] Implement a Spectral rule to check that if allowed values (enums) are defined for server variables, the default value should be a part of those allowed values.
    - For instance, if allowed values for `version` are `v1`, `v2`, and `v3`, and the default value is `v4`, then this should throw an error.

- [ ] Implement a rule to ensure that all server variables used in the server URL are defined in the server variables section.
    - For example, if `version` is used in a server URL like `https://api.example.com/{version}`, ensure that `version` is defined as a server variable.

- [ ] Implement a rule to ensure all defined server variables are used in the server URL.
    - If a server variable like `stage` is defined but not used in any server URL, this should trigger a warning.

P0lip:
Closes #2524

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


packages/rulesets/src/oas/index.ts
```diff
@@ -671,5 +692,18 @@ const ruleset = {
         function: oasUnusedComponent,
       },
     },
+    'oas3-server-variables': {
+      description: 'Server variables must be defined and valid and there must be no redundant variables.',
```
mnaumanali94:
```suggestion
      description: 'Server variables must be defined and valid and there must be no unused variables.',
```

packages/rulesets/src/shared/functions/serverVariables/index.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    const results: IFunctionResult[] = [];
+
+    const foundVariables = parseUrlVariables(url);
+    const definedVariablesKeys = Object.keys(variables);
+
+    const redundantVariables = getRedundantProps(foundVariables, definedVariablesKeys);
+    for (const variable of redundantVariables) {
+      results.push({
+        message: `Server's "variables" object has redundant defined "${variable}" url variable.`,
```
mnaumanali94:
```suggestion
        message: `Server's "variables" object has unused defined "${variable}" url variable.`,
```

mnaumanali94 (dismissed):

docs/reference/openapi-rules.md
```diff
@@ -846,3 +846,79 @@ schemas:
...
+
+**Recommended**: Yes
+
+**Bad Examples**
+
+1. **Missing definition for a URL variable**:
+
+```yaml
+servers:
+  - url: "<https://api>.{region}.example.com/v1"
```
billiegoose: Why is there `< >` around `<https://api>`? Is that a typo or just some Markdown / YAML think I don't understand

```suggestion
  - url: "https://api.{region}.example.com/v1"
```

billiegoose (changes requested):
Aside from the one (typo?) in the docs, it looks pretty solid to me.

docs/reference/openapi-rules.md
P0lip: @mnaumanali94?

billiegoose (approved):

PR #2532, opened 2023-09-05
feat(formatters): add sarif formatter

PhilippHeuer:
Adds a formatter to output results in the Static Analysis Results Interchange Format (SARIF) Version 2.1.0.

https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

- It would be nice to have all rules listed in the formatter instead of just rules that have been triggered.


P0lip:
Hey! Thanks for the PR, I'll try to review it soon. I had a peek at it and everything looks right at first glance, but I'd like to have a deeper look later on.

PhilippHeuer:
Alright. I will take another look to see if i can figure out how to pass the ruleset to the formatter, as the sarif rules should normally include all rules.

P0lip:
Getting a list of all rules might be a bit tricky due to [overrides](https://github.com/stoplightio/spectral/blob/develop/docs/guides/4d-overrides.md). Technically you can apply a ruleset that's applicable only to a given set of files that match a provided glob pattern.

I haven't read the spec yet, but do we need to provide a set of rules on a per-file or per-project basis?

Either way, you could start here https://github.com/stoplightio/spectral/blob/429572868edd430e28fc5c2aaa375e63aefb492b/packages/cli/src/services/linter/linter.ts#L15 and expose ruleset next to results. You could then pass it along to `formatOutput` function.

If one needs a per-file basis ruleset (with overrides applied), we'd need to expose that ruleset in the core package here https://github.com/stoplightio/spectral/blob/develop/packages/core/src/spectral.ts#L73 and then use `runWithResolved` here https://github.com/stoplightio/spectral/blob/429572868edd430e28fc5c2aaa375e63aefb492b/packages/cli/src/services/linter/linter.ts#L44 to obtain it.

PhilippHeuer:
Thanks for the links, i added the ruleset to the response of the `lint` function and passed it to `formatOutput` as you suggested.

The `rules` section contains information about all supported analysis rules and is provided once per-run, not per-file. In our case this would be the provided ruleset.

If a rule has a override to restrict it to specific files and gets skipped due to that, i still think that it would be correct to document it as a rule that spectral lint has been run with. Therefore i would include all rules from the resolved ruleset into the sarif report.

P0lip (dismissed):
LGTM. I'll release it this week

P0lip (dismissed):

P0lip (approved):

PR #2533, opened 2023-09-13
fix(rulesets): tweak server variables function

P0lip:
Other than the removal of abrupt return when `variables` was not defined, we now also demand `default` to be provided in the case of OpenAPI Server Variable objects.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No


billiegoose (approved):
👍🏻

PR #2538, opened 2023-09-19
feat(rulesets): add scope validation to oas{2,3}-operation-security-defined rules

### LINKED ISSUE #2529: Make sure op-security-defined rules work correctly with AND auth
**Chore summary**
- [ ] We have rules for oas3/oas2 op-security-defined. Make sure these work correctly for AND Auth usecases/syntax. If it doesn't update them to make sure it works. 
- [ ] Within the oas3/oas2 op-security-defined rules: Validate if scopes specified within a security requirements object are defined within security schemes

P0lip:
Closes #2529 

TBH we should probably rename that rule since it also validates top-level securities.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/packages/rulesets/src/oas/__tests__/oas2-operation-security-defined.test.ts
+++ b/packages/rulesets/src/oas/__tests__/oas2-operation-security-defined.test.ts
@@ -3,12 +3,21 @@ import testRule from './__helpers__/tester';
 
 testRule('oas2-operation-security-defined', [
   {
-    name: 'a correct object (just in body)',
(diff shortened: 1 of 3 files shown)


mnaumanali94:
Do we consider renaming a breaking change? 🤔

mnaumanali94 (approved):

P0lip:
> Do we consider renaming a breaking change? 🤔

That's what we used to do in the past.

PR #2542, opened 2023-10-06
fix(rulesets): make rule `oas2-operation-formData-consume-check` consistent with Swagger Editor similar rule

pavelkornev:
Swagger Editor has a similar rule to check values of `consumes` array:
https://github.com/swagger-api/swagger-editor/blob/249d4f0044d2104fdbe582cf53f27f3538b361d7/src/plugins/validate-semantic/validators/form-data.js#L50-L96

The difference is that Spectral rule does not fail if `consumes` array is completely missing which seems to be a mistake.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [x] Yes
- [ ] No

Potentially, it might be a breaking change for those who use this rule as it will though an additional validation issue in this edge case.

### DIFF
--- a/packages/rulesets/src/oas/__tests__/oas2-operation-formData-consume-check.test.ts
+++ b/packages/rulesets/src/oas/__tests__/oas2-operation-formData-consume-check.test.ts
@@ -40,4 +40,26 @@ testRule('oas2-operation-formData-consume-check', [
       },
     ],
   },
+
+  {
+    name: 'return error when consumes is missing',
+    document: {
+      swagger: '2.0',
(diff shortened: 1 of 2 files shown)


pavelkornev:
@P0lip, where can i find the failing test file?
`test-harness/tests/examples.oas2/examples.oas2.test.js`

P0lip:
That'd be https://github.com/stoplightio/spectral/blob/develop/test-harness/scenarios/examples.oas2.scenario

mnaumanali94:
Closing as stale.

PR #2573, opened 2024-01-10
fix(rulesets): required readOnly and writeOnly properties should not …

### LINKED ISSUE #1274: `oas3-valid-oas-content-example` rule needs to obey readOnly/writeOnly property
**Describe the bug**

According to https://swagger.io/docs/specification/data-models/data-types/ :
> If a readOnly or writeOnly property is included in the required list, required affects just the relevant scope – responses only or requests only. That is, read-only required properties apply to responses only, and write-only required properties – to requests only.

The rule `oas3-valid-oas-content-example` should not raise error when a `readOnly` and `required` property is missing in request example. And also `writeOnly` property isn't required in response example.

The rule `oas3-valid-media-example` has the same bug as well.

**To Reproduce**

1. Given this OpenAPI/AsyncAPI document [index.yaml](https://github.com/stoplightio/spectral/files/4889475/index.yaml.txt)
2. Run this CLI command `spectral lint index.yaml`
3. See error
> 16:21    error  oas3-valid-oas-content-example  Object should have required property `self`

**Expected behavior**
This error report is not expected.

The `self` property is readOnly, it isn't required in request.

 - Library version: 5.4.0

pplr:
…be considered required for respectively request and response bodies

Fixes #1274.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


packages/rulesets/src/oas/functions/oasExample.ts
```diff
@@ -146,6 +182,41 @@ function cleanSchema(schema: Record<string, unknown>): void {
...
+    traverse(schema, {}, <traverse.Callback>((
+      fragment,
+      jsonPtr,
+      rootSchema,
+      parentJsonPtr,
+      parentKeyword,
+      parent,
+      propertyName,
+    ) => {
+      if ((fragment.readOnly === true && readOnlyProperties) || (fragment.writeOnly === true && writeOnlyProperties)) {
```
P0lip, line 209: could we also change the schema to `false` or `not: true`?
That's to ensure we catch if the property is defined.
To illustrate it with example:

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "string",
      "readOnly": true
    }
  },
  "required": [
    "foo"
  ]
}
```

The above schema would be transformed to

```json
{
  "type": "object",
  "properties": {
    "foo": false
  }
}
```

This way we wouldn't let property `"foo"` be defined.

P0lip (commented):
Thanks for the PR!
I left one remark. LMK if you have questions.

packages/rulesets/src/oas/functions/oasExample.ts
pplr: Hi @P0lip, do you request to make readOnly / writeOnly properties forbidden on request / response? Not sure that this is the exact definition of the read and writeOnly flags.

The 3.0 OAS spec says "SHOULD NOT".

For OAS 3.1 / json schema, the readOnly / writeOnly definition is more open.

> If `readOnly` has a value of boolean true, it indicates that the value of the
> instance is managed exclusively by the owning authority, and attempts by an
> application to modify the value of this property are expected to be ignored or
> rejected by that owning authority.

Your proposal could be optional and enabled with a flag like 'strict-read-write-only' ?

mnaumanali94 (approved):

PR #2581, opened 2024-01-22
feat(rulesets): add oas3_1-servers-in-webhook and oas3_1-callbacks-in…

kaylachun:
…-webhook rules

Fixes #[ISSUE_NUMBER].

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

>

**Screenshots**

**Additional context**

Added two new spectral rules:
- oas3_1-servers-in-webhook
  - warns if servers is defined in a webhook
- oas3_1-callbacks-in-webhook
  - warns if callbacks is defined in webhook operation


packages/rulesets/src/oas/index.ts
```diff
@@ -707,5 +707,23 @@ const ruleset = {
         },
       },
     },
+    'oas3_1-servers-in-webhook': {
+      message: 'Servers should not be defined in a webhook.',
+      formats: [oas3_1],
+      recommended: true,
+      given: ['$.webhooks.servers'],
```
daniel-white: i think we want this inside of the webhook operation, not at the webhooks object

packages/rulesets/src/oas/index.ts
```diff
@@ -707,5 +707,23 @@ const ruleset = {
...
+        function: undefined,
+      },
+    },
+    'oas3_1-callbacks-in-webhook': {
+      message: 'Callbacks should not be defined in a webhook.',
+      formats: [oas3_1],
+      recommended: true,
+      given: ['$.webhooks[*][*].callbacks'],
+      then: {
+        function: undefined,
```
daniel-white, line 734: are these required?

kaylachun: Are you referring to the `function: undefined` part?  That function is checking if callbacks is undefined and if it's not then it'll show the warning.

packages/rulesets/src/oas/index.ts
kaylachun: Ohhh okay servers can be defined on both.  Should the warning only be at the webhook operation level?

kaylachun: I'm reading through the oas docs and it seems like we should be showing a warning at both levels (the servers array at the path item object level would apply to all operations and servers at the operation level is for that operation only)

kaylachun: Ok added that here https://github.com/stoplightio/spectral/pull/2581/commits/fb14f54d522c3c6ac523dc59ab6c5149c327900d

mnaumanali94 (approved):

PR #2765, opened 2024-12-21
feat(functions): expand xor function and add or function

cuttingclyde:
Replaces PR https://github.com/stoplightio/spectral/pull/2616 with addition of independent or function.

Fixes #2396 . This has now been merged with stoplightio:develop containing prior "xor" PR https://github.com/stoplightio/spectral/pull/2614.

**Checklist**

- [X] Tests added / updated
- [X] Docs added / updated

**Does this PR introduce a breaking change?**

- [ ] Yes
- [X] No

(Just saw JeanArhancet's similar pull request, https://github.com/stoplightio/spectral/pull/2614 .)

FDX (see below) has extended XOR here to support from 1 to any number of properties. This would match the XOR logical function, that as long as exactly one of the (1 or more) properties is defined, then the rule passes. With just one property defined in the functionOptions, xor becomes redundant with defined function.

Updated error response messages to account for possibility of schema matching any number of properties or schema not matching any of any number of functionOption properties.

Updated functionOptions error messages to show example arrays containing from one to three ("and etc") properties.

Since then can take an array which implements AND-ing of rule requirements, this PR also defines and adds a new OR function. OR works just like xor for requiring at least one matching property, but then permits more than one matching property as valid and satisfying the rule.

Updated functions documentation for `xor` changes to allow more than two properties and new `or` function.

**Additional context**

The Financial Data Exchange (FDX, https://financialdataexchange.org) industry consortium for Open Banking delivers 13 OpenAPI 3.1 files with nearly 17,000 lines, 47 paths, 70 operations and 374 defined schemas, We enforce our FDX API Style Guide using 48 custom spectral rules and 23 overrides over 870 ruleset lines. We wanted to enforce that `type` was defined for each schema, but a schema can get its type in 3 (or 4) ways: `type`, `$ref`, or `oneOf` . (We are not using `allOf` to define `type`.) This was ideal for an unresolved xor rule with properties `type`, `$ref`, or `oneOf` . We initially implemented this as a custom function which met our needs, so are now contributing this to the spectral community. Our rule using our custom `xor` function:

``` yaml
  fdx-property-type-defined:
    description: A valid type must be set for each property (fdx-schema-type ensures oneOf and $ref entities have type defined)
    message: '{{property}} property has zero or multiple defined types from: "type", "oneOf", "$ref"'
    severity: error
    given:
      - $..properties[*]
    # Do not resolve here so $ref can be used as xor function option
    resolved: false
    then:
[...]

PR #2798, opened 2025-03-20
feat(functions): add or function

### LINKED ISSUE #2396: Support more than 2 logicals in XOR core function

frankkilcommins:
This PR supersedes #2765 to provide a clean history with only the intended changes from CuttingClyde.

Changes:
- Added xor and or functions.
- Updated optionSchemas.ts and tests.

Original work credited to @cuttingclyde.

fixes #2396

### DIFF
--- a/docs/reference/functions.md
+++ b/docs/reference/functions.md
@@ -245,9 +245,34 @@ unused-definition:
       reusableObjectsLocation: "#/definitions"
 ```
 
+## or
+
+Communicate that one or more of these properties is required to be defined. `functionOptions` must contain at least two properties. **or** will require that _at least_ one of them is defined.
+
+<!-- title: functionOptions -->
+
+| name       | description             | type       | required? |
+| ---------- | ----------------------- | ---------- | --------- |
+| properties | the properties to check | `string[]` | yes       |
+
+<!-- title: example -->
+
+```yaml
+schemas-descriptive-text-exists:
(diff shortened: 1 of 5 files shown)


mnaumanali94 (approved):
Looks good! Is there an existing rule that could use or that we could update? 🤔

PR #2809, opened 2025-04-17
feat(functions): add unicodeRegExp option to schema core function

### LINKED ISSUE #2419: Spectal complains when a openapi pattern uses a Unicode character set
**Describe the bug**

spectral incorrectly flags an example corresponding to a valid regexp pattern as invalid.

JSON Schema says that patterns should be processed as Unicode regexp, but it appears that the fix #1787 for #1782 turned off `unicodeRegExp` :
```
unicodeRegExp: false,
```

Thus for the following property

```
openapi: 3.1.0
components:
  schemas:
    label:
      description: The human readable label for the report type.
      type: string
      minLength: 1
      maxLength: 55
      pattern: '^\P{Cc}{1,55}$'
      example:
        Admin User Activity
```

We get

```
 11:9    error  oas3-valid-schema-example  "example" property must match pattern "\P{Cc}{1,55}"      components.schemas.label.example
```

even though this is a valid example that matches the regex

```node
> 'Admin User Activity'.match(/^\P{Cc}{1,55}$/u)
'Admin User Activity'.match(/^\P{Cc}{1,55}$/u)
[
  'Admin User Activity',
  index: 0,
  input: 'Admin User Activity',
  groups: undefined
]
> 
```

**To Reproduce**

1. Save the sample as `openapi.yaml`
3. Run this CLI command 
```
echo 'extends: spectral:oas' > .spectral.yaml
spectral lint openapi.yaml
```

5. See error

```
 11:9    error  oas3-valid-schema-example  "example" property must match pattern "\P{Cc}{1,55}"      components.schemas.label.example
```

**Expected behavior**

no false negative error for a valid pattern that matches a pattern with a charset pattern

Since a bug fix intentionally disables Unicode, and enabling it may break others, I would like an _option_ to enable it, 
either in the CLI options or in the spectral ruleset.

 - Library version: spectral cli 6.4.2
 - OS: MacOS 13.2.1
 - Browser: n/a

mtso:
Fixes #2419.

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [ ] Yes
- [x] No

**Screenshots**

Before:

Screenshot 2025-04-17 at 8 09 46 PM
  
  
After: 

Screenshot 2025-04-17 at 8 09 02 PM

**Added unicodeRegExp to core function docs**
Screenshot 2025-04-17 at 8 21 44 PM

**Additional context**

Adds an option `unicodeRegExp` to the `schema` core function. Defaults to "false". When "true", uses the unicode flag "u" with "pattern" and "patternProperties" fields during schema validation; otherwise does not use the "u" flag when "false" or omitted.

The `unicodeRegExp` option is optional, and thus backwards compatible with the existing behavior of spectral's `schema` core function. When the option is omitted, the behavior is defaulted to "false".

The `unicodeRegExp` option is also added as an optional flag to `oasSchema` and `oasExample` oas functions. To preserve the existing behavior of the standard oas ruleset, `unicodeRegExp: false` has been explicitly added to the standard `oas` ruleset's `duplicated-entry-in-enum`, `oas2-valid-schema-example`, `oas2-valid-media-example`, `oas3-valid-media-example` and `oas3-valid-schema-example` rules.

Adding this option allows rulesets that use the `schema` core function to opt into using the unicode "u" flag when validating the "pattern" and "patternProperty" fields.

Adding this option to `oasExample` and `oasSchema` also allows rulesets that extend from `spectral:oas` to replace the `oas3-valid-media-example` and `oas3-valid-schema-example` rules with a version that opts into using unicode regular expressions. Doing so is currently a little involved, though, requiring re-specifying the `given` block of the rules as shown below. But it at least makes the option available to extending rulesets.

```js
// functions/oasExample.js
import { default as oasExample } from '@stoplight/spectral-rulesets/dist/oas/functions/oasExample';

export default oasExample;
```

```yaml
# spectral.yaml
extends: ['spectral:oas']

functions:
- 'oasExample'

rules:
  oas3-valid-schema-example:
    given:
      - $.components.schemas..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]
      - $..content..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]
      - $..headers..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]
      - $..parameters..[?(@property !== 'properties' && @ && (@ && @.example !== void 0 || @.default !== void 0) && (@.enum || @.type || @.format || @.$ref || @.properties || @.items))]
    then:
      function: oasExample
      functionOptions:
        schemaField: '$'
        oasVersion: 3
        type: 'schema'
        unicodeRegExp: true

  oas3-valid-media-example:
    given:
      - '$..content..[?(@ && @.schema && (@.example !== void 0 || @.examples))]'
      - '$..headers..[?(@ && @.schema && (@.example !== void 0 || @.examples))]'
      - '$..parameters..[?(@ && @.schema && (@.example !== void 0 || @.examples))]'
    then:
      function: oasExample
      functionOptions:
        schemaField: 'schema'
        oasVersion: 3
        type: 'media'
        unicodeRegExp: true
```


mtso:
@mnaumanali94 / @frankkilcommins , when you get the chance, could you please take a look?

jwulf:
@mnaumanali94 @frankkilcommins @mtso — any chance we can get this merged?

slegarraga (approved):
Useful addition for schemas with Unicode property names — `unicodeRegExp` opt-in on the schema function keeps default behavior unchanged (#2419).

Ajv instance caching keyed by options is a nice touch; tests + docs updated. LGTM.

PR #2853, opened 2025-10-01
feat(formatters): add severity custom attribute to TeamCity inspections

Vampire:
**Checklist**

- [x] Tests added / updated
- n/a Docs added / updated

- [ ] Yes
- [x] No

**Additional context**

This PR adds the `SEVERITY` custom attribute to the inspection TeamCity service messages.
It is true that this is just a custom attribute, but it the one shown as example in the TeamCity documentation,
and it also is just half-custom, because `SEVERITY="ERROR"` is indeed considered by TeamCity when segregating the reported inspectinos into warnings and errors.

image


slegarraga (approved):
Adding `SEVERITY='…'` on TeamCity inspections is a nice QoL fix — consumers can filter/route by severity without parsing the message prefix.

Tests + harness scenario updated consistently. LGTM.

PR #2863, opened 2025-10-31
fix(rulesets): security-scheme-name rule added to validate security-scheme name

SB-akshaythakar:
This PR is adding rule to validate securityScheme name. 

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No

### DIFF
--- a/docs/reference/openapi-rules.md
+++ b/docs/reference/openapi-rules.md
@@ -479,6 +479,65 @@ TheBadModel:
         type: array
 ```
 
+### security-scheme-name
+
+Security scheme name must not contain whitespace.
+
+Security scheme names are used as identifiers in OpenAPI documents and should follow a consistent naming pattern. This rule ensures that the `name` property within security scheme definitions only contains alphanumeric characters, dots (.), underscores (_), and hyphens (-), preventing issues with whitespace or special characters.
+
+This rule applies to OpenAPI v2.0, v3.0, and v3.1.
+
+**Recommended:** Yes
+
+**Good Example**
+
+For OpenAPI v3.0 and v3.1:
+
+```yaml
+components:
+  securitySchemes:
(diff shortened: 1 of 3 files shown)


prafullaAtSB (dismissed):
Changes are good to go ahead.

bcoughlan:
Hi, reviewing this per email request. This change affects all Spectral users, as it adds a new linting rule to the [public OAS styleguide](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules). Wouldn't this validation change belong in the Prism repo?

SergioV85:
> This change affects all Spectral users, as it adds a new linting rule to the [public OAS styleguide](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules). Wouldn't this validation change belong in the Prism repo?

I think it's worth it to be in Spectral, because this will implement the rule from [HTTP Standart](https://datatracker.ietf.org/doc/html/rfc7230) -- no whitespace allowed in header name. So, we shouldn't allow users to create wrong specifications rather than validate them on execution/mock stage

bcoughlan:
I disagree that parsing issues belong in the linting tool, but I'm willing to disagree and commit as the consensus is otherwise.

I don't know the Spectral codebase at all, but I see a few issues with this PR as it stands:

1. Grammar of the rule description is inconsistent: `description: 'Ensure that security scheme should have valid name',`, whereas the rules above it are all in the format like `description: 'Server URL must not have trailing slash.`
2. Documentation needs to be updated for the change: `docs/reference/openapi-rules.md`
3. The rule specifies `formats: [oas2, oas3]` but OAS2 does not use the `components.securitySchemes` feature

bcoughlan (approved):

PR #2884, opened 2026-01-29
fix(repo): release step marking repo as save for git

tomek-tursa-sb:


.github/workflows/release.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       with:
         fetch-depth: 0
         token: ${{ secrets.GH_TOKEN }}
+    - name: Mark repo as safe for Git
```
barryadk: Should this be under the `Configure Git` section?

barryadk (approved):

PR #2911, opened 2026-03-12
chore: remove `es-aggregate-error`

### LINKED ISSUE #2899: suggestion: replace `es-aggregate-error` with native `AggregateError`
[`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError#browser_compatibility) is available since Node 15.0.0

gameroman:
Closes #2899


slegarraga (approved):
Nice cleanup — dropping the `es-aggregate-error` polyfill makes sense if the supported Node range already has native `AggregateError`.

Diff is mostly test import churn + lockfile; no behavioral risk spotted. Thanks for keeping deps lean.

slegarraga (approved):
Re-checked the updated branch after the test and lint fixes. The diff is still a focused removal of the `es-aggregate-error` polyfill and its imports, and the updated assertion line numbers look correct. Re-approving.

PR #2934, opened 2026-04-06
fix(rulesets): exclude non-schema enum properties from duplicated-entry-in-enum rule

slegarraga:
The duplicated-entry-in-enum rule was incorrectly triggering when enum is used as a property name in contexts like discriminator mappings or example values, rather than as a JSON Schema keyword. For example, a discriminator mapping like `mapping: { enum: ... }` or example data with an `enum` field would wrongly trigger the rule.

The fix narrows the JSONPath filter to only match when the enum value is actually an array, since a JSON Schema enum keyword must have an array value. This avoids false positives while still correctly catching duplicate entries in actual schema enums.

**Checklist**

- [x] Tests added / updated
- [ ] Docs added / updated

- [ ] Yes
- [x] No


tomek-tursa-sb (approved):

packages/rulesets/src/arazzo/schemas/validators.ts
tomek-tursa-sb, line 1: changes in this file are probably redundant

tomek-tursa-sb: @slegarraga would you mind to have a look?

slegarraga:
Thanks for the ping, @tomek-tursa-sb. You were right — the generated Arazzo validators bundle change was redundant for this fix.

I restored `packages/rulesets/src/arazzo/schemas/validators.ts` to match `develop`, so the PR now only changes the OAS rule plus the regression tests. Commit: https://github.com/slegarraga/spectral/commit/9b51f58006e589c8f7add3620bacbd0140fc5096

tomek-tursa-sb:
@slegarraga thank you for this contribution 💪

tomek-tursa-sb (approved):

packages/rulesets/src/oas/index.ts
```diff
@@ -159,7 +159,7 @@ const ruleset = {
       severity: 'warn',
       recommended: true,
       message: '{{error}}',
-      given: ["$..[?(@property !== 'properties' && @ && @.enum)]"],
+      given: ["$..[?(@property !== 'properties' && @.enum && @.enum.constructor.name === 'Array')]"],
```
anderaus, line 162: Was the `@ &&` null guard removed on purpose? I believe I'm seeing crashes when nimma hits a `null` node.

anderaus: Shouldn't this be:
`$..[?(@property !== 'properties' && @ && @.enum && @.enum.constructor.name === 'Array')]`

PR #2955, opened 2026-05-14
ci: run tests on all versions of Node.js supported by this package

gameroman:


slegarraga (commented):
CI matrix expansion makes sense. Quick check: Spectral package.json `engines` still lists minimum Node — if 16.x is below supported range, matrix will fail on purpose (good) or should be dropped. Otherwise LGTM from a contributor perspective.

slegarraga (commented):
Node matrix CI is valuable here — one note: if the matrix explodes wall-clock on PRs, consider `fail-fast: false` only on release branches or cap to LTS + current.

Otherwise supportive of broader Node coverage aligned with `engines` in package.json.

slegarraga (approved):
Aligns CI with the Node versions declared in \`engines\` — good coverage expansion.

One thought for maintainers: if wall-clock on PRs becomes an issue, consider pinning the matrix to LTS + current rather than every supported line. Otherwise looks good to me.

PR #2960, opened 2026-05-21
fix(rulesets): duplicated-entry-in-enum - support null values

### LINKED ISSUE #2959: Missing null guard causes "Cannot read properties of null (reading 'enum')"
I believe a bug was introduced in https://github.com/stoplightio/spectral/pull/2934/, see link below

> Was the `@ &&` null guard removed on purpose? I believe I'm seeing crashes when nimma hits a `null` node.  

 _Originally posted by @anderaus in [#2934](https://github.com/stoplightio/spectral/pull/2934/changes#r3280802340)_

Bug can be reproduced using this input:

```
{
  "openapi": "3.0.2",
  "info": { "title": "Test", "version": "1.0" },
  "paths": {},
  "components": {
    "examples": {
      "test": {
        "value": {
          "foo": null
        }
      }
    }
  }
}
```

DougEdey-Slice:
Fixes #2959.

This handles `null` enum values, which is supported/required by [OAS3.1](https://www.speakeasy.com/openapi/schemas/enums#openapi-v31) for nullable enums

This updates the tests & the ruleset to support nulls, which OAS3.0 does support: https://swagger.io/docs/specification/v3_0/data-models/enums/#nullable-enums

First time contributor so let me know if I need anything else :)

**Checklist**

- [x] Tests added / updated
- [x] Docs added / updated

- [x] No


TadejOstanek (approved):

slegarraga:
Nice follow-up to the enum edge cases — null in enum arrays is easy to miss. From #2934 land: worth confirming `resolved: false` behavior on $ref-wrapped enums matches your test cases. Thanks for pushing this.

slegarraga (approved):
**LGTM** — null enum handling matches OAS 3.0 nullable enums and 3.1 semantics from #2959.

- Test matrix covers null in `enum` arrays without false positives on valid enums.
- `validators.ts` tweak is minimal and scoped to the duplicated-entry rule path.

No concerns from my side; thanks for the fix.

slegarraga (approved):
LGTM — null enum handling matches #2959; tests cover nullable enums without false positives. Thanks for the fix.

slegarraga:
@DougEdey-Slice Thanks — this addresses the regression from #2934 (my merge). Suggest adding a test case for issue #2959 reproducer (`components.examples.*.value` with `foo: null`) so we don’t regress the JSONPath guard again. Approved from my side.

slegarraga:
Opened **#2963** with the #2959 reproducer test + same  guard — happy to close #2963 if you prefer landing this PR instead.

slegarraga:
@tomek-tursa-sb Maintainer review when possible — 2 contributor approvals, fixes #2959 (production crash). Alternative: **#2963** has the same guard + explicit reproducer test from the issue. Either PR unblocks users on spectral-cli 6.16.x.

slegarraga (commented):
Reviewed this against #2959 and my own #2963, since they overlap.

The 3.1 coverage here is the right call: `null` is a valid enum member in OAS 3.1 (`enum: [1, 2, 3, null]`), and flagging duplicated nulls (`[null, …, null]` → items #0 and #6) is a real gap the original rule missed. Nice.

One coordination note: this and #2963 fix different facets of #2959 and will conflict on `oas/index.ts` and `duplicated-entry-in-enum.test.ts`. #2963 adds `@ != null` to the rule's `given` selector to stop a TypeError when the selector hits a *null node* (e.g. `foo: null` inside an example value, where reading `@.enum` throws). That null-node crash isn't exercised by the cases here (these are all schema `enum`s), so I think both changes are needed: the selector guard to avoid the crash, plus your 3.1 null-entry semantics. Happy to rebase #2963 on top of this so it's a single clean change if that's easier to merge.

DougEdey-Slice:
@tomek-tursa-sb @pareshb3089 Could we get a review on this please?

Patrycja-dz (approved):

MichakrawSB (approved):

Issue #2643, opened 2024-06-14
OAS 3.1 Security Schemes Relax "Scope" definition requirements - Updated Function/Rule ?


dtolb:
**Describe the bug**

According to the [OAS 3.1](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#security-requirement-object) security requirement objects. BearerAuth schemes in 3.1 are allowed to use scopes on the path without using Oauth2 schema. The rule [oas3-operation-security-defined](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#oas3-operation-security-defined) is incorrectly reporting failure for 3.1.x documents.  

**To Reproduce**

1. Given this OpenAPI/AsyncAPI document '

```yaml
openapi: 3.1.0
info:
  title: Non-oAuth Scopes example
  version: 1.0.0
paths:
  /users:
    get:
      security:
        - bearerAuth:
            - 'read:users'
            - 'public'
components:
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: jwt
      description: 'note: non-oauth scopes are not defined at the securityScheme level'
```

3. Run this CLI command '....'

`spectral -r oas.default.yaml my_oas.yaml`

5. See error

`... must be listed among scopes.  `

**Expected behavior**
Given this is a 3.1 document, I wouldn't expect an error for invalid OAS formatting from OAS 3.0. The [oasSecurityDefined.ts](https://github.com/stoplightio/spectral/blob/develop/packages/rulesets/src/oas/functions/oasSecurityDefined.ts) function should support a new option 3_1 that allows updated security scopes

ponelat:
Hey @dtolb, yup you are correct OAS 3.1, [does indeed support having "roles"](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#patterned-fields-3) (to use the non-oAuth terminology) in the array within the security requirement. 
So shouldn't report an error. 

Would welcome a fix here, but it is a low priority.

slegarraga:
Repro makes sense — in OAS 3.1, \`bearerAuth\` scopes on the **operation** security requirement don't have to be declared on the scheme the same way as OAuth2.

Likely fix path: teach \`oas3-operation-security-defined\` (or 3.1-specific variant) to allow non-OAuth bearer scopes when \`type: http\` + \`scheme: bearer\`. Happy to help test a PR if someone picks this up.

[...]

PR #2961, opened 2026-05-22
feat(oas): add object-with-properties-requires-type-object rule

### LINKED ISSUE #2203: Add "Objects with properties must have type: object" rule to builtin spectral:oas ruleset
I still think this is a sensible (opinionated, admittedly) addition to the spectral:oas ruleset.
IMHO if you define properties in an OAS schema ("struct" data) you probably never want to accept data as valid that doesn't even have these properties/attributes.

(I confusingly added that to #2136 so that kind of got mixed up with this separate JSON schema related issue, sorry)

Here goes, again:

As suggested by @P0lip on #2139 I propose to add to such a built-in ruleset:

```
---
extends: "spectral:oas"
rules:
  properties-type-object:
    description: "Objects with properties must have type: object."
    given: "$..[?(@ && @.properties)]"
    severity: error
    # must not be resolved because $ref-defined objects shall have its type
    # defined in the referenced schema, not the referencing field (and $ref 
    # must be the only property, anyway)
    resolved: false
    # type field must be defined and contain the value object
    then:
      - field: "type"
        function: defined
      - field: "type"
        function: enumeration
        functionOptions:
          values:
            - object
```

Reasoning:
The way OpenAPI/JSON Schema validation works wrt to structures with properties but no `type: object` is probably not what most users would expect (I certainly didn't):

Because IIUC a schema like

```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "properties": {
    "number": { "type": "number" },
    "street_name": { "type": "string" }
  },
  "required": ["number"],
}
```

is not a "de facto" restriction on a JSON object i.e. the key-value mapping aka dictionary to carry the given properties. According to [JSON Schema Cort: Assertions and Instance Primitive Types](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.7.6.1) only those assertions are used for validating a value which belong to the "Validation Keywords" of the data type.

Which means that all of these are valid against the above schema:

Valid:

```
{
  "number": 10
}
```

Valid:
```
""
```

Valid:
```
false
```

(basically, anything - just by accidentally omitting `type: object` for a "struct")

I gather the user will usually expect "struct" data to be only treated as  valid if (required) properties are present in the instance object data model. The proposed rule thus omits an error for OpenAPI object schemas that define properties but not `type: object`.

Best, Holger

_Originally posted by @hjoukl in https://github.com/stoplightio/spectral/issues/2136#issuecomment-1114544997_

slegarraga:
## Summary

Implements the builtin rule proposed in #2203: schemas that define `properties` must also declare `type: object`.

- Rule id: `object-with-properties-requires-type-object`
- `recommended: false` (opinionated; opt-in like other style rules)
- `resolved: false` (type belongs on the schema node, not on `$ref` targets)
- Regression tests for OAS2 + OAS3

## Why

Without `type: object`, JSON Schema validation does not restrict instances to objects — strings/booleans can pass unexpectedly when only `properties` is set.

## Test plan

- [x] `yarn jest packages/rulesets/src/oas/__tests__/object-with-properties-requires-type-object.test.ts`

Closes #2203

Made with [Cursor](https://cursor.com)


slegarraga:
@maintainers Friendly nudge — rolling review welcome. Rule is opt-in (`recommended: false`) per #2203 discussion; happy to flip to recommended or adjust JSONPath if you prefer a stricter oas3.1 `type: [object]` case.

slegarraga:
Ready for review when you have bandwidth — mergeable, Wiz CI green, tests added. Closes #2203. Thanks.

slegarraga:
@tomek-tursa-sb Hi — you merged #2934 last week; this follow-up adds the opt-in rule from #2203 (tests + docs). Mergeable, Wiz green. A review when you have a cycle would be appreciated. Thanks.

slegarraga:
Local jest on rule tests: 5/5 pass. Branch is mergeable; let me know if you want this marked `recommended: true` before merge. Thanks.

slegarraga:
Related: #2962 (OAS 3.1 bearer scopes) — same ruleset area if useful to batch review.

slegarraga:
**Ready for merge** — closes #2203, MERGEABLE, CI summary ✅, local rule tests pass.

Opt-in rule (not `recommended: true` unless you want it default-on). Happy to flip that flag in a follow-up commit.

slegarraga:
@tomek-tursa-sb Batch review request when you have a cycle — three small Spectral PRs from the same contributor window:

| PR | What |
|----|------|
| **#2963** | Hotfix: null guard regression from #2934 → fixes #2959 crash (9 tests) |
| **#2961** | New opt-in rule `object-with-properties-requires-type-object` → closes #2203 |
| **#2962** | OAS 3.1 bearer scopes on operations → closes #2643 |

All MERGEABLE, Wiz green. #2963 overlaps #2960 — land either for the null guard. Thanks again for merging #2934.

PR #2966, opened 2026-06-03
fix(formatters): render warnings in markdown output

### LINKED ISSUE #2928: `markdown` output format ignores absence of `--display-only-failures`
The `markdown` formatters never renders warnings, even though `--display-only-failures` is not given.

slegarraga:
## Summary

The `markdown` formatter never rendered warnings, even when `--display-only-failures` was not passed. Fixes #2928.

## Root cause

`markdown.ts` filtered results with `result.severity <= failSeverity`. `failSeverity` defaults to `error` (`DiagnosticSeverity.Error === 0`), so only errors survived and warnings (severity `1`) were dropped regardless of `--display-only-failures`.

`failSeverity` controls the process exit code, not what is displayed. The `--display-only-failures` filtering is already applied in the CLI before formatting (`filterResultsBySeverity` in `commands/lint.ts`), and the other formatters (`stylish`, `text`, `json`, `pretty`) render every result they receive.

## Fix

Drop the redundant `failSeverity` filter from the markdown formatter so it renders all results it is given, consistent with the other formatters.

## Test

Added a regression test asserting warnings are rendered with the default `failSeverity` of `error`. The existing markdown test continues to pass.

```
PASS @stoplight/spectral-formatters packages/formatters/src/__tests__/markdown.test.ts
  Markdown formatter
    ✓ should format as markdown table
    ✓ renders warnings regardless of failSeverity (#2928)
```


micryc:
Closing this PR as it seems to be duplicated by [3039](https://github.com/stoplightio/spectral/pull/3039)

PR #2968, opened 2026-06-03
fix(cli): support dashed formats in `--output.<format>`

### LINKED ISSUE #2926: Output file for `github-actions` and `code-climate` formatters cannot be specified
When giving multiple `--format` to the executable, you also need to specify an `--output.<format>`  per format.
But for `github-actions` and `code-climate` this does not work, at least not how documented.

Using `--output.github-actions` and `--output.code-climate` like documented and working for other formats,
you get "The number of outputs must match the number of formats".

Using - as a guess - `--output.githubActions` and `--output.codeClimate` instead,
you get "Missing outputs for the following formats: github-actions, code-climate".

I guess somewhere the CLI parser does note like the dash or something like that.
Or if it is possible somehow, at least it is well hidden.
Also tried `--output.github.actions` / `--output.code.climate` and `--output.github_actions` / `--output.code_climate`

slegarraga:
## Summary

`--output.github-actions` / `--output.code-climate` did not work, even though the docs show `--output.<format>`. Fixes #2926.

## Root cause

yargs camel-cases dotted keys, so `--output.github-actions <file>` is parsed as `output.githubActions` (and may keep the original `github-actions` key too). The validation then compares the `output` keys against the `--format` values, which use the dashed names, producing either:

- `The number of outputs must match the number of formats` (duplicate key), or
- `Missing outputs for the following formats: github-actions, code-climate` (camel-cased key).

## Fix

Add a `coerce` to the `output` option that re-keys the object by the canonical format names, taking the value from either the dashed or the camel-cased key. Dashed formats now line up with `--format`.

## Test

Added a test in `lint.test.ts`:

```
lint --format github-actions --format code-climate \
     --output.github-actions ga.txt --output.code-climate cc.json <doc>
```

asserts `writeOutput` is called with `ga.txt` and `cc.json`. Full suite: 21 passed.


kopernic-pl:
@slegarraga please rebase. I think it's OK to be merged as soon as conflict is resolved and we get another review.

kopernic-pl (approved):

micryc (approved):

tomek-tursa-sb (approved):

PR #2995, opened 2026-07-02
test: cover optional operation security alternatives

kriptoburak:
## Summary

Adds a focused OpenAPI 3.1 regression case for operation-level security alternatives that include:

- API key auth
- bearer auth
- an empty security requirement object for anonymous access

This is valid OpenAPI behavior and is already accepted by the rule. The new case keeps that behavior covered directly.

## Validation

- `git diff --check`
- Targeted Jest was not run locally because the repo dependency install hit local disk exhaustion while populating the Yarn cache.

### DIFF
--- a/packages/rulesets/src/oas/__tests__/oas3-operation-security-defined.test.ts
+++ b/packages/rulesets/src/oas/__tests__/oas3-operation-security-defined.test.ts
@@ -36,6 +36,50 @@ testRule('oas3-operation-security-defined', [
   },
 
   {
+    name: 'valid OpenAPI 3.1 operation with optional security alternatives',
+    document: {
+      openapi: '3.1.0',
+      components: {
+        securitySchemes: {
+          apiKey: {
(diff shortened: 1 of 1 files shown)


slegarraga (approved):
Nice focused regression case. The empty `{}` requirement correctly represents anonymous access, and the fixture covers both root and operation-level alternatives. CI will be the authoritative check since local Jest was not run. Looks good to me.

kriptoburak:
Rebased onto the current `develop` branch at `e129838e329e106fb1f5efb784fb3a334899c414`. `git range-diff` confirms the regression patch is unchanged.

Local validation:

- Full Jest suite: 200 suites passed; 1,971 tests passed; 1 skipped
- Focused OAS3 security suite: 5 tests passed
- Full workspace build passed
- Full ESLint passed with 0 errors (40 existing warnings)
- `git diff --check` passed

This remains a test-only change.

